PR #1, opened 2015-10-19
add event script macros

yenatch:


YamaArashi:
In the "events.s" file:
Line 505 has a comment that doesn't begin with a semicolon.
"storeitems" is redefined on line 1408.

PR #8, opened 2016-09-11
rng

DizzyEggg:


YamaArashi:
This did not match the current system used for symbols. The RNG has been decompiled in 8f9ed6e585a361e957650b871fc56a0406c2dac8

PR #9, opened 2016-10-31
Ported task.c and others from pokeruby

Diegoisawesome:


tools/aif2pcm/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2016 huderlem
```
YamaArashi: could you remove this file?

PR #22, opened 2017-08-31
bring pokeemerald to usable state

DizzyEggg:
Label things, move decompiled and labelled files.


DizzyEggg:
I'd say it's a good start. Should be merge-able now, I suppose.

PR #27, opened 2017-09-10
decompile script.s

DizzyEggg:


src/script.c
```diff
@@ -0,0 +1,434 @@
...
+#include "util.h"
+
+#define RAM_SCRIPT_MAGIC 51
+
+extern u8* gUnknown_020375C0;
+
+extern bool32 sub_801B27C(void);
+
+// ewram bss
+IWRAM_DATA u8 sScriptContext1Status;
```
camthesaxman: Make these actually static.

src/script.c
```diff
@@ -0,0 +1,434 @@
...
+    scriptData->magic = RAM_SCRIPT_MAGIC;
+    scriptData->mapGroup = mapGroup;
+    scriptData->mapNum = mapNum;
+    scriptData->objectId = objectId;
+    memcpy(scriptData->script, script, scriptSize);
+    gSaveBlock1Ptr->ramScript.checksum = CalculateRamScriptChecksum();
+    return TRUE;
+}
+
+u8 *GetRamScript(u8 objectId, u8 *script)
```
camthesaxman: Did this not work with the same if structure from pokeruby?

DizzyEggg: Nope, the ruby one is a 'fake-match'. :P

src/script.c
DizzyEggg: Fixed.

src/script.c
camthesaxman: Yeah, I tried porting the Ruby one (which looks better) here, and it had stupid register differences. Oh well.

PR #29, opened 2017-09-13
decompile berry

DizzyEggg:


include/global.h
```diff
@@ -649,8 +649,10 @@ struct DaycareData
     u8 stepCounter;
 };
 
-#define FLAGS_NUMBER    300
-#define VARS_NUMBER     256
+#define MAP_OBJECTS_NUMBER  16
```
camthesaxman: I would call these `MAP_OBJECTS_COUNT`, etc. to be a bit more specific.

src/berry.c
```diff
@@ -0,0 +1,1254 @@
...
+
+extern const u8 BerryTreeScript[];
+extern const struct BerryTree gBlankBerryTree;
+
+#define BERRY_NAME_LENGTH 6
+
+#define FIRST_BERRY ITEM_CHERI_BERRY
+#define LAST_BERRY ITEM_ENIGMA_BERRY
+
+#ifdef ENGLISH
```
camthesaxman: You can get rid of this, since we're only building English Emerald.

DizzyEggg: Can the defines stay though? Don't want to remove all the berry names DEFINES manually :P.

PR #31, opened 2017-09-16
rom_81700F8

PikalaxALT:


include/rom4.h
```diff
@@ -0,0 +1,15 @@
+//
+// Created by scott on 9/15/2017.
+//
+
+#ifndef POKEEMERALD_ROM4_H
```
camthesaxman: Please be consistent and use `GUARD_ROM4_H`. Also, we've renamed rom4 to overworld in pokeruby.

src/berry.c
```diff
@@ -807,15 +807,15 @@ const struct Berry gBerries[] =
 // unused
 void ClearEnigmaBerries(void)
 {
-    CpuFill16(0, &gSaveBlock1Ptr->enigmaBerry, sizeof(gSaveBlock1Ptr->enigmaBerry));
+    CpuFill16(0, &gSaveBlock1Ptr->enigmaBerry, 52 /*FIXME: sizeof(gSaveBlock1Ptr->enigmaBerry)*/);
```
Diegoisawesome: This has already been fixed in an earlier PR so you can remove the FIXME sections.

PR #36, opened 2017-09-18
Decompile Egg Hatch

DizzyEggg:
Huh?
...
...
...
A decompiled file hatched from the Egg!
Would you like to merge the newly hatched PR?


src/trig.c
```diff
@@ -1,9 +1,6 @@
 #include "global.h"
 #include "trig.h"
 
-// Converts a number to Q8.8 fixed-point format
```
Diegoisawesome: Why move one and not the other?

yenatch:
What is `data/data3`?

include/text.h
```diff
@@ -138,28 +138,37 @@ struct FontInfo
...
 
-struct KeypadIcon {
+struct KeypadIcon
+{
     u16 tile_offset;
     u8 width;
     u8 height;
 };
 
+struct __attribute__((packed)) TextColor
```
yenatch: `__attribute__((packed))` doesn't do anything in struct declares. you have to use it on the variable.

DizzyEggg: No, it does. It will make all variables of that struct packed. Compile my code with and without it and you'll see it doesn't match without it.

src/trig.c
DizzyEggg: So far I didn't encounter any other situation that would require this macro.

data/field_map_obj.s
```diff
@@ -1268,16 +1268,16 @@ gUnknown_0850DC50:: @ 850DC50
...
 .4byte gUnknown_0850E3A0
 .4byte gUnknown_0850E3AC
 .4byte gUnknown_0850E5C4
 .4byte gUnknown_0850E5CC
 .4byte gUnknown_0850E5D4
 .4byte gUnknown_0850E468
-.4byte gUnknown_0850E864
-.4byte gUnknown_0850E870
+.4byte 0x850E864    @ FIX ME: ADD LABEL
+.4byte 0x850E870    @ FIX ME: ADD LABEL
```
yenatch: whats up with these

PR #37, opened 2017-09-18
Berry fix program

scnorton:


data/data_emerald2.s
```diff
@@ -34,36 +34,7 @@ gUnknown_08617E34:: @ 8617E34
...
-gUnknown_0861815B:: @ 861815B
-	.incbin "baserom.gba", 0x61815b, 0x5
-
-gUnknown_08618160:: @ 8618160
-	.incbin "baserom.gba", 0x618160, 0x18
-
-gUnknown_08618178:: @ 8618178
-	.incbin "baserom.gba", 0x618178, 0x49
-
+    .section .rodata.2
```
Diegoisawesome: This should create a new file, not a new section of the same file.

data/multiboot_berry_glitch_fix.s
```diff
@@ -3,6 +3,7 @@
 gUnknown_089A6550:: @ 89A6550
 	.incbin "baserom.gba", 0x9a6550, 0xc0
 
+gMultiBootProgram_BerryGlitchFix_Script:: @ 89A6610
```
Diegoisawesome: Why two names for the same address?

get_diffs.py
```diff
@@ -0,0 +1,24 @@
+addrs =  sorted([0xdd7360,
```
Diegoisawesome: This file seems to have been added by mistake.

PR #38, opened 2017-09-19
Battle dome cards

scnorton:


include/battle_dome_cards.h
```diff
@@ -0,0 +1,13 @@
+#ifndef GUARD_BATTLE_DOME_CARDS_H
```
Diegoisawesome: Are these functions called from any other file but battle_dome_cards.c? If not, no need to add a header file.

PikalaxALT: These functions are called in other components yet to be decompiled.  The idea is to have the header file now because it will be needed eventually.

src/battle_dome_cards.c
```diff
@@ -0,0 +1,430 @@
...
+};
+static const struct OamData gUnknown_0860B06C = {
+    .affineMode = 1, .size = 3
+};
+
+// .text
+
+static void nullsub_122(struct Sprite *sprite)
+{
+
```
ProjectRevoTPP: inconsistent style for nullsubs?

PR #39, opened 2017-09-20
Extract baserom calls from various battle data files and crt0.s

Diegoisawesome:


data/battle_4.s
```diff
@@ -7,44 +7,134 @@
...
+	.4byte gUnknown_082DB36A
+	.4byte gUnknown_082DB36A
+	.4byte gUnknown_082DB36A
+	.4byte gUnknown_082DB36A
+	.4byte gUnknown_082DB36A
+	.4byte gUnknown_082DB36A
+	.4byte gUnknown_082DB3F4
+
+	.align 2
+	window_template 0x00, 0x01, 0x03, 0x07, 0x0f, 0x1f, 0x003f
```
ProjectRevoTPP: is this really a window template?

data/battle_4.s
```diff
@@ -7,44 +7,134 @@
...
+	.align 2
+gUnknown_0831C3C0:: @ 831C3C0
+	.2byte 0xd75a @ tileTag
+	.2byte 0xd75a @ paletteTag
+	.4byte gUnknown_0831C3B8
+	.4byte gDummySpriteAnimTable
+	.4byte NULL
+	.4byte gDummySpriteAffineAnimTable
+	.4byte sub_804F450
+
```
ProjectRevoTPP: define a macro for this?

data/battle_4.s
```diff
@@ -7,44 +7,134 @@
...
+	.align 2
 sProtectSuccessRates:: @ 831C3D8
-	.incbin "baserom.gba", 0x31c3d8, 0x8
+	.2byte 0xffff, 0x7fff, 0x3fff, 0x1fff
 
+	.align 2
 gUnknown_0831C3E0:: @ 831C3E0
-	.incbin "baserom.gba", 0x31c3e0, 0x28
+	.2byte 0x0076, 0x00a5, 0x00a6, 0x0066
```
ProjectRevoTPP: move list?

data/battle_4.s
Diegoisawesome: Not 100% sure but it fits perfectly with the pattern I've consistently seen for all window templates.

data/battle_4.s
Diegoisawesome: Is the intermediate step of a macro absolutely necessary? It's going to end up in a C struct anyway.

PR #44, opened 2017-09-23
decompile field_special_scene.c

ProjectRevoTPP:


src/field_special_scene.c
```diff
@@ -0,0 +1,365 @@
...
+    box2 = GetTruckBoxMovement(data[0]) * 2; // bottom left box.
+    sub_808E82C(2, gSaveBlock1Ptr->location.mapNum, gSaveBlock1Ptr->location.mapGroup, 0, box2 - 3);
+    box3 = GetTruckBoxMovement(data[0]) * 4; // bottom right box.
+    mapNum = gSaveBlock1Ptr->location.mapNum;
+    mapGroup = gSaveBlock1Ptr->location.mapGroup;
+    zero = 0;
+    sub_808E82C(3, mapNum, mapGroup, -3, box3);
+
+    if (++data[0] == SECONDS(500)) // this will never run
+        data[0] = zero; // reset the timer if it gets stuck.
```
yenatch: Do either of these work instead of `zero`?
```c
data[0] -= data[0];
```
```c
data[0] -= SECONDS(500);
```

ProjectRevoTPP: tried both. the main problem is that zero (r4) is allocated in the middle of the function call, which is strange.

src/field_special_scene.c
```diff
@@ -0,0 +1,365 @@
...
+{
+    if (!FuncIsActiveTask(Task_HandleTruckSequence))
+    {
+        sub_808E82C(1, gSaveBlock1Ptr->location.mapNum, gSaveBlock1Ptr->location.mapGroup, 3, 3);
+        sub_808E82C(2, gSaveBlock1Ptr->location.mapNum, gSaveBlock1Ptr->location.mapGroup, 0, -3);
+        sub_808E82C(3, gSaveBlock1Ptr->location.mapNum, gSaveBlock1Ptr->location.mapGroup, -3, 0);
+    }
+}
+
+extern u8 GetSSTidalLocation(s8 *, s8 *, s16 *, s16 *); // should be in field_specials.h
```
Diegoisawesome: Any reason why you haven't created that already then? These externs would be cleaner in a shared header.

ProjectRevoTPP: Because it would be an extern as well since the function is a part of field_specials.c, but that file isnt decompiled yet.

Diegoisawesome: Fair enough. As a matter of tidiness, they should still probably not be down here scattered in between functions; you should probably move them to the top of the file, at least until the proper header file is created.

PR #45, opened 2017-09-23
Decompile Battle 4

DizzyEggg:
I'd like to rename:
"battle_4.c" to "battle_script_commands.c"
"battle_ai.c" to "battle_ai_script_commands.c"

Is that fine with everyone?


src/pokemon_3.c
```diff
@@ -1374,7 +1374,7 @@ void sub_806E994(void)
     gBattleTextBuff2[2] = gBankInMenu;
     gBattleTextBuff2[3] = pokemon_order_func(gBattlePartyID[gBankInMenu]);
     gBattleTextBuff2[4] = EOS;
-    StrCpyDecodeBattle(gText_PkmnsXPreventsSwitching, gStringVar4);
+     BattleStringExpandPlaceholders(gText_PkmnsXPreventsSwitching, gStringVar4);
```
ProjectRevoTPP: extra space

DizzyEggg: maybe it's shy and needs more space than other functions

DizzyEggg:
I was curious, did some math and battle_scripting_commands.c(battle_4 previously) has:
- 285 functions in total
- 5 non matching functions
- 2 fake matches functions

That's nice progress considering Ruby had around 20 non matches.

src/battle_script_commands.c
```diff
@@ -0,0 +1,11415 @@
...
+        gBattleMoveFlags |= MOVESTATUS_ENDURED;
+    }
+    else if (gSpecialStatuses[gBankTarget].focusBanded)
+    {
+        gBattleMoveFlags |= MOVESTATUS_HUNGON;
+        gLastUsedItem = gBattleMons[gBankTarget].item;
+    }
+
+    END:
+        gBa
```
ProjectRevoTPP: why not use < 40 so a define can be used for next ball level cap?

include/pokemon.h
```diff
@@ -620,6 +626,13 @@ bool8 IsPokeSpriteNotFlipped(u16 species);
 bool8 IsMonShiny(struct Pokemon *mon);
 bool8 IsShinyOtIdPersonality(u32 otId, u32 personality);
 
+void MonGainEVs(struct Pokemon *mon, u16 defeatedSpecies);
+bool8 IsTradedMon(struct Pokemon *mon);
+void HandleSetPokedexFlag(u16 nationalNum, u8 caseId, u32 personality);
+s32 sub_806D864(u16 a1);
+bool16 sub_806D82C(u8 id);
+u16 MonTryLearningNewMove(struct Pokemon* mon, bool8);
```
Diegoisawesome: These forward declarations (and the include directive) should be moved to the top of the file.

PR #46, opened 2017-09-24
Add Windows shortcut extension to gitignore

DizzyEggg:
.lnk

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.exe
+*.lnk
 *.o
 *.i
 *.elf


yenatch:
You could probably have every extension under the sun in .gitignore if you wanted. The repo .gitignore should just match extensions that it's responsible for.

To add stuff to gitignore globally, there is a global gitignore file. It seems to be different depending on the platform so run `git config --get core.excludesfile` to see where it is.

DizzyEggg:
Sure, I don't see any harm in adding this one though. I think most people keep a shortcut to an emulator, program or anything that helps document/label/see how things are in the rom. Or Im the only one. Hm...

Diegoisawesome:
I agree with yenatch on this one, this isn't really something that should be added to the project's .gitignore.

PR #47, opened 2017-09-25
Split data3

scnorton:


data/strings.s
```diff
@@ -50,6 +50,8 @@ gText_EggNickname:: @ 85E8264
 
 gText_Pokemon:: @ 85E8268
 	.string "POKéMON$"
+
+gUnknown_085e8270:: @ 85E8270
```
yenatch: Please use uppercase for unknowns.

PR #50, opened 2017-09-28
decompile roamer

DizzyEggg:
nothing really changed from ruby


src/roamer.c
```diff
@@ -0,0 +1,218 @@
+#include "global.h"
+#include "roamer.h"
+#include "pokemon.h"
+#include "rng.h"
+#include "species.h"
+#include "event_data.h"
+
+enum
+{
+    MAP_GRP = 0, // map group
```
ProjectRevoTPP: is = 0 really necessary?

DizzyEggg: I copied it from pokeruby.

data/rom_85CECB0.s
```diff
@@ -0,0 +1,154 @@
+	.include "asm/macros.inc"
```
Diegoisawesome: This file was split from rom_8158B30.s, which was named after its corresponding asm/ file. This should follow a similar naming scheme instead of naming it after the offset of the data it contains.

PR #56, opened 2017-09-30
Add Avara's map_constants.h

PikalaxALT:


Diegoisawesome:
I think you committed the wrong version.

PR #64, opened 2017-10-07
TV Engine

PikalaxALT:


asm/pokemon_summary_screen.s
```diff
@@ -6514,7 +6514,7 @@ sub_81C3220: @ 81C3220
 	b _081C3282
 	.pool
 _081C326C:
-	bl sub_80F0020
+	bl player_id_to_dword
```
DizzyEggg: Why use underscores in the function name? 
GetTrainerIdWord seems like a batter name.

include/battle.h
```diff
@@ -901,6 +902,10 @@ struct MonSpritesGfx
     struct SpriteTemplate templates[4];
 };
 
+extern struct BattleSpritesGfx* gBattleSpritesGfx;
+extern u8 gBattleOutcome;
```
DizzyEggg: gBattleSpritesGfx has been renamed to gMonSpritesGfx.

include/item.h
```diff
@@ -58,5 +58,6 @@ ItemUseFunc ItemId_GetFieldFunc(u16 itemId);
 u8 ItemId_GetBattleUsage(u16 itemId);
 ItemUseFunc ItemId_GetBattleFunc(u16 itemId);
 u8 ItemId_GetSecondaryId(u16 itemId);
+u16 itemid_get_market_price(u16 itemId);
```
DizzyEggg: Why use underscores?

include/battle.h
DizzyEggg: I'd rather not have any externed variables in that header and for now just extern any needed variable. Then as it's more organized, move the variables to a specific's file header.

include/moves.h
```diff
@@ -359,6 +359,7 @@ enum
     MOVE_DOOM_DESIRE,
     MOVE_PSYCHO_BOOST,
 };
+extern const u8 gMoveNames[][13];
```
DizzyEggg: Shouldn't it be externed in another header? I beilieve it's data2.h in pokeruby?

include/item.h
PikalaxALT: that one wasn't me actually

include/moves.h
PikalaxALT: 🤷‍♂️

data/event_scripts.s
```diff
@@ -822,7 +822,15 @@ gUnknown_0827EE09:: @ 27EE09
 	.incbin "baserom.gba", 0x27ee09, 0x2
 
 gUnknown_0827EE0B:: @ 827EE0B
-	.incbin "baserom.gba", 0x27ee0b, 0x118fa
+	.incbin "baserom.gba", 0x27ee0b, 0x46a
+
+    .include "data/text/tv.inc"
```
Diegoisawesome: Please standardize to tabs.

data/text/pokemon_news.inc
```diff
@@ -0,0 +1,165 @@
+
+gUnknown_0828D0F8:: @ 0828D0F8
```
Diegoisawesome: Why are these unnamed and separate from the other TV strings?

scnorton: 1) I haven't gotten around to naming them.
2) They're separated from the other TV strings by a size-0x90F incbin, same as in Ruby.

include/link.h
```diff
@@ -177,12 +177,13 @@ u32 LinkMain1(u8 *, u16 *, u16[CMD_LENGTH][MAX_LINK_PLAYERS]);
...
 void SerialCB(void);
+u8 GetLinkPlayerCount(void);
 
 void sub_800E0E8(void);
 bool8 sub_800A520(void);
 bool8 sub_8010500(void);
 void sub_800DFB4(u8, u8);
 void sub_800ADF8(void);
-u8 sub_8009FCC(void);
+u8 GetLinkPlayerCount(void);
```
Diegoisawesome: This shouldn't be in here twice.

include/moves.h
Diegoisawesome: In decompiling scrcmd I've made that change; when merging that in I'll take this out.

tvtextgroups.py
```diff
@@ -0,0 +1,30 @@
+import subprocess
```
Diegoisawesome: Don't think this file was meant to be in here.

Diegoisawesome:
`c:\devkitPro\devkitARM\bin\arm-none-eabi-ld.exe: Warning: alignment 4 of symbol `gTVStringVarPtrs' in src/tv.o is smaller than 16 in src/battle_2.o
c:\devkitPro\devkitARM\bin\arm-none-eabi-ld.exe: Warning: alignment 4 of symbol `gTVStringVarPtrs' in src/tv.o is smaller than 16 in src/new_game.o`

PR #68, opened 2017-10-08
decompile diploma.c

ProjectRevoTPP:


src/diploma.c
```diff
@@ -0,0 +1,218 @@
...
+        SetGpuReg(REG_OFFSET_BG1HOFS, DISPCNT_MODE_0);
+        StringCopy(gStringVar1, gText_DexHoenn);
+    }
+    StringExpandPlaceholders(gStringVar4, gText_PokedexDiploma);
+    PrintDiplomaText(gStringVar4, 0, 1);
+    PutWindowTilemap(0);
+    CopyWindowToVram(0, 3);
+}
+
+static const struct BgTemplate gDiplomaBgTemplates[2] =
```
DizzyEggg: Didn't we all agree to put const data at the top before the actual code?

ProjectRevoTPP: I have changed my style to do it before the code. I find it much easier to make it line up when its stated before the first function of use.

ProjectRevoTPP: PS: will update save failed screen to comply.

PR #73, opened 2017-10-09
Add diploma header and fix static names 

DizzyEggg:


include/diploma.h
```diff
@@ -0,0 +1,6 @@
+#ifndef GUARD_DIPLOMA_H
+#define GUARD_DIPLOMA_H
+
+void CB2_ShowDiploma(void);
```
Diegoisawesome: Why is a header needed? It's only one function, called from one file.

DizzyEggg: Isn't that enough reason?

PR #75, opened 2017-10-09
Decompile reshow battle screen

DizzyEggg:


data/data2b.s
```diff
@@ -4509,7 +4510,7 @@ gUnknown_08305D0C:: @ 8305D0C
 @ 8305DCC
 	.include "data/enemy_mon_elevation.inc"
 
-@ 8305F68
+@ 8309AAC
```
Diegoisawesome: Why did you change this? It's definitely not located there.

src/reshow_battle_screen.c
```diff
@@ -0,0 +1,365 @@
...
+        SetMainCallback2(BattleMainCB2);
+        sub_805EF14();
+        break;
+    }
+
+    gBattleScripting.reshowMainState++;
+}
+
+// todo: find a better place for the struct declaration
+struct BGCntrlBitfield
```
Diegoisawesome: bg.h would make sense for this struct.

data/data2b.s
DizzyEggg: fixed

src/reshow_battle_screen.c
DizzyEggg: moved

PR #77, opened 2017-10-12
ABE

DizzyEggg:


src/battle_util.c
```diff
@@ -2176,9 +2167,9 @@ u8 AbilityBattleEffects(u8 caseID, u8 bank, u8 ability, u8 special, u16 moveArg)
                     if (gBattleMons[bank].maxHP == gBattleMons[bank].hp)
                     {
                         if ((gProtectStructs[gBankAttacker].notFirstStrike))
-                            gBattlescriptCurrInstr = BattleScript_MoveHPDrain;
+                            gBattlescriptCurrInstr = gUnknown_082DB592;
                         else
-                            gBattlescriptCurrInstr = BattleScript_MoveHPDrain_PPLoss;
+                            gBattlescriptCurrInstr = gUnknown_082DB591;
```
scnorton: The descriptive names should be restored.

src/battle_util.c
```diff
@@ -2478,15 +2457,16 @@ u8 AbilityBattleEffects(u8 caseID, u8 bank, u8 ability, u8 special, u16 moveArg)
             {
                 if (gBattleMons[i].ability == ABILITY_TRACE && (gStatuses3[i] & STATUS3_TRACE))
                 {
-                    u8 opposite = (GetBankIdentity(i) ^ 1) & 1;
-                    u8 target1 = GetBankByIdentity(opposite);
-                    u8 target2 = GetBankByIdentity(opposite + 2);
+                    u8 target2;
```
scnorton: Does this still match if these variables are declared at the top of the function rather than mid-function?

src/battle_util.c
DizzyEggg: It's a different battlescript.

src/battle_util.c
DizzyEggg: Yeah

src/battle_util.c
scnorton: Gotcha.

src/battle_util.c
```diff
@@ -2021,18 +2017,15 @@ u8 AbilityBattleEffects(u8 caseID, u8 bank, u8 ability, u8 special, u16 moveArg)
...
-                        //asm("":::"r4","r5");
-                        effect = CastformDataTypeChange(i);
+                        effect = CastformDataTypeChange(target1);
                         if (effect != 0)
                         {
                             BattleScriptPushCursorAndCallback(BattleScript_CastformChange);
-                            gBattleScripting.bank = i;
-                            gBattleStruct->formToChangeInto = effect - 1;
+                            gBattleScripting.bank = target1;
+                            *(&gBattleStruct->formToChangeInto) = effect - 1;
```
Diegoisawesome: Is making it into a pointer and dereferencing necessary?

DizzyEggg: Just check out other battle files. It's a leftover from R/S direct ewram usage and everything being a u8 array;

PR #81, opened 2017-10-13
Decompile rom_3

DizzyEggg:
Renamed "rom_3" to "battle_controllers".


src/battle_controllers.c
```diff
@@ -0,0 +1,1551 @@
...
+        if (gTasks[taskId].data[13] == 0)
+        {
+            gTasks[taskId].data[13] = 1;
+            gTasks[taskId].data[11] = 3;
+        }
+        break;
+    }
+}
+
+// fix me
```
Diegoisawesome: Did you forget to fix this?

DizzyEggg: Nope, I was unable to just as as the person who decompiled this function in Ruby. It's a fake match and the comment is there, so maybe someone fixes it in the future.

src/battle_controllers.c
```diff
@@ -0,0 +1,1551 @@
...
+    PrepareBufferDataTransfer(bufferId, gBattleBuffersTransferData, 4);
+}
+
+void EmitChooseMove(u8 bufferId, bool8 isDoubleBattle, bool8 NoPpNumber, struct ChooseMoveStruct *movePpData)
+{
+    s32 i;
+
+    gBattleBuffersTransferData[0] = CONTROLLER_CHOOSEMOVE;
+    gBattleBuffersTransferData[1] = isDoubleBattle;
+    gBattleBuffersTransferData[2] = NoPpNumber;
```
Diegoisawesome: What does "NoPpNumber" mean exactly?

DizzyEggg: It means it won't display a pp number when you select a move. In the games it's always set to FALSE.

src/battle_controllers.c
```diff
@@ -0,0 +1,1551 @@
...
+        gBattleBuffersTransferData[1 + i] = arg1[i];
+    PrepareBufferDataTransfer(bufferId, gBattleBuffersTransferData, 4);
+}
+
+void EmitChoosePokemon(u8 bufferId, u8 caseId, u8 arg2, u8 abilityId, u8* arg4)
+{
+    s32 i;
+
+    gBattleBuffersTransferData[0] = CONTROLLER_CHOOSEPOKEMON;
+    gBattleBuffersTransferData[1] = caseId;
```
Diegoisawesome: And what does "caseId" mean here?

DizzyEggg: It means open the pokemon move that:
0 - can't switch, because ingrain
1 - can't switch, because ability
2 - tag battles summary
3 - double battles summary,
(note: the numbers I just came up with them, but it's sth like that)

src/battle_controllers.c
```diff
@@ -0,0 +1,1551 @@
...
+{
+    gBattleBuffersTransferData[0] = CONTROLLER_38;
+    gBattleBuffersTransferData[1] = b;
+    PrepareBufferDataTransfer(bufferId, gBattleBuffersTransferData, 2);
+}
+
+void EmitCmd39(u8 bufferId)
+{
+    gBattleBuffersTransferData[0] = CONTROLLER_39;
+    gBattleBuffersTransferData[1] = 39;
```
Diegoisawesome: Shouldn't these just be use CONTROLLER_39 enum? (Same goes for all instances where the value is repeated four times, seemingly as filler)

DizzyEggg: Yeah, it's possible they used filler to align the buffer to at least 4 bytes. I wanted to check if the seemingly filler bytes are used by any controllers first tho

src/battle_controllers.c
```diff
@@ -0,0 +1,1551 @@
...
+void EmitEffectivenessSound(u8 bufferId, u16 songId)
+{
+    gBattleBuffersTransferData[0] = CONTROLLER_EFFECTIVENESSSOUND;
+    gBattleBuffersTransferData[1] = songId;
+    gBattleBuffersTransferData[2] = (songId & 0xFF00) >> 8;
+    gBattleBuffersTransferData[3] = 0;
+    PrepareBufferDataTransfer(bufferId, gBattleBuffersTransferData, 4);
+}
+
+void EmitPlayFanfareOrBGM(u8 bufferId, u16 songId, bool8 playBGM)
```
Diegoisawesome: Shouldn't this be EmitPlayFanfareOrSE? Is this used for playing actual music or just sound effects and fanfares?

src/battle_controllers.c
Diegoisawesome: It seems obvious to me that these are just repeats of the enum value, regardless of their final usage, so it would make sense to me to make those also use the value from the enum.

src/battle_controllers.c
```diff
@@ -0,0 +1,1551 @@
...
+extern u8 gBattleOutcome;
+extern u8 gLastUsedAbility;
+extern u8 gStringBank;
+
+extern const struct BattleMove gBattleMoves[];
+
+extern void task00_08081A90(u8 taskId); // cable_club
+extern void sub_81B8D64(u8 bank, u8 arg1); // party_menu
+
+// this file's funcionts
```
Diegoisawesome: "this file's functions"

src/battle_controllers.c
DizzyEggg: No, it either play a fanfare or BGM

PR #82, opened 2017-10-13
Decompile functions for reading mail

scnorton:


src/mail.c
```diff
@@ -0,0 +1,577 @@
...
+    const u8 *tileMap;
+    u16 var0C;
+    u16 var0E;
+    u16 color10;
+    u16 color12;
+};
+
+// Static RAM declarations
+
+static EWRAM_DATA struct
```
Diegoisawesome: This struct should be declared in a header elsewhere.

PikalaxALT: I disagree.  There is exactly one object of this type anywhere in the code, and it's here.

PR #85, opened 2017-10-14
Decompile battle_anim.s

DizzyEggg:


asm/battle_1.s
```diff
@@ -1424,8 +1424,8 @@ _0803575A:
 	.pool
 	thumb_func_end ApplyPlayerChosenFrameToBattleMenu
 
-	thumb_func_start load_battle_oval_graphics
-load_battle_oval_graphics: @ 8035770
+	thumb_func_start LoadFittingBackgroundForBattle
```
Diegoisawesome: What's meant by "fitting" here?

DizzyEggg: If battle flag groudon -> loads groudon background
if battle flag kyogre -> loads kyogre background
same with rayquaza
then checks if the trainer class is leader/champion
then either loads it from gBattleTerrain or one that fits the map background

asm/battle_anim_80FE840.s
```diff
@@ -727,30 +727,30 @@ sub_80FEE1C: @ 80FEE1C
...
 	adds r0, r4, 0
-	bl move_anim_task_del
+	bl DestroyAnimVisualTask
 	b _080FEE70
 _080FEE3E:
-	ldr r0, =gAnimationBankTarget
+	ldr r0, =gAnimBankTarget
 	ldrb r0, [r0]
-	bl GetBankIdentity_permutated
+	bl sub_80A8364
```
Diegoisawesome: Why was this renamed from an actual name to a sub_?

DizzyEggg: Because the name was wrong. I think it gets a background id but certainly not a bank.

asm/battle_1.s
Diegoisawesome: So then seems like it should just be called "LoadBackgroundForBattle".

DizzyEggg: Makes sense I guess.

src/battle_anim.c
```diff
@@ -0,0 +1,1852 @@
...
+        gBattle_BG2_Y = 0;
+    }
+}
+
+static void task_pA_ma0A_obj_to_bg_pal(u8 taskId)
+{
+    u8 r4;
+    u8 r6;
+    s16 r3;
+    s16 r2;
```
Diegoisawesome: You should name these vars more descriptively.

DizzyEggg: copied from pokeruby, I know too little what the function does to give them better names

src/battle_anim.c
```diff
@@ -0,0 +1,1852 @@
...
+#undef t2_MONBG_BANK
+
+static void ScriptCmd_setalpha(void)
+{
+    u16 half1, half2;
+
+    sBattleAnimScriptPtr++;
+    half1 = *(sBattleAnimScriptPtr++);
+    half2 = *(sBattleAnimScriptPtr++) << 8;
+    SetGpuReg(REG_OFFSET_BLDCNT, 0x3F40);
```
Diegoisawesome: Consider adding some definition to io_reg.h so you're not just setting an opaque value.

src/battle_anim.c
Diegoisawesome: Seems pretty clear to me that r3 is X, r2 is Y, r6 is palette index, and r4 is sprite index.

src/battle_anim.c
```diff
@@ -0,0 +1,1852 @@
...
+}
+
+s8 BattleAnimAdjustPanning(s8 pan)
+{
+    if (!IsContest() && gBattleSpritesDataPtr->healthBoxesData[gAnimBankAttacker].flag_x10)
+    {
+        if (GetBankSide(gAnimBankAttacker) != SIDE_PLAYER)
+            pan = 63;
+        else
+            pan = -64;
```
Diegoisawesome: These pan values should be defined in a header.

PR #88, opened 2017-10-16
decompile battle interface

DizzyEggg:


ProjectRevoTPP:
Is it just as mathy and spaghetti code as it was in ruby?

DizzyEggg:
Do you mean the first function? So far everything except that has been rather fine. The first function however I've no idea what does. It's also unused, so I don't really care.

src/battle_interface.c
```diff
@@ -0,0 +1,2641 @@
...
+    {ARRAY_COUNT(sStatusSummaryBar_Subsprites_0), sStatusSummaryBar_Subsprites_0}
+};
+
+static const struct SubspriteTable gUnknown_0832C2CC[] =
+{
+    {ARRAY_COUNT(gUnknown_0832C2AC), gUnknown_0832C2AC}
+};
+
+// unused unknown data
+static const u16 gUnknown_0832C2D4[] =
```
Diegoisawesome: Appears to be some kind of 4BPP graphics data.

src/battle_interface.c
```diff
@@ -0,0 +1,2641 @@
...
+    .x = 0,
+    .matrixNum = 0,
+    .size = 3,
+    .tileNum = 0,
+    .priority = 1,
+    .paletteNum = 0,
+    .affineParam = 0,
+};
+
+static const struct OamData sOamData_StatusSummayBalls =
```
Diegoisawesome: "Summary"

src/battle_interface.c
```diff
@@ -0,0 +1,2641 @@
...
+enum
+{
+    PAL_STATUS_PSN,
+    PAL_STATUS_PAR,
+    PAL_STATUS_SLP,
+    PAL_STATUS_FRZ,
+    PAL_STATUS_BRN
+};
+
+static const u16 sStatusIconPalettes[] =
```
Diegoisawesome: Should be extracted as a palette file.

src/battle_interface.c
```diff
@@ -0,0 +1,2641 @@
...
+    else
+    {
+        sprite->pos2.x -= var2 >> 4;
+        if (sprite->pos2.x < 0)
+            sprite->pos2.x = 0;
+    }
+
+    if (sprite->pos2.x == 0)
+    {
+        pan = 63;
```
Diegoisawesome: These should be changed to use the defines from the last PR.

PR #89, opened 2017-10-17
Decoration + decoration_inventory

PikalaxALT:


include/decoration.h
```diff
@@ -125,6 +125,47 @@ enum DecoId {
...
+    DECORSHAPE_4x2,
+    DECORSHAPE_2x2,
+    DECORSHAPE_1x2,
+    DECORSHAPE_1x3, // unused
+    DECORSHAPE_2x4,
+    DECORSHAPE_3x3,
+    DECORSHAPE_3x2
+};
+
+enum DecoCat {
```
Diegoisawesome: Should be named DecorCat (or better, DecorationCategory) to keep consistent with the values of the enum.

include/decoration.h
```diff
@@ -137,6 +178,18 @@ struct Decoration
     /*0x1c*/ const u16 *tiles;
 };
 
+struct DecoPCPointers
```
Diegoisawesome: This should be renamed DecorPCPointers or DecorationPCPointers, as the usual shorthand for decoration is decor.

PR #97, opened 2017-10-28
Secret base

PikalaxALT:


src/secret_base.c
```diff
@@ -0,0 +1,2272 @@
...
+            {
+                data[1] = 1;
+            }
+            break;
+    }
+}
+
+void sub_80EA828(u8 sbId, struct SecretBaseRecord *base, u32 version, u32 language)
+{
+    int strlen;
```
Diegoisawesome: strlen->stringLength

src/secret_base.c
```diff
@@ -0,0 +1,2272 @@
...
+        {
+            return i;
+        }
+    }
+    return 0;
+}
+
+u8 sub_80EAA64(struct SecretBaseRecord *base, u32 version, u32 language)
+{
+    s16 sbId;
```
Diegoisawesome: sbId->secretBaseID

src/secret_base.c
```diff
@@ -0,0 +1,2272 @@
...
+    if (VarGet(VAR_0x40F0) && gUnknown_0203A01D == TRUE && !CurrentMapIsSecretBase())
+    {
+        VarSet(VAR_0x40F0, FALSE);
+        gUnknown_0203A01D = FALSE;
+        sub_80EEA70();
+        VarSet(VAR_0x40EC, 0);
+        VarSet(VAR_0x40ED, 0);
+        VarSet(VAR_0x40EE, 0);
+        VarSet(VAR_0x40EF, 0);
+        VarSet(VAR_0x40F0, FALSE);
```
Diegoisawesome: Why are some of these set to 0 and some set to FALSE? This should be consistent.

PikalaxALT: I'll double-check, but it's intended to reflect the difference between whether the var is ised as a u16 or as a bool16.

src/secret_base.c
```diff
@@ -0,0 +1,2272 @@
...
+void sub_80E933C(void)
+{
+    u8 *roomdecor;
+    u8 *roomdecorpos;
+    u8 decidx;
+    u8 objid;
+    u8 metatile;
+    u8 category;
+    u8 permission;
+    u8 ndecor;
```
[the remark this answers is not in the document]
PikalaxALT: ndecor, sure; but the rest are fine as is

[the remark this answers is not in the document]
Diegoisawesome:
```u8 *roomdecor;
u8 *roomdecorpos;
u8 decidx;
u8 objid;```
They should also be camelCase. I added the comment to the bottom of the declarations and meant for it to include all of them.

PR #98, opened 2017-10-28
Decompile bard music

PikalaxALT:


src/bard_music.c
```diff
@@ -0,0 +1,65 @@
...
+    return gBardSoundPitchTables[arg0][songPos];
+}
+
+const struct BardSound *GetWordSounds(u16 word)
+{
+    u32 category;
+    u32 subword;
+    const struct BardSound (*ptr)[6];
+
+    if (IsECWordInValidRange(word))
```
Diegoisawesome: This should be renamed to reflect the fact it checks if it's invalid.

PR #103, opened 2017-10-30
update build instructions for linux.

ConnorRigby:
I just copied this from the `pokeruby` repository. Don't know if it is the correct solution, but it compiled correctly on linux. Don't have a windows machine to check windows.


Uyuiyu:
Since "tools" is developed in the other repository, add it as a submodule

yenatch:
Submodules don't work as build dependencies (see pokered/pokecrystal). They should be included in the repo somehow, but I'm coming up short besides copying them (which maybe is still preferable, since the tools have stabilized a bit).

Diegoisawesome:
Submodules are checked out when the master repo is checked out, at least with recent versions of Git.

camthesaxman:
I suggest just splitting the tools into a separate repo, and we can distribute binaries via GitHub releases.

Diegoisawesome:
So what's the consensus on this?

ConnorRigby:
Sorry all, i forgot about this Pull Request. What i did was take PokeRuby as an example. A submodule will work just the same. As will distributing binaries via github releases..

camthesaxman:
I'd rather not copy the tools into pokeemerald. If we're going to fix this, someone needs to take the tools from pokeruby and put them into a separate repository. I would have already done this if I had permissions to create repos in pret. Using a submodule is not necessary. A simple install.sh script just like we have in agbcc would suffice.

yenatch:
That's even worse than a submodule.

PR #113, opened 2017-11-12
Start Pokemon Storage System

DizzyEggg:
I wanted to take on this file, but came across an unused function refusing to match and got discouraged. So yeah, ain't gonna work on it for some time, switching to battle stuff instead. Still, I decompiled a few functions, so it's always something.


Diegoisawesome:
Seems okay to merge but I'm going to try my hand at matching that function before I do.

PR #118, opened 2017-11-23
Decompile field poison

PikalaxALT:


src/field_poison.c
```diff
@@ -0,0 +1,157 @@
...
+    CreateTask(sub_80F9654, 80);
+    ScriptContext1_Stop();
+}
+
+unsigned overworld_poison(void)
+{
+    int i;
+    unsigned cnt1;
+    unsigned cnt2;
+    unsigned hp;
```
Diegoisawesome: These 'unsigned' types should be called something more complete, like "unsigned int" or "u32", etc.

PR #124, opened 2017-11-25
Dump Battle Scripts

DizzyEggg:


asm/macros/battle_script.inc
```diff
@@ -0,0 +1,1280 @@
...
+
+	.macro useitemonopponent
+	.byte 0x75
+	.endm
+
+	.macro various bank, param1
+	.byte 0x76
+	.byte \bank
+	.byte \param1
+	.endm
```
yenatch: there should be macros for the `VARIOUS` constants instead of using them directly:
```asm
	.macro cancel_multi_turn_moves bank
	.byte 0x76
	.byte \bank
	.byte VARIOUS_CANCEL_MULTI_TURN_MOVES
	.endm
```

PR #128, opened 2017-11-28
Decompile battle setup

DizzyEggg:


src/battle_setup.c
```diff
@@ -0,0 +1,586 @@
...
+}
+
+void StartGroudonKyogreBattle(void)
+{
+    ScriptContext2_Enable();
+    gMain.savedCallback = CB2_EndScriptedWildBattle;
+    gBattleTypeFlags = BATTLE_TYPE_LEGENDARY | BATTLE_TYPE_KYOGRE_GROUDON;
+
+    if (gGameVersion == VERSION_RUBY)
+        CreateBattleStartTask(B_TRANSITION_SHARDS, BGM_BATTLE34); // GROUDON
```
ProjectRevoTPP: lol that this is still in emerald

include/trainer_ids.h
```diff
@@ -2,6 +2,11 @@
 #define GUARD_TRAINER_IDS_H
 
 #define NO_OF_TRAINERS      854
+
+#define TRAINER_ID_SIDNEY   261
```
camthesaxman: Can you please match the defines that are in constants/trainer_constants.inc? All of those are defined in there already, and you just need to convert that .inc file into a C header. It will make things a whole lot easier in the near future when I make the .s files use the C preprocessor.

DizzyEggg: Are the defines in trainer constants inc correct? If so, then sure, but if they're copied from Ruby I'd rather not do it, it's boring work. :P

camthesaxman: Actually, these were created before pokeruby existed, so they should be correct. Just use a regex to automate things and make it easy.

include/trainer.h
```diff
@@ -0,0 +1,19 @@
+#ifndef GUARD_TRAINER_H
+#define GUARD_TRAINER_H
+
+#define TRAINER_ENCOUNTER_MUSIC_MALE         0 // standard male encounter music
```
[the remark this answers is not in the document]
DizzyEggg: taken from pokeruby

src/trainer_see.c
```diff
@@ -0,0 +1,68 @@
...
+        u8 retVal;
+
+        if (!gMapObjects[i].active)
+            continue;
+        if (gMapObjects[i].trainerType != 1 && gMapObjects[i].trainerType != 3)
+            continue;
+
+        retVal = CheckIfTrainerWantsBattle(i);
+        if (retVal == 2)
+            break; // two trainers has been found
```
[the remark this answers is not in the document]
DizzyEggg: shame on me

sym_ewram.txt
```diff
@@ -706,55 +706,11 @@ gUnknown_02038BC4: @ 2038BC4
...
-
-gUnknown_02038BF8: @ 2038BF8
-	.space 0x1
-
-gUnknown_02038BF9: @ 2038BF9
-	.space 0x3
+.align 2
 
-gUnknown_02038BFC: @ 2038BFC
+gApproachingTrainerId: @ 2038BFC
```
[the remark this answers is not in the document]
DizzyEggg: thinking more of trainer see, once the file is decompiled, we're going to choose where it fits better

PR #130, opened 2017-12-02
Decompile pokeball

DizzyEggg:


graphics/interface/ball/dive.pal
```diff
@@ -0,0 +1,19 @@
+JASC-PAL
```
camthesaxman: You don't actually need these .pal files, since the .gbapal can be automatically extracted from dive.png.

src/sprite.c
```diff
@@ -168,14 +168,7 @@ static const struct Sprite sDummySprite =
...
     .animLoopCounter = 0,
-    .data0 = 0,
-    .data1 = 0,
-    .data2 = 0,
-    .data3 = 0,
-    .data4 = 0,
-    .data5 = 0,
-    .data6 = 0,
-    .data7 = 0,
+    .data = {0, 0, 0, 0, 0, 0, 0},
```
camthesaxman: You can just do `{0}` here. It will initalize all elements to 0.

PR #139, opened 2017-12-17
Decompile and split battle_1

DizzyEggg:
I renamed battle_1 to `battle_bg`.


src/rom_8034C54.c
```diff
@@ -1,17 +1,5 @@
-
-// Includes
 #include "global.h"
 
-// Static type declarations
-
-// Static RAM declarations
-
 IWRAM_DATA u32 gUnknown_03000DD4;
```
camthesaxman: Any idea which file these belong to?

DizzyEggg: rom_8034C54.s which is an unknown link file.

PR #145, opened 2017-12-21
Port generic event script macros from pokeruby

huderlem:
Also changed `choosecontestpkmn` to `choosecontestmon` to match naming convention.


asm/macros/event.inc
```diff
@@ -283,8 +295,8 @@
 	@ Initializes the RTC`s local time offset to the given hour and minute. In FireRed, this command is a nop.
 	.macro initclock hour, minute
 	.byte 0x2c
-	.2byte \hour
-	.2byte \minute
+	.byte \hour
+	.byte \minute
```
Diegoisawesome: These are actually 2bytes, why did you change them?

PR #151, opened 2017-12-31
Dump battle anim scripts

DizzyEggg:


src/battle_anim.c
```diff
@@ -67,7 +68,7 @@ static void ScriptCmd_return(void);
 static void ScriptCmd_setarg(void);
 static void ScriptCmd_choosetwoturnanim(void);
 static void ScriptCmd_jumpifmoveturn(void);
-static void ScriptCmd_jump(void);
+static void ScriptCmd_goto(void);
```
Diegoisawesome: Since jump was renamed to goto, jumpifmoveturn should be renamed too.

data/item_menu_icons.s
```diff
@@ -1,4 +1,4 @@
-#include "constants/items.h"
+	#include "constants/items.h"
```
Diegoisawesome: Remove indent

data/battle_scripts_1.s
```diff
@@ -10,6 +10,7 @@
 
 	.section script_data, "aw", %progbits
 
+.align 2
```
Diegoisawesome: Add indent

PR #153, opened 2018-01-01
Decompile Battle Transition

DizzyEggg:
First 15 transitions are basically the same, the champion one is different(Steven/Wallace) and there's a lot of new ones (17-41).


src/battle_transition.c
```diff
@@ -0,0 +1,4346 @@
...
+#include "decompress.h"
+#include "bg.h"
+#include "blend_palette.h"
+
+struct TransitionData
+{
+    vs8 VBlank_DMA;
+    u16 WININ;
+    u16 WINOUT;
+    u16 field_6;
```
Diegoisawesome: Wouldn't this be WIN0H?

DizzyEggg: seems like it

src/battle_transition.c
```diff
@@ -0,0 +1,4346 @@
...
+    u16 field_A;
+    u16 field_C;
+    u16 BLDCNT;
+    u16 BLDALPHA;
+    u16 BLDY;
+    s16 field_14;
+    s16 field_16;
+    s16 field_18;
+    s16 field_1A;
+    s16 field_1C;
```
Diegoisawesome: Should be BG0VOFS

src/battle_transition.c
```diff
@@ -0,0 +1,4346 @@
...
+{
+    u8 i, r5;
+    u16* toStore;
+    s16 value;
+    s32 mergedValue;
+
+    sTransitionStructPtr->VBlank_DMA = FALSE;
+
+    toStore = gUnknown_02038C28[0];
+    r5 = task->tData1;
```
Diegoisawesome: Those fields should be named something related to BG0HOFS

Diegoisawesome: Does *= 2 work? If so, use it as it's more legible

Diegoisawesome: Does i & 1 work?

DizzyEggg: any suggestions?

Diegoisawesome: 2742: Were you unable to fix this?

Diegoisawesome: Maybe BG0HOFS_1 and _2? Nothing too complicated

DizzyEggg: huh, fix what? That was on gf's part :D

DizzyEggg: yup, changed it

DizzyEggg: nope, it must be lsl

Diegoisawesome: The compiler seriously turns *= 2 into a mul command? Usually it's smart enough to take multiplication by powers of two and make them lsl

DizzyEggg: it messes up registers

src/battle_transition.c
```diff
@@ -0,0 +1,4346 @@
...
+    u16 BLDCNT;
+    u16 BLDALPHA;
+    u16 BLDY;
+    s16 field_14;
+    s16 field_16;
+    s16 field_18;
+    s16 field_1A;
+    s16 field_1C;
+    s16 field_1E; // unused
+    s16 field_20;
```
Diegoisawesome: Should probably be a bool16

src/battle_transition.c
DizzyEggg: done

src/battle_transition.c
DizzyEggg: done

PR #178, opened 2018-01-21
Use standard flags to link libgcc, libc

PikalaxALT:


include/global.h
```diff
@@ -1,6 +1,7 @@
 #ifndef GUARD_GLOBAL_H
 #define GUARD_GLOBAL_H
 
+#include <string.h>
```
Diegoisawesome: I'm not sure about adding in a library header to this... Maybe an approach like pokeruby would make more sense.

PikalaxALT: 1) The library header is taken from agbcc (-I tools/agbcc/include).
2) The approach in pokeruby is being similarly deprecated.

PR #180, opened 2018-01-21
Decompile mystery event script

DizzyEggg:


asm/macros/event.inc
```diff
@@ -262,6 +262,13 @@
 	.2byte \output
 	.2byte SPECIAL_\function
 	.endm
+	
+	@ temporary solution
+	.macro specialvar_ output, functionId
```
Diegoisawesome: What's this for?

DizzyEggg: oh, it was something I tried to do before the data was dumped and forgot to remove it

src/mystery_event_script.c
```diff
@@ -0,0 +1,396 @@
...
+    u8 value = ScriptReadByte(ctx);
+    ctx->data[2] = value;
+    return FALSE;
+}
+
+bool8 MEScrCmd_setmsg(struct ScriptContext *ctx)
+{
+    u8 value = ScriptReadByte(ctx);
+    u8 *str = (u8 *)(ScriptReadWord(ctx) - ctx->data[1] + ctx->data[0]);
+    if (value == 255 || value == ctx->data[2])
```
Diegoisawesome: 255->-1

src/mystery_event_script.c
```diff
@@ -0,0 +1,396 @@
...
+    *a0 = sMysteryEventScriptContext.data[2];
+
+    return ret;
+}
+
+u32 RunMysteryEventScript(u8 *script)
+{
+    struct ScriptContext *ctx = &sMysteryEventScriptContext;
+    InitMysteryEventScript(ctx, script);
+    while (RunMysteryEventScriptCommand(ctx))
```
Diegoisawesome: Convention so far has been to keep the semicolon on the same line as the while statement, but visually separate the while statement from the surrounding lines.

PR #184, opened 2018-01-22
decompilie pokemon animation

DizzyEggg:


Diegoisawesome:
[898] Absolutely sure a single cast won't work here?

PR #200, opened 2018-02-04
decompile title screen

DizzyEggg:
Similar to the ruby's one but better.


src/title_screen.c
```diff
@@ -0,0 +1,783 @@
...
+#include "sprite.h"
+#include "task.h"
+#include "scanline_effect.h"
+#include "gpu_regs.h"
+#include "trig.h"
+#include "constants/rgb.h"
+
+#define VERSION_BANNER_SHAPE 1
+#define VERSION_BANNER_RIGHT_TILEOFFSET 64
+#define VERSION_BANNER_BYTES 0x1000
```
Diegoisawesome: Why did this need to be a define?

DizzyEggg: To keep in sync with pokeruby where this is a define.

PR #228, opened 2018-03-04
Decompile item menu icons

DizzyEggg:
Another nonmatching bites the dust


src/item_menu_icons.c
```diff
@@ -0,0 +1,676 @@
...
+            mem1 += 0x20;
+    }
+}
+
+static void sub_80D5070(u8 berryId)
+{
+    struct CompressedSpritePalette pal;
+
+    if (berryId == ITEM_TO_BERRY(ITEM_ENIGMA_BERRY) - 1)
+        IsEnigmaBerryValid(); // what's the point of calling it if the return value is ignored?
```
Diegoisawesome: Try and see if an empty if statement block works here, as it may have gotten optimized out

PR #232, opened 2018-03-28
Start menu

HrPeanut:
I have decompiled start_menu.s and renamed some functions. 
Building with `make compare` results in `pokeemerald.gba: OK`.


src/start_menu.c
```diff
@@ -224,82 +363,1069 @@ static void BuildStartMenuActions_BattlePyramid(void)
...
+        EnableBothScriptContexts();
+    }
+}
+
+void sub_80A08CC(void) // Referenced in data/specials.inc and data/scripts/maps/BattleFrontier_BattleTowerLobby.inc
+{
+    u32 taskId1;
+    u32 taskId2;
+
+    taskId1 = CreateTask(&sub_8153688, 0x5);
```
ProjectRevoTPP: Try ditching the temporary variables, taskId temps are basically never given u32 typing, so it was likely a direct CreateTask call assignment.

HrPeanut: I originally had it as direct calls and I do not remember why I changed it. I think I  did it because it is easier to read. I also do not remember if I tried using `u8` or `u16`.

PR #233, opened 2018-03-28
[WIP] Learn move

Edwearth:
Decompilation of `learn_move.s`


src/learn_move.c
```diff
@@ -0,0 +1,84 @@
...
+  //*partyMon = r1;
+  gUnknown_0203BC34 = AllocZeroed(0x118);//gUnknown_0203BC34 is a global pointer to LearnMoveStruct ?
+  gUnknown_0203BC34->partyMon = gSpecialVar_0x8004;
+
+  SetVBlankCallback(sub_8160624);
+  sub_81607EC();
+  sub_81D2824(0);
+
+  //gUnknown_0203BC38 ?
+  (&gUnknown_0203BC38)[0] = 0;
```
ProjectRevoTPP: this is pretty odd C. are you sure its not just an array?

Edwearth: I don't know what `gUnknown_0203BC38` is. I just tried to translate the asm. It will probably be changed in the future.

ProjectRevoTPP: It's not a literal translation. You write what its doing on a high level via your own interpretation, and then change definitions as necessary to get the output to match the ROM.

Edwearth: Do you think it is a good thing to change it to static array?  It could be a struct, but I don't know yet.

src/learn_move.c
```diff
@@ -105,15 +128,42 @@ void sub_8160664(u8 taskId)
...
+  ShowBg(1);
+  SetGpuReg(0x50, 0);
+}
+void sub_816082C(void)
+{
+  //FIXME
+  //sub_8160868();
+  //void(*sub_8160868)(void) = (void (*)(void))0x08160868;
+  //sub_8160868();
+  __asm__("bl sub_8160868");
```
ProjectRevoTPP: No. Please dont do this. Just call it. It will generate the correct code.

Edwearth: But the function is not yet decompiled. I will replace it.

ProjectRevoTPP: You can just extern it. It will work.

PR #240, opened 2018-04-25
Move gItems to c

Slawter666:


include/data/items.h
```diff
@@ -1,6460 +1,6079 @@
...
-		.fieldUseFunc = ItemUseOutOfBattle_CannotUse,
-		.battleUsage = 0,
-		.battleUseFunc = NULL,
-		.secondaryId = 0,
-	},
-	
-	//Pokeballs
-	
-	[ITEM_MASTER_BALL] = {
-		.name = _("MASTER BALL$"),
```
ketsuban:
```
[21:15] ProjectRevoTPP: dont add a $
[21:15] ProjectRevoTPP: the string builder does that
[21:15] ProjectRevoTPP: const u8 myString[] = _("Foo");
```

Did this turn out not to be the case?

ProjectRevoTPP:
```
+//Pokeballs
+
+    {
+        .name = _("MASTER BALL"),
```

I don't see a $.

ketsuban: Disregard this I'm an idiot, I got confused between remove and add.

huderlem:
I'm not in favor of the most recent commit "Convert item_graphics.inc to C".  It doesn't add much value to move that chunk into its own .c file--especially if that file would live in the top level `src/`.  Perhaps in the future, we could do a pass on `graphics.s` and decide if anything should be done about it, but I think we should avoid it at this point in time..

Slawter666:
I think that, for ease of editing `graphics.s` should be split, although it seems lots of battle animation graphics are intertwined with other graphics so I'm not sure where to draw the boundaries.

That can be for a future PR though, I have reverted my latest commit.

include/data/item_icon_table.h
```diff
@@ -0,0 +1,907 @@
+#ifndef GUARD_DATA_ITEM_ICON_TABLE_H
+#define GUARD_DATA_ITEM_ICON_TABLE_H
+
+#include "global.h"
+//external declarations from "data/graphics/items/item_graphics.inc"
+extern const u8 gItemIcon_QuestionMark[];
```
huderlem: Let's move all these extern gfx to `include/graphics.h`.

include/data/item_icon_table.h
```diff
@@ -0,0 +1,907 @@
...
+const u8 * const gItemIconTable[][2] = 
+{
+	gItemIcon_QuestionMark, gItemIconPalette_QuestionMark,
+// Pokeballs
+	gItemIcon_MasterBall, 	gItemIconPalette_MasterBall,
+	gItemIcon_UltraBall, 	gItemIconPalette_UltraBall,
+	gItemIcon_GreatBall, 	gItemIconPalette_GreatBall,
+	gItemIcon_PokeBall, 	gItemIconPalette_PokeBall,
+	gItemIcon_SafariBall, 	gItemIconPalette_SafariBall,
+	gItemIcon_NetBall, 	gItemIconPalette_NetBall,
```
huderlem: Files outside of `asm/` directory should use 4 spaces instead of tabs.

huderlem:
After addressing those two comments above, this should be ready for merge.

PR #266, opened 2018-06-16
Decompile cut field effect

DizzyEggg:
This was such a fun file to document, especially given how it changed from its Ruby counterpart. I added a couple of nice defines and corrected two functions and would appreciate if the pokeruby's file got updated with these.


src/fldeff_strength.c
```diff
@@ -14,11 +14,11 @@ static void sub_8145E74(void);
 // text
 bool8 SetUpFieldMove_Strength(void)
 {
-    if (npc_before_player_of_type(87) == TRUE)
+    if (SetLastTalkedObjectInFrontOfPlayer(87) == TRUE)
```
Diegoisawesome: Use define

src/rom6.c
```diff
@@ -121,14 +121,14 @@ bool8 SetUpFieldMove_RockSmash(void)
...
     {
         gSpecialVar_Result = GetCursorSelectionMonId();
-        gUnknown_03005DB0 = FieldCallback_Teleport;
-        gUnknown_0203CEEC = sub_8179834;
+        gFieldCallback2 = FieldCallback_PrepareFadeInFromMenu;
+        gPostMenuFieldCallback = sub_8179834;
         return TRUE;
     }
-    else if (npc_before_player_of_type(0x56) == TRUE)
+    else if (SetLastTalkedObjectInFrontOfPlayer(0x56) == TRUE)
```
Diegoisawesome: Use define

src/fldeff_cut.c
```diff
@@ -1,19 +1,684 @@
...
+                while (1)
+                {
+                    if (sHyperCutStruct[i].unk2[j] == 0)
+                        break;
+                    if (cutTiles[(u8)(sHyperCutStruct[i].unk2[j] - 1)] == FALSE)
+                    {
+                        tileCuttable = FALSE;
+                        break;
+                    }
+                    if (++j >= 2)
```
Diegoisawesome: Do while?

src/fldeff_cut.c
```diff
@@ -1,19 +1,684 @@
...
+    FieldEffectStart(FLDEFF_CUT_GRASS);
+}
+
+bool8 FldEff_CutGrass(void)
+{
+    s16 x, y;
+    u8 tileBehavior;
+    u8 i;
+
+    for (i = 0, PlaySE(SE_W015), PlayerGetDestCoords(&gPlayerFacingPosition.x, &gPlayerFacingPosition.y); i < CUT_HYPER_AREA; i++)
```
Diegoisawesome: I think you should move the two functions out of the for statement (and if necessary, the i = 0 if it comes before those in the code)

DizzyEggg: Copied directly from pokeruby.

src/fldeff_cut.c
DizzyEggg: Sure.

src/rom6.c
DizzyEggg: done

src/fldeff_strength.c
DizzyEggg: done

src/fldeff_cut.c
Diegoisawesome: Still, if it can be done better than pokeruby...

src/rom6.c
Diegoisawesome: Rename `SetLastTalkedObjectInFrontOfPlayer` to `CheckObjectGraphicsInFrontOfPlayer`

Diegoisawesome: Egg pls

PR #87, opened 2017-10-14
[WIP] Battle tent

PikalaxALT:

PR #268, opened 2018-06-17
Decompile a bit of battle tent

DizzyEggg:
Credits to Pikalax. I copied what he had done in his #87.

I also created a struct for battle frontier and got rid of some field names, because I feel they were incorrect. Documenting the frontier related things is going to be fun. :smiley:


src/battle_tent.c
```diff
@@ -0,0 +1,267 @@
...
+}
+
+void sub_81B9E88(void)
+{
+    sub_819DC00();
+}
+
+bool8 sub_81B9E94(void)
+{
+    return (gMapHeader.regionMapSectionId == 8
```
Diegoisawesome: Use constants

huderlem: `MAPSEC_SLATEPORT_CITY` is 8

DizzyEggg: Fixed.

PR #271, opened 2018-06-29
libisagbprn added volatile qualifier to touched pointers

felixjones:
This fixes libisagbprn as a standalone library for compilers that eagerly optimise away pointer access.


Diegoisawesome:
Pending testing in pokefirered.

Diegoisawesome:
Tested in FireRed, it doesn't seem to match unfortunately.

PR #283, opened 2018-08-02
Battle code enums in switches

DizzyEggg:
Easier to modify that way(adding or changing order of cases)


src/battle_util.c
```diff
@@ -490,7 +490,22 @@ u8 GetImprisonedMovesCount(u8 battlerId, u16 move)
     return imprisionedMoves;
 }
 
-u8 UpdateTurnCounters(void)
+enum
+{
+	ENDTURN_ORDER,
```
huderlem: nit: looks like you're using tabs

DizzyEggg: Is it good now?

PR #287, opened 2018-08-11
Pikalax's Easy chat

DizzyEggg:
#125 can be closed down.


include/global.h
```diff
@@ -719,8 +719,8 @@ struct LilycoveLadyQuiz
     /*0x000*/ u8 id;
     /*0x001*/ u8 phase;
     /*0x002*/ u16 unk_002[9];
-    /*0x014*/ u16 unk_014;
-    /*0x016*/ u16 unk_016;
+    /*0x014*/ u16 unk_014[1];
```
Diegoisawesome: Is this not unk_014[2], perhaps? This is just bizarre

src/easy_chat.c
```diff
@@ -0,0 +1,810 @@
...
+{
+    u16 v0;
+    s16 *data;
+
+    data = gTasks[taskId].data;
+    switch (data[EZCHAT_TASK_STATE])
+    {
+    case 0:
+        SetVBlankCallback(sub_811A290);
+        BlendPalettes(-1, 16, 0);
```
Diegoisawesome: Use 0xFFFFFFFFF to show it's a bitmask

src/easy_chat.c
```diff
@@ -0,0 +1,810 @@
...
+    }
+    gUnknown_0203A118->unk_0d = (sub_811F3AC() - 1) / 2 + 1;
+    return TRUE;
+}
+
+static void EasyChat_FreeResources(void)
+{
+    if (gUnknown_0203A118 != NULL)
+    {
+        free(gUnknown_0203A118);
```
Diegoisawesome: FREE_AND_SET_NULL macro

include/global.h
DizzyEggg: Don't ask me, it's Pikalax's.

Diegoisawesome: Try making the change and see if it works

DizzyEggg: I just made it two seperate u16, like it was before.

PR #304, opened 2018-08-24
Move data from graphics.s to C

Slawter666:
The following contained multiple .incbin's in graphics.s, however I couldn't get them to work as arrays or successive INCBIN's so I've given them seperate labels:
const u8 gHealthboxElementsGfxTable[]
const u8 gUnknown_08C1249C[]
const u8 gMonIcon_Deoxys[]
const u16 gTitleScreenBgPalettes[]


Diegoisawesome:
`src/data/graphics` would probably be better than `src/graphics`

PR #313, opened 2018-09-05
Implemented gRegionMapSectionId_To_PopUpThemeIdMapping array to be de…

ultima-soul:
…pendent on Map Section ID enum rather than order dependent. 

This allows for adding new maps easily and to prevent the map name pop-up not be corrupt due to editing arrays incorrectly for the new maps.

This was a collaborative effort from mvit, !Firefly, shinny, and I.


include/constants/region_map_sections.h
```diff
@@ -220,4 +220,6 @@
 #define MAPSEC_IN_GAME_TRADE                0xFE
 #define MAPSEC_FATEFUL_ENCOUNTER            0xFF
 
+#define MAPSEC_SUBSTRACT_KANTO (MAPSEC_SPECIAL_AREA - MAPSEC_DYNAMIC)
```
Diegoisawesome: Subtract

PR #326, opened 2018-09-15
Decompile cable_car_util.c

DizzyEggg:
Mind you, this file is in no way related to cable car, so I renamed it. Tho, I am not really sure what it is about.
Used by hall_of_fame.c to control oams, I think.
Seems to be connected in some way to rom_8034C54.c.


ProjectRevoTPP:
Part of this file needs to be split into roulette_util. Check pokeruby, please.

DizzyEggg:
> 
> 
> Part of this file needs to be split into roulette_util. Check pokeruby, please.

What? It has no connection to roulette.

PR #329, opened 2018-09-16
Use macros for EventObjectGraphicsInfo

Slawter666:
There are a few special cases that means the formatting is not as clean and concise as I would like, mostly involving shadowSize.


Slawter666:
I've tried to simplify the macros as much as I can, though I'm not too keen on them taking varying numbers of arguments.

DizzyEggg:
@Diegoisawesome can this be merged?

PR #336, opened 2018-09-21
port keysplit_tables from pokeruby

Kurausukun:
The key splits themselves are the same between the two games, so there's no reason not to make it easy for those who want to mess with the tables in Emerald.


sound/keysplit_tables.inc
```diff
@@ -395,3 +391,7 @@
 	.byte 1  @ 105
 	.byte 1  @ 106
 	.byte 1  @ 107
+
+	voice_square_1 0, 2, 0, 0, 15, 0  @ 86B4698
```
huderlem: These belong at the end of `voicegroup_86B429C`, not here.

PR #348, opened 2018-10-07
Decompile/Port battle_anim_80A5C6C

DizzyEggg:


src/battle_anim_80D51AC.c
```diff
@@ -0,0 +1,1007 @@
+#include "global.h"
+#include "battle.h"
+#include "battle_anim.h"
+#include "sprite.h"
+#include "task.h"
+#include "trig.h"
+
+#define TASK gTasks[taskId]
+#define SPRITE gSprites[TASK.data[0]]
```
huderlem: Not a fan of the `SPRITE` define.  It's jarring to see in the source.  I believe `TASK_SPRITE` would be an improvement because then you at least know the sprite id lives in the task.  At that point, it is really better than simply using `gSprites[TASK.data[0]]`?

PR #351, opened 2018-10-12
[DON'T MERGE YET] Decompile some of battle_frontier_2

DizzyEggg:
Which probably is apprentice


src/apprentice.c
```diff
@@ -0,0 +1,2380 @@
...
+#include "strings.h"
+#include "task.h"
+#include "text.h"
+#include "constants/apprentice.h"
+#include "constants/items.h"
+#include "constants/songs.h"
+#include "constants/species.h"
+#include "constants/moves.h"
+
+#define PLAYER_APPRENTICE gSaveBlock2Ptr->playerApprentice
```
Diegoisawesome: Was this define really necessary?

DizzyEggg: It gets a bit hard to read with field_B8[field_B1]

PR #356, opened 2018-10-14
Clean a couple of battlestructs' fields

DizzyEggg:


include/battle.h
```diff
@@ -128,42 +128,40 @@ struct UnknownFlags
...
-    /*0x15*/ u8 battlerWithSureHit;
-    /*0x16*/ u8 isFirstTurn;
-    /*0x17*/ u8 unk17;
-    /*0x18*/ u8 truantCounter:1;
-    /*0x18*/ u8 truantUnknownBit:1;
-    /*0x18*/ u8 unk18_a_2:2;
-    /*0x18*/ u8 unk18_b:4;
-    /*0x19*/ u8 rechargeCounter;
-    /*0x1A*/ u8 unk1A[2];
+     u32 transformedMonPersonality;
```
ProjectRevoTPP: where did the comment fields go? those are really helpful.

DizzyEggg: They WERE helpful, now that all code that uses these structs has been decompiled, there is no need to have them.

PR #361, opened 2018-10-16
Synchronise with pokeruby

Slawter666:


include/berry.h
```diff
@@ -1,6 +1,13 @@
 #ifndef GUARD_BERRY_H
 #define GUARD_BERRY_H
 
+#define BERRY_NONE 0
+#define FIRST_BERRY ITEM_CHERI_BERRY
+#define LAST_BERRY  ITEM_ENIGMA_BERRY
+
+#define GETBERRYID(berry) ((berry - FIRST_BERRY) + 1)
+#define GETITEMID(berry) ((berry + FIRST_BERRY) - 1)
```
Diegoisawesome: I know you want to sync with Ruby but these names are just bad. They should be `ITEM_ID_TO_BERRY_ID` and `BERRY_ID_TO_ITEM_ID`, or something along those lines.

DizzyEggg: ITEM_TO_BERRY already exists here https://github.com/pret/pokeemerald/blob/master/include/constants/items.h#L470

PR #372, opened 2018-10-21
[WIP] synchronize sprite tags

garakmon:

PR #373, opened 2018-10-23
Split battle_anim_80FE840 and move data to C

garakmon:


garakmon:
any thoughts on this? (might be easier to look at after #372 is merged)

garakmon:
OK this is done

ld_script.txt
```diff
@@ -459,7 +476,26 @@ SECTIONS {
         src/field_special_scene.o(.rodata);
         src/rotating_gate.o(.rodata);
         src/item_use.o(.rodata);
-        data/battle_anim_80FE840.o(.rodata);
+        /*src/battle/anim/effects_1.o(.rodata);*/
```
Diegoisawesome:
```suggestion

```
Remove this line

src/effects_1.c
```diff
@@ -0,0 +1,2093 @@
+#include "global.h"
```
Diegoisawesome: Will these files eventually contain code? If not they should go into src/data

garakmon: yes they will

ld_script.txt
garakmon: oops my bad

PR #381, opened 2018-11-11
Refactor for compatibility with modern gcc

PikalaxALT:
To build:

    make -f Modern.make


DizzyEggg:
I don't really like how every header file includes global.h. Your opinion @Diegoisawesome ?

PikalaxALT:
Oh, that change was to get my IDE to stfu about undefined types and macros (`u8` etc.)

aaaaaa123456789:
GCC errors out on `u8` if not defined. So, if GCC can figure out what `u8` is, it means that the relevant inclusion must be somewhere... and your IDE isn't following.

PikalaxALT:
The erroneous IDE complaints are in the header files.  CLion used to be able to resolve these correctly if the headers were included below `global.h` in the C sources.  That feature was deprecated for whatever asinine reason.

PikalaxALT:
It's OK on Mac and Cygwin, but not on CentOS.

PikalaxALT:
Revised building instructions for using modern gcc:

    make modern

PR #383, opened 2018-11-13
Update createsprite macro and rename BANK to BATTLER

garakmon:


DizzyEggg:
Eh, I dont really like making the defines longer than necessary.

Can't ANIM_TARGET and such stay the way they already are?

garakmon:
they could I guess I just changed them to match ruby

DizzyEggg:
Ruby can change them to match pokeemerald's 😃

garakmon:
why no build

PR #387, opened 2018-11-17
Battle pyramid

DizzyEggg:
Once I finish Battle Pyramid, my Battle Frontier adventure shall come to an end.


ld_script.txt
```diff
@@ -269,6 +269,7 @@ SECTIONS {
         src/battle_factory.o(.text);
         src/battle_pike.o(.text);
         asm/battle_frontier_2.o(.text);
```
Diegoisawesome: You forgot to remove battle_frontier_2.asm and its entry in the ld_script.

DizzyEggg: I didn't. There is code there, it has nothing to do with battle frontier though. Mossdeep gym tiles maybe?

Diegoisawesome: Maybe you should rename it to something that's not battle_frontier_2 if you're sure it's not frontier-related then.

PR #393, opened 2018-11-21
Match one function in text.c

DizzyEggg:


src/dewford_trend.c
```diff
@@ -1,13 +1,14 @@
...
 #include "easy_chat.h"
-#include "constants/easy_chat.h"
 #include "event_data.h"
 #include "link.h"
 #include "malloc.h"
 #include "random.h"
 #include "text.h"
 #include "tv.h"
+#include "string_util.h"
+#include "constants/easy_chat.h"
```
PikalaxALT: Why was this cut-and-pasted down here?

DizzyEggg: That's how it is in most files now.

PR #395, opened 2018-11-22
Document some battle ui

Slawter666:


Slawter666:
Those defines are already in constants/pokemon.h so I've removed the duplicates

DizzyEggg:
Oh, all good then :D

PR #422, opened 2018-12-03
Battle tower data

garakmon:


DizzyEggg:
@Diegoisawesome

include/constants/trainers.h
```diff
@@ -200,6 +200,8 @@
 #define FACILITY_CLASS_RS_BRENDAN           0x50
 #define FACILITY_CLASS_RS_MAY               0x51
 
+#define FACILITY_CLASSES_COUNT 0x52
```
Diegoisawesome: Align properly

src/battle_tower.c
```diff
@@ -102,6 +76,982 @@ static void SetEReaderTrainerChecksum(struct BattleTowerEReaderTrainer *ereaderT
...
+    BattleFrontier_BattleTowerMultiBattleRoom_Text_248D04
+};
+
+struct
+{
+    u32 facilityClass;
+    const u8 *const *strings;
+} const gUnknown_085DD500[] =
+{
+    {0x00000047, gUnknown_085DD118},
```
Diegoisawesome: Use constants

src/battle_tower.c
```diff
@@ -147,7 +147,7 @@ const u16 gBattleFrontierHeldItems[] =
 #include "data/battle_frontier/battle_frontier_trainers.h"
 #include "data/battle_frontier/battle_frontier_mons.h"
 
-const u8 gTowerMaleFacilityClasses[] = 
+const u8 gTowerMaleFacilityClasses[30] =
```
Meowsy: Is there any point to specifying the length of the arrays `gTowerMaleFacilityClasses`, `gTowerFemaleFacilityClasses`, `gTowerMaleTrainerGfxIds`, and `gTowerFemaleTrainerGfxIds`? The compiler should be able to figure that out automatically.

huderlem: yeah, they should be removed.  They are only useful when declaring them as extern.

DizzyEggg: They have to be there, otherwise you won't be able to extern it while specyfing the array length.

garakmon:
fixed all the things

src/battle_tower.c
Diegoisawesome: What do you mean? If you need to extern it from another file then you can do so without specifying the length.

Diegoisawesome: Apparently removing these causes Travis to have an internal compiler error?

garakmon: I guess it has something to do with apprentice.c using ARRAY_COUNT on this array but it not being defined in that file so the length needs to be explicit.

PR #430, opened 2018-12-06
Document trainer battle scripts

DizzyEggg:
And some other clean-up


src/trainer_see.c
```diff
@@ -630,26 +630,26 @@ static void Task_DestroyTrainerApproachTask(u8 taskId)
     EnableBothScriptContexts();
 }
 
-void sub_80B45D0(void)
+void TryPrepareSecondApproachingTrainer(void)
 {
     if (gNoOfApproachingTrainers == 2)
```
huderlem: `gNoOf` -> `gNum`

PR #441, opened 2018-12-10
Document bag pocket magic numbers.

Meowsy:


src/item_menu.c
```diff
@@ -526,9 +526,9 @@ void GoToBagMenu(u8 bagMenuType, u8 pocketId, void ( *postExitMenuMainCallback2)
             gUnknown_0203CE58.location = bagMenuType;
         if (postExitMenuMainCallback2)
             gUnknown_0203CE58.bagCallback = postExitMenuMainCallback2;
-        if (pocketId <= 4)
+        if (pocketId <= POCKETS_COUNT - 1)
```
huderlem: `if (pocketId < POCKETS_COUNT)`

src/item_menu.c
```diff
@@ -455,60 +455,60 @@ extern const u16 gUnknown_0860F074[];
...
 void CB2_ChooseBerry(void)
 {
-    GoToBagMenu(4, 3, CB2_ReturnToFieldContinueScript);
+    GoToBagMenu(4, BERRIES_POCKET, CB2_ReturnToFieldContinueScript);
 }
 
 void sub_81AABF0(void(*callback)(void))
 {
-    GoToBagMenu(5, 3, callback);
+    GoToBagMenu(5, BERRIES_POCKET, callback);
```
huderlem: The first parameter for `GotToBagMenu()` is essentially an index into `gReturnToXStringsTable`, if you're feeling up to documenting that, too.

Meowsy: Nice find!

src/item_menu.c
Meowsy: Hahaha, I was _hoping_ that would be equivalent, but I forgot to actually check.

src/item_menu.c
Meowsy: Documented.

PR #442, opened 2018-12-10
Add defines for Unown forms.

Meowsy:
Add defines for Unown forms.


include/pokemon.h
```diff
@@ -132,6 +132,35 @@
 #define STATUS_PRIMARY_FAINTED   0x7
 
 #define MAX_TOTAL_EVS 510
+
+#define UNOWN_LETTER_A 1
```
huderlem: If these constants aren't used anywhere, there isn't much purpose to have them.

Meowsy: That's a fair point. I was using them in my hack while testing how to have Unown spawn with specific forms, but this doesn't add anything to the disassembly as-is. I'll close this for now, and consider ways I could document the Unown forms that'd actually improve the disassembly.

PR #452, opened 2018-12-17
Add macros to simplify defining front animations and still pics.

Meowsy:
dd macros to simplify defining front animations and still pics. Also change some magic numbers to their correct defines.


src/pokemon.c
```diff
@@ -5935,13 +5935,13 @@ u16 HoennToNationalOrder(u16 hoennNum)
 
 u16 SpeciesToCryId(u16 species)
 {
-    if (species <= 250)
+    if (species <= SPECIES_CELEBI - 1)
         return species;
 
-    if (species < 276)
+    if (species < SPECIES_TREECKO - 1)
         return 200;
```
Diegoisawesome: Should be SPECIES_UNOWN - 1

Meowsy: That causes it to become nonmatching.

huderlem: The `return 200` should be `return SPECIES_UNOWN - 1`.

Meowsy: *Ah.* I wish Github would highlight the referenced line in yellow for these things. :V

PR #454, opened 2018-12-18
Decompile party menu

golemgalvanize:
It isn't quite documented nor is the massive amount of data ported over yet, but decompilation was done to the best I could.

Also due to being a large file and thus github comments acting oddly, I'd prefer if any edit comments leave the line in question to make fixing this PR more convenient for me and maintainers alike


src/party_menu.c
```diff
@@ -0,0 +1,7231 @@
...
+            InitWindows(gUnknown_08615810);
+            break;
+        case 1:
+            InitWindows(gUnknown_08615850);
+            break;
+        case 2:
+            InitWindows(gUnknown_08615890);
+            break;
+        default:
+            In
```
huderlem: Line 5799: you forgot to delete this whole commented out function

huderlem:
General feedback--try to not explicitly check for `== FALSE` or `!= FALSE`.  I noticed it in a ton of places, and it's a bit jarring sometimes.  The rest of the PR looks fine for the time being.
Bad:
```
if (ItemIsMail(gUnknown_0203CEFC) != FALSE)
if (IsDoubleBattle() == FALSE)
```

Good:
```
if (ItemIsMail(gUnknown_0203CEFC))
if (!IsDoubleBattle())
```

src/party_menu.c
Diegoisawesome: 5441: WaitFanfare's arg is a bool

Diegoisawesome: 5492 & 5514: Use -1 and -2 (should use constants but they don't exist yet)

Diegoisawesome: 6277: Please add some newlines here, way too long

golemgalvanize: when I tried it didn't match, maybe some other way of using them may work but simply substituting in -1 and -2 will not

PR #463, opened 2018-12-20
Use a macro for OLD_UNOWN base stats

ShinyDragonHunter:
I thought that with the SPECIES_OLD_UNOWN data having identical stats, it'd be better to use a macro for them.


src/data/pokemon/base_stats.h
```diff
@@ -1,6 +1,37 @@
-// Maximum value for a female pokemon is 254 (MON_FEMALE) which is 100% female.
-// 255 (MON_GENDERLESS) is reserved for genderless pokemon.
+// Maximum value for a female Pok�mon is 254 (MON_FEMALE) which is 100% female.
+// 255 (MON_GENDERLESS) is reserved for genderless Pok�mon.
 #define PERCENT_FEMALE(percent) min(254, ((percent * 255) / 100))
+#define OLD_UNOWN {							\
```
melthelesbian: i'd change this to `OLD_UNOWN_STAT_DATA` or something similar

ShinyDragonHunter: I'm actually no good at coming up with names. I did name it `OLD_UNOWN_DATA` previously.

src/data/pokemon/base_stats.h
```diff
@@ -8289,830 +8289,87 @@ const struct BaseStats gBaseStats[] =
...
-        .growthRate = GROWTH_MEDIUM_FAST,
-        .eggGroup1 = EGG_GROUP_UNDISCOVERED,
-        .eggGroup2 = EGG_GROUP_UNDISCOVERED,
-        .ability1 = ABILITY_NONE,
-        .ability2 = ABILITY_NONE,
-        .safariZoneFleeRate = 0,
-        .bodyColor = BODY_COLOR_BLACK,
-        .noFlip = FALSE,
-    },
+#define OLD_UNOWN_BASE_STATS {				\
```
Diegoisawesome: Move this macro to the top of the file

PR #466, opened 2018-12-22
improve compare macros

melthelesbian:


Diegoisawesome:
For the most part I like the changes in this PR, but something about rolling the checkflag into the goto macros feels weird. This will need to be discussed first.

melthelesbian:
@Diegoisawesome I ported the macros from [pokeruby](http://github.com/pret/pokeruby), if you want them unrolled, we should revert them there too, but I believe the reason they are rolled together is because `0` can mean either `unset` or `less than` depending on whether there's a trainer flag, so we can't just use constants.

PR #484, opened 2018-12-27
Misc overworld documentation

huderlem:


DizzyEggg:
Was commiting a porymap file intentional?

PR #501, opened 2018-12-31
Use macros for EventObjectGraphicsInfo

Slawter666:


Slawter666:
The pokeruby equivalent PR is here: https://github.com/pret/pokeruby/pull/722

huderlem:
Shelving--I've never been a fan of this approach, due to the inconsistent nature of these macros, and the amount that they obscure the underlying labels/data.

PR #507, opened 2019-01-02
Finish off labeling flags

Tetrable:


huderlem:
The typo "RECIEVED" was made 49 times

PR #509, opened 2019-01-03
Decompile match_call.c

huderlem:
0 NONMATCHING


include/strings.h
```diff
@@ -1993,4 +2000,211 @@ extern const u8 gText_CommunicationStandby5[];
 // roulette
 extern const u8 gUnknown_082A5B89[];
 
+// match call
+extern const u8 BattleFrontier_PokeNav_2A971C[];
```
huderlem: forgot to fix these names

huderlem: fixed

PR #514, opened 2019-01-06
Compile with optional debug information

SBird1337:
Compile with debug symbols, and expose then to the linker, to compile with debug information, use

`make DEBUG=1 -jN`

Comments appreciated, in the linker script I included some common dwarf sections, probably a bit more than agbcc will create, but they are linked if and only if compiled with `-g` so it does not affect the normal build process at all.


melthelesbian:
May want to change the build flag name, I think a similar PR would be useful for ruby, but `DEBUG` is used there to build roms with the in-game debug tools.

PR #516, opened 2019-01-06
Use SCRCMD_DEF define from firered

PikalaxALT:


DizzyEggg:
Thats clever, I like it

huderlem:
I'm not convinced this is an improvement.  It obscures the function signature, and the readability isn't any better, in my opinion.

SBird1337:
I'd agree with @huderlem - What's the point in doing that, apart from having a few less characters?

Diegoisawesome:
Agreed, I don't like this change for the same reasons stated above.

PikalaxALT:
Okay, I'll close this and expand the macro in firered.

PR #518, opened 2019-01-09
[WIP] start decompiling trainer_card.s

shinny456:


src/trainer_card.c
```diff
@@ -0,0 +1,311 @@
...
+    case 5:
+        sub_80C45C0(gUnknown_02039CE8->var_598);
+        gUnknown_02039CE8->var_0++;
+        break;
+    case 6:
+        sub_80C4630();
+        gUnknown_02039CE8->var_0++;
+        break;
+    case 7:
+        if(gWirelessCommType == TRUE && gReceivedRemoteLinkPlayers == TRUE)
```
huderlem: Do not use `TRUE` when checking `gWirelessCommType`.  It takes other values besides 0 and 1.

src/trainer_card.c
```diff
@@ -0,0 +1,311 @@
...
+        {
+            sub_800E0E8();
+            CreateWirelessStatusIndicatorSprite(230, 150);
+        }
+        BlendPalettes(-1, 16, gUnknown_02039CE8->var_52C);
```
huderlem: palette blending functions take a bitmask as their first argument to select which palettes it's applying to.  Therefore, use `0xFFFFFFFF` instead of  -1

PR #521, opened 2019-01-19
Dump trainer hill trainers

DizzyEggg:


Diegoisawesome:
Could you redump with the easy chat word constants?

PR #522, opened 2019-01-19
Trainer card

DizzyEggg:
The continuation of shinny's work.


src/trainer_card.c
```diff
@@ -0,0 +1,1857 @@
...
+    if (trainerCard->gender == FEMALE)
+        trainerCard->var_4F = gUnknown_08329D54[(trainerCard->trainerId % 8) + 8];
+    else
+        trainerCard->var_4F = gUnknown_08329D54[trainerCard->trainerId % 8];
+}
+
+void TrainerCard_GenerateCardForPlayer(struct TrainerCard *trainerCard)
+{
+    memset(trainerCard, 0, 0x60);
+    trainerCard->version = VERSION_EMERALD;
```
Diegoisawesome: Use GAME_VERSION

src/trainer_card.c
```diff
@@ -0,0 +1,1857 @@
...
+    SetPlayerCardData(trainerCard, CARD_TYPE_EMERALD);
+    trainerCard->var_3A = HasAllFrontierSymbols();
+    *((u16*)&trainerCard->berruCrushPoints) = gSaveBlock2Ptr->frontier.field_EBA;
```
Diegoisawesome: field_EBA might be a signed field which is causing the cast

src/trainer_card.c
```diff
@@ -0,0 +1,1857 @@
...
+    {
+        xOffset = GetStringRightAlignXOffset(1, gStringVar4, 144);
+        top = 56;
+    }
+    else
+    {
+        xOffset = GetStringRightAlignXOffset(1, gStringVar4, 128);
+        top = 57;
+    }
+    AddTextPrinterParameterized3(1, 1, xOffset, top, gUnknown_0856FB0C, -1, gStringVar4);
```
Diegoisawesome: TEXT_SPEED_FF when using any AddTextPrinterParameterized function

src/trainer_card.c
```diff
@@ -0,0 +1,1857 @@
...
+    CpuSet(gMonIconPalettes, sData->var_468, 0x60);
+    switch (sData->trainerCard.var_4E)
+    {
+    case 0:
+        break;
+    case 1:
+        TintPalette_CustomTone(sData->var_468, 96, 0, 0, 0);
+        break;
+    case 2:
+        TintPalette_CustomTone(sData->var_468, 96, 500, 330, 310);
```
Diegoisawesome: These should be fixed point values

src/trainer_card.c
```diff
@@ -0,0 +1,1857 @@
...
+        sData->var_52C = 0x7FFF;
+    else
+        sData->var_52C = 0;
+
+    if (InUnionRoom() == TRUE)
+        sData->isLink = TRUE;
+    else
+        sData->isLink = FALSE;
+
+    sData->language = LANGUAGE_ENGLISH;
```
Diegoisawesome: Use GAME_LANGUAGE

src/trainer_card.c
```diff
@@ -0,0 +1,1857 @@
...
+#include "script_pokemon_util_80F87D8.h"
+#include "constants/songs.h"
+#include "constants/flags.h"
+#include "constants/game_stat.h"
+#include "constants/battle_frontier.h"
+
+enum
+{
+    CARD_TYPE_FR,
+    CARD_TYPE_RUBY,
```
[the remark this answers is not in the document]
DizzyEggg: eh, I think it's better my way personally

src/trainer_card.c
DizzyEggg: That's not it.

src/trainer_card.c
DizzyEggg: All of them?

Diegoisawesome: Just the last three (the R, G, and B coefficients)

src/trainer_card.c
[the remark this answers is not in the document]
Diegoisawesome: They're the same between their companion games though and generally values reflect this

PR #524, opened 2019-01-24
Obsolete common_syms and rewrite sym files, to keep up with pokeruby

nullableVoidPtr:
As per @luckytyphlosion's suggestion.


DizzyEggg:
I dont like this at all and I don't want this to be merged.

I hope @Diegoisawesome, you share my point. The obvious reason is having to define the size of each variable instead of the compiler/linker automatically calculating it.

Also the same goes for having to write .align 2 everytime, instead of it being done by the compiler/linker.

PR #535, opened 2019-02-01
Start Use Pokeblock

DizzyEggg:
So I started this file like a very long time ago but dropped it for some reason. I don't really feel like working on it anymore, but it's a good start.


DizzyEggg:
Wait what, this file was already started lmao. FAIL

PR #536, opened 2019-02-02
Use Menu and List Menu input defines

DizzyEggg:


src/player_pc.c
```diff
@@ -1218,21 +1218,23 @@ static void sub_816C4FC(u8 taskId)
...
         break;
-    case -2:
+    case LIST_B_PRESSED:
         if (gMain.newKeys & A_BUTTON)
         {
             ItemStorage_DoItemSwap(taskId, FALSE);
         }
         else
-            ItemStorage_DoItemSwap(taskId, TRUE);
+		{
```
Diegoisawesome: Tabs? Seriously?

DizzyEggg: angery

PR #547, opened 2019-02-08
Contribute upstream - Intro movie documentation & Misc

Phlosioneer:
Hello,

I'm doing a bit of a side-project to change emerald. I wasn't planning to contribute anything major upstream, but I saw the recent activity and decided to share what I've documented. I had to undo some changes that I had already made, but it should be clean now.

It may not be strictly assembly-accurate after a few of these changes; I haven't been trying to preserve the structure. If there's any large discrepancies, let me know and I'll try to fix them for this merge.


ProjectRevoTPP:
```
pokeemerald.gba: FAILED
/usr/bin/sha1sum: WARNING: 1 computed checksum did NOT match
```

We require the ROM to match. Any changes made to codegen that cause the ROM to change are not allowed and cause Travis to fail the checksum check. Please go over your changes to ensure none of them caused things like the generated assembly to change.

src/palette.c
```diff
@@ -578,10 +582,15 @@ static u8 UpdateFastPaletteFade(void)
     s8 b;
 
     if (!gPaletteFade.active)
-        return 0;
+        return PALETTE_FADE_STATUS_DONE;
 
-    if (IsSoftwarePaletteFadeFinishing())
-        return gPaletteFade.active;
+    if (IsSoftwarePaletteFadeFinishing()) {
```
ProjectRevoTPP: This change might cause the assembly to change. Check and make sure it does not.

src/task.c
```diff
@@ -44,6 +44,8 @@ u8 CreateTask(TaskFunc func, u8 priority)
         }
     }
 
+    // Error: Too many tasks!
+    // FIXME: This silently fails.
```
ProjectRevoTPP: This is supposed to silently fail. In the gen 3 debug ROM (specifically german ruby) it actually calls the crash function here instead of returning a 0.

src/overworld.c
```diff
@@ -2284,11 +2283,6 @@ void sub_8086C2C(void)
     c1_link_related_func_set(sub_80870B0);
 }
 
-static void sub_8086C40(void)
```
ProjectRevoTPP: You removed/moved a function. It needs to be right here for the ROM to match.

Phlosioneer:
Hm. On second thought, then, it might be too much work to go through and figure out what changed things, and what did not.

Sorry to waste your time.

ProjectRevoTPP:
Just join our discord (in the README) and we can help you.

PR #550, opened 2019-02-10
Name the possible return values of palette fading functions

Phlosioneer:
Also replace implicit conversion from gPalette.active into a status with
if statements, to make the return value clear.

I've also added comments when the check is redundant.


include/palette.h
```diff
@@ -9,6 +9,11 @@
 #define PLTT_BUFFER_SIZE 0x200
 #define PLTT_DECOMP_BUFFER_SIZE (PLTT_BUFFER_SIZE * 2)
 
+#define PALETTE_FADE_STATUS_DELAY 2
+#define PALETTE_FADE_STATUS_ACTIVE 1
+#define PALETTE_FADE_STATUS_DONE 0
+#define PALETTE_FADE_STATUS_LOADING -1
```
garakmon: could you change this to `0xFF` pretty please

src/palette.c
```diff
@@ -483,7 +483,9 @@ static u8 UpdateNormalPaletteFade(void)
             }
         }
 
-        return gPaletteFade.active;
+        // gPaletteFade.active cannot change since the last time it was checked. So this
+        // is equivalent to `return PALETTE_FADE_STATUS_ACTIVE;`
+        return gPaletteFade.active ? PALETTE_FADE_STATUS_ACTIVE : PALETTE_FADE_STATUS_DONE;
```
Diegoisawesome: Why the need to complicate this with a ternary? Leaving it as it is is fine because it can never be anything other than 0 or 1 anyway.

PR #553, opened 2019-02-11
Document main_menu state machine, name window constants

Phlosioneer:
The new names for the constants (and newly named functions) give a much
better understanding of what the window system is trying to do.


Phlosioneer:
This now works.

asm/rom_8011DC0.s
```diff
@@ -745,7 +745,7 @@ _08017228:
 	bl DestroyListMenuTask
 	ldrb r0, [r5]
 	movs r1, 0x1
-	bl sub_819746C
+	bl ClearWindowAndBorder
```
Diegoisawesome: There are many functions that do this same functionality, you should probably distinguish them more clearly

include/window.h
```diff
@@ -1,6 +1,13 @@
 #ifndef GUARD_WINDOW_H
 #define GUARD_WINDOW_H
 
+// The tile ID for a transparent tile in the window tilesheet
+#define TRANSPARENT_TILE_NUMBER 0
+
+#define PIXEL_BUFFER_TRANSPARENT 0x00
+#define PIXEL_BUFFER_WHITE 0x11
```
Diegoisawesome: You shouldn't make the assumption this will always be the white color. Any color could be loaded here. I don't even think these need constants

src/pokemon_storage_system.c
```diff
@@ -1218,7 +1218,7 @@ static const u16 gWallpaperPalettes_Plain[][16] =
 static const u32 gWallpaperTiles_Plain[] = INCBIN_U32("graphics/pokemon_storage/plain.4bpp.lz");
 static const u32 gWallpaperTilemap_Plain[] = INCBIN_U32("graphics/pokemon_storage/plain.bin.lz");
 
-// 12�18 tilemap
+// 12�18 tilemap
```
Diegoisawesome: 12x18

include/window.h
```diff
@@ -73,6 +80,7 @@ extern u32 filler_03002F58;
 extern u32 filler_03002F5C;
 extern u32 filler_03002F64;
 
-extern u8 gUnknown_03002F60;
+// This global is set to TRANSPARENT_TILE_NUMBER and never changed.
+extern u8 gTransparentTileNumber;
```
Diegoisawesome: This is only ever used in window.c, it doesn't need an entry here.

src/window.c
```diff
@@ -6,7 +6,7 @@
 
 u32 filler_03002F58;
 u32 filler_03002F5C;
-u8 gUnknown_03002F60;
+u8 gTransparentTileNumber;
```
Diegoisawesome: Should be static

include/window.h
Phlosioneer: I named it because it's specifically the menu's palette. Clear-this-area-to-white happens extremely often in the message box code.

Phlosioneer: I ended up making a new macro for this that turned out to be much more useful.

src/window.c
Phlosioneer: It cannot be defined as static, because it is used in the `window.txt` linker file. And it can't be removed from that file without causing a compiler error:
```
`gTransparentTileNumber' referenced in section `.text' of src/window.o: defined in discarded section `COMMON' of src/window.o
```

Phlosioneer:
Updated this PR.

src/menu.c
```diff
@@ -458,12 +458,12 @@ void DisplayItemMessageOnField(u8 taskId, const u8 *string, TaskFunc callback)
     CopyWindowToVram(0, 3);
 }
 
-void DisplayYesNoMenu(void)
+void DisplayYesNoMenuWithDefault(void)
```
huderlem: I think these two names are swapped.  This one should be "default Yes".  The next one should be "with default" since the initial cursor position is an argument.

huderlem:
The rest looks fine to me.  Merge conflicts must be resolved and the name issue I mentioned above.

include/window.h
```diff
@@ -1,6 +1,8 @@
 #ifndef GUARD_WINDOW_H
 #define GUARD_WINDOW_H
 
+#define PALETTE_NUM_TO_FILL_VALUE(num) ((num) | ((num) << 4))
```
huderlem: I just realized this is not named properly.  This argument `num` is a pixel value, not a palette, as the define name suggests.  Perhaps `PIXEL_FILL` would be better?

Phlosioneer: Is it? The input numbers seem to reference colors in the current palette. For example, 0 is always used as transparent, and in the menu code 1 always means the background white color.

huderlem: Correct. They are color indexes in the current palette.  A pixel is defined as one of these color indexes.

Phlosioneer: Okay. So rename the whole thing `PIXEL_FILL`?

huderlem: That's what I would do

Phlosioneer:
This PR has been updated.

PR #554, opened 2019-02-14
fix typos

begonethotypo:


Phlosioneer:
This fails because there are three places that still need to be renamed. Two here:
https://github.com/pret/pokeemerald/blob/master/include/battle_scripts.h#L91
One here:
https://github.com/pret/pokeemerald/blob/master/src/battle_main.c#L3282

PR #559, opened 2019-02-15
Finish and polish new_game.c

Phlosioneer:
Only one function is left undocumented. I tried to go into the assembly
to at least name it, but it clears fields in the save block that aren't
even documented yet.

### DIFF
--- a/include/pokedex.h
+++ b/include/pokedex.h
@@ -1,7 +1,7 @@
 #ifndef GUARD_POKEDEX_H
 #define GUARD_POKEDEX_H
 
-extern u8 gUnknown_030060B0;
+extern u8 gUnusedU8;
 extern void (*gUnknown_030060B4)(void);
 
 #define HOENN_DEX_COUNT 202
@@ -42,5 +42,6 @@ u8 CreateDexDisplayMonDataTask(u16 dexNum, u32 trainerId, u32 personality);
 s8 GetSetPokedexFlag(u16 nationalNum, u8 caseId);
 u16 CreateMonSpriteFromNationalDexNumber(u16, s16, s16, u16);
 bool16 HasAllHoennMons(void);
+void ResetPokedexScrollPositions(void);
 
 #endif // GUARD_POKEDEX_H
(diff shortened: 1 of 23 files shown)


include/pokedex.h
```diff
@@ -1,7 +1,7 @@
 #ifndef GUARD_POKEDEX_H
 #define GUARD_POKEDEX_H
 
-extern u8 gUnknown_030060B0;
+extern u8 gUnusedU8;
```
Diegoisawesome: This name should be more descriptive, as `gUnusedU8` could be literally anywhere

src/link_rfu.c
```diff
@@ -2367,52 +2368,64 @@ bool32 sub_800E388(const u8 *name)
...
 {
     if (gWirelessCommType != 0)
     {
         s32 i;
         s32 j;
-        s32 cnt;
-        s32 sp0[5];
-        struct UnkSaveSubstruct_3b98 *sp14 = calloc(20, sizeof(struct UnkSaveSubstruct_3b98));
+        s32 nextSpace;
+        s32 connectedTrainerRecordIndecies[5];
```
Diegoisawesome: Indecies -> indices

src/new_game.c
```diff
@@ -170,7 +164,7 @@ void NewGameInitData(void)
     ClearSav1();
     ClearMailData();
     gSaveBlock2Ptr->specialSaveWarpFlags = 0;
-    gSaveBlock2Ptr->field_A8 = 0;
+    gSaveBlock2Ptr->unusedFlagField = 0;
```
Diegoisawesome: Should be left as-is until we've exhausted all gSaveBlock2Ptr references from the assembly

src/link_rfu.c
```diff
@@ -2376,18 +2376,18 @@ void RecordMixTrainerNames(void)
         s32 i;
         s32 j;
         s32 nextSpace;
-        s32 connectedTrainerRecordIndecies[5];
+        s32 connectedTrainerRecordIndicies[5];
```
Diegoisawesome: Indices

PR #565, opened 2019-02-18
Fix scaninc ignoring .h includes in asm files

Phlosioneer:
It was parsing lines starting with # as a comment, when we are using
those comments to include C header files.

### DIFF
--- a/tools/scaninc/scaninc.cpp
+++ b/tools/scaninc/scaninc.cpp
@@ -145,9 +145,11 @@ int main(int argc, char **argv)
 
             while ((incDirectiveType = file.ReadUntilIncDirective(path)) != IncDirectiveType::None)
             {
-                if (path.back() == 'h') {
+                if (path.back() == 'h')
+                {
                     std::string newPath("include/" + path);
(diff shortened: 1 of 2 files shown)


tools/scaninc/scaninc.cpp
```diff
@@ -145,6 +145,15 @@ int main(int argc, char **argv)
 
             while ((incDirectiveType = file.ReadUntilIncDirective(path)) != IncDirectiveType::None)
             {
+                if (path.back() == 'h')
+                {
+                    std::string newPath("include/" + path);
+                    if (!CanOpenFile(newPath))
+                    {
+                        newPath = "include/constants/" + path;
```
huderlem: This path assumption should not be hardcoded in the scaninc tool.  Use `includeDirs` to search for the file.

Phlosioneer: I tried that, but I couldn't get it to work correctly.

huderlem:
Closing this for now, since the real fix is much different than the proposed changes.

PR #566, opened 2019-02-19
Event handbook

Phlosioneer:
This is a huge PR. Sorry. There were a lot of renames, and a lot of false leads / re-renames; and while each individual commit compiles fine, they often reference documentation that doesn't yet. I'm pretty happy with how it looks right now, it's all cohesive.

This PR is all about documenting the event system in exacting detail. Comments on functions are not enough; there's interlocking systems here, and the control flow is convoluted and spread across multiple files. So I started an "event handbook" - a set of markdown documentation of events, and how event scripts interact and form a cohesive animation sequence.

This PR also did significant reverse-engineering of the MapScript system and some parts of map loading.


Phlosioneer:
I can split this up into individual commits, since they each compile successfully. But the comments and names in earlier commits change in later ones, so it doesn't make sense to be a fully separate PR for each. How do you want to approach this?

data/maps/LittlerootTown/scripts.inc
```diff
@@ -48,18 +48,22 @@ LittlerootTown_EventScript_1E7E5C:: @ 81E7E5C
...
+@ This will permamently move her to may's house.
+LittlerootTown_EventScript_MoveMomToMaysHouse:: @ 81E7E67
 	setobjectxyperm 4, 14, 8
 	return
 
-LittlerootTown_EventScript_1E7E6F:: @ 81E7E6F
-	compare VAR_0x4050, 0
-	goto_if_eq LittlerootTown_EventScript_1E7E86
+LittlerootTown_EventScript_SetTwinPosition:: @ 81E7E6F
+	compare VAR_BIRCH_ATTACK, BIRCH_ATTACK_INACTIVE
```
huderlem: In pokeruby, your `VAR_BIRCH_ATTACK` var is called `VAR_LITTLEROOT_STATE`.  I'm worried that other vars in this PR deviate significantly from pokeruby.  In my opinion, `VAR_LITTLEROOT_STATE` is better for this var, since this is not specific to the Birch attack.

data/maps/LittlerootTown/scripts.inc
```diff
@@ -186,21 +190,21 @@ LittlerootTown_Movement_1E7FAC: @ 81E7FAC
...
+	applymovement 8, Common_Movement_EmoteExclamationMark
 	waitmovement 0
 	delay 80
 	msgbox LittlerootTown_Text_1E8DA2, MSGBOX_DEFAULT
 	closemessage
 	clearflag FLAG_HIDE_LITTLEROOT_TOWN_RIVAL
 	clearflag FLAG_HIDE_LITTLEROOT_TOWN_BIRCH
 	delay 20
-	clearflag FLAG_SPECIAL_FLAG_0x4000
+	clearflag FLAG_HIDE_MAP_NAME
```
huderlem: minor: I would prefer `FLAG_HIDE_MAP_NAME_POPUP`.

huderlem:
I'm not sure the "event handbooks" belong in this repo's source tree.

include/constants/event_object_var_values.h
```diff
@@ -0,0 +1,12 @@
+#ifndef GUARD_CONSTANTS_EVENT_OBJECT_VAR_VALUES_H
```
huderlem: I'd rename this to `event_var_values.h`.  And then add a comment above each set clearly stating which var the set is for.

include/constants/global.h
```diff
@@ -74,4 +74,14 @@
 #define OPTIONS_BATTLE_STYLE_SHIFT 0
 #define OPTIONS_BATTLE_STYLE_SET 1
 
+// Constants for map header tags.
+#define EVENT_TAG_ON_MAP_LOAD 1
```
huderlem: these aren't "event tags", they are "map script types".

include/constants/metatile_behaviors.h
```diff
@@ -242,4 +242,96 @@
 #define MB_UNUSED_EE 0xEE
 #define MB_UNUSED_EF 0xEF
 
+// Raw metatile ID's.
+#define METATILE_ID_UNK_206 0x206
```
huderlem: metatile ids do not belong in the metatile behaviors header file.  At minimum, this should be its own file, but I think individual C files should be using their own `#define` values for this, rather than creating a header file.

include/overworld.h
```diff
@@ -68,7 +68,7 @@ void SetContinueGameWarpToDynamicWarp(int unused);
 const struct MapConnection *GetMapConnection(u8 dir);
 bool8 SetDiveWarpEmerge(u16 x, u16 y);
 bool8 SetDiveWarpDive(u16 x, u16 y);
-void mliX_load_map(u8 mapGroup, u8 mapNum);
+void LoadMapFromCameraTranition(u8 mapGroup, u8 mapNum);
```
huderlem: typo

src/field_specials.c
```diff
@@ -654,87 +679,87 @@ void MauvilleGymSpecial2(void)
         {
             switch (MapGridGetMetatileIdAt(x, y))
             {
-                case 0x220:
-                    MapGridSetMetatileIdAt(x, y, 0x230);
+                case (OBJ_0 | ON):
```
huderlem: I find this switch is now harder to comprehend that with raw numbers.

src/fieldmap.c
```diff
@@ -62,19 +62,19 @@ void InitMapFromSavedGame(void)
...
 {
-    CpuFastFill(0x03ff03ff, gBackupMapData, sizeof(gBackupMapData));
+    CpuFastFill(MAP_UNDEFINED_METATILE_ID_32, gBackupMapData, sizeof(gBackupMapData));
     GenerateBattlePyramidFloorLayout(gBackupMapData, setPlayerPosition);
 }
 
 void InitTrainerHillMap(void)
 {
-    CpuFastFill(0x03ff03ff, gBackupMapData, sizeof(gBackupMapData));
+    CpuFastFill(MAP_UNDEFINED_METATILE_ID_32, gBackupMapData, sizeof(gBackupMapData));
```
huderlem: this is two metatiles, not one

src/script.c
```diff
@@ -241,7 +241,7 @@ void ScriptContext2_RunNewScript(const u8 *ptr)
     while (RunScriptCommand(&sScriptContext2) == TRUE);
 }
 
-u8 *mapheader_get_tagged_pointer(u8 tag)
+u8 *MapHeaderGetEventTable(u8 tag)
```
huderlem: The language for this system should be changed to "map scripts", instead of "map header event table"

PR #567, opened 2019-02-19
Documented/polished pokemon_summary_screen.c

arantonitis:
I just traced a few of the functions and experimented, and re-labeled some of the unknown stuff.


arantonitis:
I'm making a new PR with a proper branch that isn't master, also I relabeled a few more functions I have yet to add.

PR #573, opened 2019-02-22
Fix while (1) and while (0) to use TRUE/FALSE

Phlosioneer:


garakmon:
whatever you used to do the replacing here messed up the encoding

src/contest_effect.c
```diff
@@ -129,7 +129,7 @@ static void ContestEffect_UserLessEasilyStartled(void)
     SetContestantEffectStringID(shared192D0.contestant,CONTEST_STRING_STOPPED_CARING);
 }
 
-// Slightly startles the POK�MON in front.
+// Slightly startles the POKéMON in front.
```
Phlosioneer: I don't understand what the problem is here. Using a UTF-8 string analyzer, both `POKéMON` displayed here are using U+00E9 for the `e`.

DizzyEggg:
@Diegoisawesome I don't really like this change tbh, while 1 is faster to write and read and it's the same thing anyway

huderlem:
Yeah, I don't feel strongly enough to enforce this on the whole codebase, especially because it's not a typical C thing, to my knowledge.  Should probably just close this.

PR #591, opened 2019-03-02
Fix bad sigs

Phlosioneer:
Fix GetMonData and GetBoxMonData to typecheck correctly using the magic of _macros_.

### DIFF
--- a/asm/pokenav.s
+++ b/asm/pokenav.s
@@ -10873,7 +10873,7 @@ _081CD690:
 	adds r4, r0
 	adds r0, r4, 0
 	movs r1, 0x38
-	bl GetMonData
+	bl GetMonDataExtended
 	lsls r0, 16
 	lsrs r0, 16
 	mov r8, r0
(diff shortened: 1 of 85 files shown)


huderlem:
I don't think this is an improvement.

huderlem:
The type-checking is an improvement, but I don't want `GetMonData_2`s floating around everywhere.

Phlosioneer:
But it's already found bugs. Would a different name be better?

Phlosioneer:
*_2 is the standard way we deal with duplicate function definitions in this repo.

huderlem:
One idea is to change the 3-arg version of `GetMonData` to `BufferMonData`.  And then the 2-arg version can remain `GetMonData`.

Phlosioneer:
Sure, as long as it's typechecked I'm happy.

Phlosioneer:
How does `GetMonData` for 2 and `GetMonDataExtended` for 3 sound?

Phlosioneer:
Get*MonData => Get*MonDataExtended
Get*MonData_2 => Get*MonData

huderlem:
Closing per offline discussion.  Bug fixes exposed by type checking will be in a followup PR.

PR #592, opened 2019-03-02
Extern cleanup

Phlosioneer:


include/water.h
```diff
@@ -0,0 +1,6 @@
+#ifndef GUARD_WATER_H
+#define GUARD_WATER_H
+
+void sub_8108C94(struct Sprite *sprite);
```
huderlem: This should go in battle_anim.  `water.c` is just some battle anim code for water effects.

include/flying.h
```diff
@@ -0,0 +1,6 @@
+#ifndef GUARD_FLYING_H
+#define GUARD_FLYING_H
+
+void sub_810E2C8(struct Sprite *sprite);
```
huderlem: This should go in `battle_anim.h`

src/field_specials.c
```diff
@@ -591,7 +589,9 @@ void SpawnLinkPartnerEventObject(void)
 
 static void LoadLinkPartnerEventObjectSpritePalette(u8 graphicsId, u8 localEventId, u8 paletteNum)
 {
-    paletteNum += 6;
+    u8 adjustedPaletteNum;
+    // Note: This temp var is necessary; paletteNum += 6 doesn't match.
+    adjustedPaletteNum = paletteNum + 6;
```
huderlem: Why change the code of the this function?  It matched before, and now it's messier.

Phlosioneer: The file had an `extern` version of LoadPalette that used `u32` for the 2nd param instead of `u16`.
Basically, it only matched because LoadPalette was incorrectly defined. This fixes it to use the correct LoadPalette signature.

From the commit message:
```
In LoadLinkPartnerEventObjectSpritePalette, using paletteNum += 6 causes
GCC to "mark" paletteNum's register as a possible overflow, even though
it performs the overflow check. Then, when passing args to LoadPalette,
GCC fails to recognize that 0x100 + ((any u8) << 4) can never overflow
as a u16. It then does an overflow check before calling LoadPalette.

Assigning to a temp variable seems to avoid this "marking" behavior.
```

huderlem: gotcha

Phlosioneer:
This PR has been updated.

huderlem:
This has a merge conflict now

Phlosioneer:
This PR is updated.

PR #596, opened 2019-03-03
Make more effective use of pokemon move macros

Phlosioneer:
I'm open to possible name changes, particularly the prefix. `LEVEL_UP_*` isn't great, but I couldn't think of a better one, so I stuck with the original.

### DIFF
--- a/include/pokemon.h
+++ b/include/pokemon.h
@@ -134,6 +134,13 @@
 #define MAX_TOTAL_EVS 510
 #define UNOWN_FORM_COUNT 28
 
+#define LEVEL_UP_MOVE(level, move) ((level << 9) | move)
+#define LEVEL_UP_END 0xFFFF
+#define LEVEL_UP_MOVE_MASK 0x1FF
+#define LEVEL_UP_LEVEL_MASK 0xFE00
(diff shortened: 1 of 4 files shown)


DizzyEggg:
I'd remove LEVEL_UP_END tbh, 0xFF, 0xFFFF and -1 are everywhere in the repo and they usually mean end of table, so it's kinda self-exploratory.

Phlosioneer:
Normally you're right, but I disagree in this case. The LEVEL_UP_END constant is directly used in if statements and such in not-obvious ways. I think spelling it out makes the code much more clear.

Phlosioneer:
Whoops

Diegoisawesome:
Maybe it would be better to name it something like `INVALID_MOVE`.

Phlosioneer:
This PR has been updated for a while.

include/pokemon.h
```diff
@@ -135,7 +135,7 @@
 #define UNOWN_FORM_COUNT 28
 
 #define LEVEL_UP_MOVE(level, move) ((level << 9) | move)
-#define LEVEL_UP_END 0xFFFF
```
Diegoisawesome: This was left in here...

PR #600, opened 2019-03-03
Summary screen macros

Phlosioneer:
Minor improvements. I might have missed some places that need macros.


huderlem:
2 thoughts:

I think these are not needed:
```
#define PSS_CHANGE_NEXT 1
#define PSS_CHANGE_PREV -1
```

I'd prefer `sMonSummaryScreen` instead of `sMonSummaryScreenData`.

Also, the build is broken.

Phlosioneer:
This PR is updated.

huderlem:
The summary screen struct's name was not changed, as stated in the commit message.  Also there is some spacing issues with find-replace:
`delta == -1&&`

Phlosioneer:
This PR is updated. I actually did the rename this time! I hope.

PR #601, opened 2019-03-04
Overhaul scaninc to work recursively

Phlosioneer:
This also fixes the bug where scaninc would ignore #include lines
in assembly files.

This is a followup on #565

### DIFF
--- a/src/item_menu.c
+++ b/src/item_menu.c
@@ -1,3 +1,4 @@
+#include <string.h>
 #include "global.h"
 #include "item_menu.h"
 #include "battle.h"
@@ -40,7 +41,6 @@
 #include "shop.h"
 #include "sound.h"
 #include "sprite.h"
-#include "string.h"
 #include "strings.h"
 #include "string_util.h"
 #include "task.h"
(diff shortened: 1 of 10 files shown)


Phlosioneer:
This PR is updated.

tools/scaninc/scaninc.h
```diff
@@ -1,4 +1,4 @@
-// Copyright(c) 2015-2017 YamaArashi
+// Copyright(c) 2019 Phlosioneer
```
Diegoisawesome: Don't remove Yama's copyright

Phlosioneer: Oops, that was just supposed to be the new files.

src/item_menu.c
```diff
@@ -1,3 +1,4 @@
+#include <string.h>
```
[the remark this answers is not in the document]
Phlosioneer: Because `#include "string.h"` was making scaninc look in our include dirs for a file named `string.h`. It's more correct to specify that it's a library header, not a local one.

PR #607, opened 2019-03-10
Contest data

Phlosioneer:
Move all contest data into C structures. Contest strings are still in asm, somewhere in event_scripts.s.


data/maps/AbandonedShip_CaptainsOffice/map.json
```diff
@@ -25,7 +25,7 @@
       "trainer_type": "0",
       "trainer_sight_or_berry_tree_id": "0",
       "script": "AbandonedShip_CaptainsOffice_EventScript_2387E2",
-      "flag": "0"
+      "flag": "0"
```
huderlem: please don't add extraneous changes to the json files.

include/contest.h
```diff
@@ -196,19 +196,35 @@ enum
...
+};
+
 struct ContestPokemon
 {
     /*0x00*/ u16 species;
     /*0x02*/ u8 nickname[POKEMON_NAME_LENGTH + 1];
-    /*0x0D*/ u8 trainerName[8];
+    /*0x0D*/ u8 trainerName[PLAYER_NAME_LENGTH + 1];
     /*0x15*/ u8 trainerGfxId;
+    // Temp note: 2 bytes of padding
```
huderlem: remove the "temp note" comments

include/sprite.h
```diff
@@ -151,6 +151,26 @@ enum
     SUBSPRITES_IGNORE_PRIORITY, // on but priority is ignored
 };
 
+#define OAM_SIZE_0 (0)
+#define OAM_SIZE_1 (1 << 0)
+#define OAM_SIZE_2 (1 << 1)
+#define OAM_SIZE_3 (1 << 2)
+
+#define SPRITE_SIZE_8x8 .shape = ST_OAM_SQUARE, .size = OAM_SIZE_0
```
huderlem: I like this.

huderlem: Hmm, actually I don't really like the current use of it, since it's hiding two fields when making the struct definition, and look rather out of place to me.  I'd rather see something like this, I think:
```
.shape = OAM_SHAPE(SHAPE_32x64),
.size = OAM_SIZE(SIZE_32x64),
// or maybe
.shape = OAM_SHAPE(32, 64),
.size = OAM_SIZE(32, 64),
```

Plus these should be used in `struct OamData` definitions, but this PR is not actually using them.

src/data/contest_opponents.h
```diff
@@ -0,0 +1,3066 @@
...
+    }
+};
+
+const struct ContestPokemon gContestOpponents[] =
+{
+    [CONTEST_OPPONENT_JIMMY] = {
+        .species = SPECIES_POOCHYENA,
+        .nickname = _("POOCHY"),
+        .trainerName = _("JIMMY"),
+        .trainerGfxId = 7,
```
huderlem: Use `EVENT_OBJ_GFX_`* constants for `.trainerGfxId` field.

src/data/contest_opponents.h
```diff
@@ -0,0 +1,3066 @@
...
+        .personality = 0,
+        .otId = 0xFFFF
+    },
+    [CONTEST_OPPONENT_MAYRA] = {
+        .species = SPECIES_ALTARIA,
+        .nickname = _("TARIA"),
+        .trainerName = _("MAYRA"),
+        .trainerGfxId = 18,
+        .flags = 0x80400FFF,
+        .whichRank = 3,
```
huderlem: Use the contest rank enum you created.

src/data/contest_opponents.h
```diff
@@ -0,0 +1,3066 @@
...
+        .moves =
+        {
+            MOVE_MEAN_LOOK,
+            MOVE_FAINT_ATTACK,
+            MOVE_KNOCK_OFF,
+            MOVE_CONFUSE_RAY
+        },
+        .cool = 25,
+        .beauty = 35,
+        .c
```
huderlem: There are some tabs vs. spaces going on in this file.

src/data/contest_text_tables.h
```diff
@@ -0,0 +1,414 @@
...
+extern const u8 gText_0827E894[];
+extern const u8 gText_0827E89E[];
+extern const u8 gText_0827E8AA[];
+extern const u8 gText_0827E8B4[];
+extern const u8 gText_0827E8BF[];
+extern const u8 gText_0827E8CA[];
+
+const u8 *const gContestEffectDescriptionPointers[] =
+{
+    gContestEffect00hDescription,
```
huderlem: some more tabs vs. spaces

include/sprite.h
Phlosioneer: Yeah, I was iffy about the details.
Oh, that's a good idea. Combining them and then splitting them like we do for map numbers / map groups.

And I didn't do it for Oam Data stuff specifically because I wasn't happy about it hiding two fields. I can update all the Oam stuff to use the new system, but that's probably best saved as a separate PR.

Phlosioneer:
Waiting for #610 to land before updating this.

include/sprite.h
garakmon: how do you feel about these? they would require people to understand the allowable GBA sprite dimensions because you can't just put any number for width and height and expect good behavior, 
```
#define OAM_SHAPE(width, height) \
    (width == height ? ST_OAM_SQUARE : \
    (width < height ? ST_OAM_V_RECTANGLE : ST_OAM_H_RECTANGLE))

#define OAM_SIZE(width, height) \
    (width * height <= 128 ? OAM_SIZE_0 : \
    (width * height == 256 ? OAM_SIZE_1 : \
    (width * height <= 1024 ? OAM_SIZE_2 : OAM_SIZE_3))

```

Phlosioneer: Oh, I've already solved this problem and updated this branch. I must have forgotten to push.

I did it almost like that.

Phlosioneer:
This PR is updated.

PR #608, opened 2019-03-10
Add a constant for "no music" on maps.

Phlosioneer:


huderlem:
There are a couple places in `src/sound.c` that should use this constant.

Phlosioneer:
This PR is updated.

PR #612, opened 2019-03-13
Document the water droplet data.

Phlosioneer:


src/title_screen.c
```diff
@@ -64,7 +64,22 @@ static const u32 sTitleScreenRayquazaTilemap[] = INCBIN_U32("graphics/title_scre
 static const u32 sTitleScreenLogoShineGfx[] = INCBIN_U32("graphics/title_screen/logo_shine.4bpp.lz");
 static const u32 sTitleScreenCloudsGfx[] = INCBIN_U32("graphics/title_screen/clouds.4bpp.lz");
 
-const u16 gIntroWaterDropData[] =
+// This is BLDALPHA data. It's equivalent to:
```
garakmon: can you not use this macro here then?

Phlosioneer: I can; but expanding it out would be 32 lines, so I wasn't sure Shanty/Diego would be happy with that.

garakmon: Well _I_ would prefer the macro. Also, I know in modern C you can initialize ranges like so:
```
int array[] = {
    [1] = 0x0110,
    [2 ... 100] = 0x1234,
    [101] = 0x1000,
    ....
}
```
which you can probably use with agbcc to make the end of that array a single line

Phlosioneer:
Updated this PR.

src/title_screen.c
```diff
@@ -64,16 +64,41 @@ static const u32 sTitleScreenRayquazaTilemap[] = INCBIN_U32("graphics/title_scre
...
+    BLDALPHA_BLEND(8, 16),
+    BLDALPHA_BLEND(7, 16),
+    BLDALPHA_BLEND(6, 16),
+    BLDALPHA_BLEND(5, 16),
+    BLDALPHA_BLEND(4, 16),
+    BLDALPHA_BLEND(3, 16),
+    BLDALPHA_BLEND(2, 16),
+    BLDALPHA_BLEND(1, 16),
+    BLDALPHA_BLEND(0, 16),
+    [32 ... 63] = BLDALPHA_BLEND(0, 16)
```
huderlem: I didn't know this syntax was a thing.

Phlosioneer: I didn't know either! Someone suggested it.

PR #648, opened 2019-04-02
Fix hardcoded array sizes

Phlosioneer:


Phlosioneer:
Added an explicit check of the hardcoded indices. Forgetting to change them is now a compiler error.

PR #652, opened 2019-04-02
Mevent server

PikalaxALT:
i'm guessing mevent_server_ish is the client end handler?


include/mevent_client.h
```diff
@@ -0,0 +1,33 @@
+#ifndef GUARD_MEVENT_SERVER_ISH_H
```
huderlem: Use updated header guard names

sym_ewram.txt
```diff
@@ -12,14 +12,11 @@
     .include "src/union_room_battle.o"
 	.include "src/mevent2.o"
 	.include "src/mevent_801BAAC.o"
+    .include "src/mevent_server.o"
```
huderlem: The rest of the file uses tabs

src/scrcmd.c
```diff
@@ -2217,7 +2217,7 @@ bool8 ScrCmd_checkmonobedience(struct ScriptContext *ctx)
 
 bool8 ScrCmd_cmdCF(struct ScriptContext *ctx)
```
huderlem: Can you document this function?  It appears `gotoram` is misnamed, and should be something like `returnram`.

PikalaxALT: Resolved in e30ea04

sym_ewram.txt
PikalaxALT: Resolved in e30ea04

include/mevent_client.h
PikalaxALT: Resolved in e30ea04

PR #654, opened 2019-04-02
Rematch trainer macro

Phlosioneer:
Simplify the rematch entries, explicitly name the maps they're using, and comment some hardcoded constants in `pokenav.s`.


src/battle_setup.c
```diff
@@ -230,86 +230,124 @@ static const struct TrainerBattleParameter sTrainerBContinueScriptBattleParams[]
...
-    [REMATCH_NORMAN] = {{TRAINER_NORMAN_1, TRAINER_NORMAN_2, TRAINER_NORMAN_3, TRAINER_NORMAN_4, TRAINER_NORMAN_5}, 0x0, 0x0},
-    [REMATCH_WINONA] = {{TRAINER_WINONA_1, TRAINER_WINONA_2, TRAINER_WINONA_3, TRAINER_WINONA_4, TRAINER_WINONA_5}, 0x0, 0x4},
-    [REMATCH_TATE_AND_LIZA] = {{TRAINER_TATE_AND_LIZA_1, TRAINER_TATE_AND_LIZA_2, TRAINER_TATE_AND_LIZA_3, TRAINER_TATE_AND_LIZA_4, TRAINER_TATE_AND_LIZA_5}, 0x0, 0x6},
-    [REMATCH_JUAN] = {{TRAINER_JUAN_1, TRAINER_JUAN_2, TRAINER_JUAN_3, TRAINER_JUAN_4, TRAINER_JUAN_5}, 0x0, 0x7},
-    [REMATCH_SIDNEY] = {{TRAINER_SIDNEY, TRAINER_SIDNEY, TRAINER_SIDNEY, TRAINER_SIDNEY, TRAINER_SIDNEY}, 0x0, 0x8},
-    [REMATCH_PHOEBE] = {{TRAINER_PHOEBE, TRAINER_PHOEBE, TRAINER_PHOEBE, TRAINER_PHOEBE, TRAINER_PHOEBE}, 0x0, 0x8},
-    [REMATCH_GLACIA] = {{TRAINER_GLACIA, TRAINER_GLACIA, TRAINER_GLACIA, TRAINER_GLACIA, TRAINER_GLACIA}, 0x0, 0x8},
-    [REMATCH_DRAKE] = {{TRAINER_DRAKE, TRAINER_DRAKE, TRAINER_DRAKE, TRAINER_DRAKE, TRAINER_DRAKE}, 0x0, 0x8},
-    [REMATCH_WALLACE] = {{TRAINER_WALLACE, TRAINER_WALLACE, TRAINER_WALLACE, TRAINER_WALLACE, TRAINER_WALLACE}, 0x0, 0x8},
+    [REMATCH_ROSE] = REMATCH(ROSE, 5, ROUTE118),
```
huderlem: I'm not a fan of this `REMATCH` macro implementation for a couple reasons:
1. `5` and `1` are the only valid options for the second parameter.
2. The macro can't be used by all rematch entries.

My opinion is that explicitly listing the 5 trainers is clearer and doesn't needlessly couple the trainer constant names with the rematch system.
```
[REMATCH_ROSE] = REMATCH(TRAINER_ROSE_1, TRAINER_ROSE_2, TRAINER_ROSE_3, TRAINER_ROSE_4, TRAINER_ROSE_5, ROUTE118),
```

Phlosioneer: Hmm... I would like to remove _some_ of the noise in the lists, though. We know that they have to be trainer names, so they have to have the `TRAINER_` prefix. What if it looked more like this:
```
[REMATCH_ROSE] = REMATCH(ROSE_1, ROSE_2, ROSE_3, ROSE_4, ROSE_5, ROUTE118),
```

huderlem: I see where you're coming from, but I'm not convinced it's noisy enough to warrant that.  I typically lean against truncating constants like that because it makes grepping harder and obscures what they really are.  The `MAP_GROUP` and `MAP_NUM` macros do that, and it trips me up on occasion.

Diegoisawesome: Agreed, and I keep coming across that same MAP_GROUP/NUM problem constantly.

DizzyEggg: I'm glad I'm not the only one who copied the whole MAP_GROUP_NUM string into the macro. 😛

huderlem:
This PR is still worth completing if anyone wants to pick it up.  The macros should be reworked to have the format suggested above:
```c
[REMATCH_ROSE] = REMATCH(TRAINER_ROSE_1, TRAINER_ROSE_2, TRAINER_ROSE_3, TRAINER_ROSE_4, TRAINER_ROSE_5, ROUTE118),
```

PikalaxALT:
Tasks remaining:
- Address comment by @huderlem above
- Resolve merge conflicts with master

Deadline: 07/31 16:00 UTC

PikalaxALT:
Closing due to inactivity

PR #656, opened 2019-04-03
Decompile some of menu_specialized 

ultima-soul:
2 NONMATCHING


DizzyEggg:
For statStoreLocation[X], could you use defines `STAT_HP` and similar?

PR #661, opened 2019-04-04
contest & contest_ai cleanup

Phlosioneer:
Not as thorough as I'd like, about halfway done. But I want to move on to other things.


include/contest.h
```diff
@@ -1,6 +1,24 @@
 #ifndef GUARD_CONTEST_H
 #define GUARD_CONTEST_H
 
+#include "palette.h"
```
Diegoisawesome: Why is this include in the header?

include/contest.h
```diff
@@ -408,30 +424,31 @@ struct ContestResourcesField20
...
-#define shared18000 (*(struct Shared18000 *)(gHeap + 0x1a000))
+#define eContest (*gContestResources->contest)
+#define eContestantStatus (gContestResources->status)
+#define eContestResources8 (*gContestResources->field_8)
+#define eContestAI (*gContestResources->aiData)
+#define eContestResources10 (*gContestResources->field_10)
+#define eContestResources14 (*gContestResources->field_14)
+#define eUnzippedContestAudienceGfx (gHeap + 0x18000)
+#define eUnknownHeap19000 (gHeap + 0x19000)
+#define eContestDebugMode (gHeap[0x1a000])
```
Diegoisawesome: Why is this bracketed when the others aren't?

src/contest_link_80FC4F4.c
```diff
@@ -72,22 +72,22 @@ static void sub_80FC5DC(u8 taskId)
 
     gContestPlayerMonIndex = GetMultiplayerId();
     gNumLinkContestPlayers = GetLinkPlayerCount();
-    gIsLinkContest = 1;
+    gLinkContestFlags = LINK_CONTEST_FLAG_IS_LINK;
     if (gWirelessCommType == 1)
-        gIsLinkContest = 3;
+        gLinkContestFlags = LINK_CONTEST_FLAG_IS_LINK | LINK_CONTEST_FLAG_IS_WIRELESS;
 
     for (i = 0; i < gNumLinkContestPlayers && (u32)(gLinkPlayers[i].version & 0xFF) - 1 > VERSION_RUBY - 1; i++)
```
Diegoisawesome: You should be able to remove both those subtractions and cast and still have things work.

include/contest.h
```diff
@@ -1,6 +1,24 @@
...
+#define APPLAUSE_METER_SIZE 5
+#define CONTEST_TURN_COUNT 5
+
+#define CONTEST_DEBUG_MODE_OFF 0
+// Prints the totalPoints value for each contestant.
+#define CONTEST_DEBUG_MODE_PRINT_POINT_TOTAL 1
+// Prints the ContestResourcesField1C::unk_C value as a bitstring for each contestant.
+#define CONTEST_DEBUG_MODE_PRINT_UNK_C 2
+// Prints the ContestResourcesField1C::unk_D value as a bitstring for each contestant.
+#define CONTEST_DEBUG_MODE_PRINT_UNK_D 3
```
Diegoisawesome: These could be an enum.

src/contest.c
```diff
@@ -45,18 +45,42 @@
...
+#define CONTEST_WINDOW_GENERAL_TEXT 4
+// The available moves, from top to bottom
+#define CONTEST_WINDOW_MOVE0 5
+#define CONTEST_WINDOW_MOVE1 6
+#define CONTEST_WINDOW_MOVE2 7
+#define CONTEST_WINDOW_MOVE3 8
+// The small "/" character between the move category and the
+// appeal/jam display
+#define CONTEST_WINDOW_SLASH 9
+#define CONTEST_WINDOW_MOVE_DESCRIPTION 10
```
Diegoisawesome: These could be an enum

src/contest.c
```diff
@@ -2560,15 +2588,15 @@ static void sub_80DA7A0(u8 taskId)
...
     gTasks[taskId].func = TaskDummy1;
-    sub_80DBF68();
+    ContestPrintLinkStandby();
     sub_80DC490(FALSE);
 }
 
 static void sub_80DA7EC(u8 taskId)
 {
     DestroyTask(taskId);
     BeginNormalPaletteFade(0xFFFFFFFF, 0, 0, 16, 0);
```
Diegoisawesome: RGB_BLACK

src/contest.c
```diff
@@ -4538,29 +4655,29 @@ static void task08_080CD1CC(u8 taskId)
...
 
-static void sub_80DDCDC(s8 a)
+static void ShowAndUpdateApplauseMeter(s8 unused)
 {
-    u8 taskId = CreateTask(sub_80DDD20, 5);
+    u8 taskId = CreateTask(Task_ShowAndUpdateApplauseMeter, 5);
 
-    gTasks[taskId].data[0] = a;
-    sContest.unk1920A_5 = 1;
+    gTasks[taskId].data[0] = unused;
```
Diegoisawesome: `data[0]` may not be used but the local variable here is.

src/contest_effect.c
```diff
@@ -82,101 +82,101 @@ static void ContestEffect_HighlyAppealing(void)
...
 static void ContestEffect_UserLessEasilyStartled(void)
 {
-    sContestantStatus[shared192D0.contestant].resistant = TRUE;
-    SetContestantEffectStringID(shared192D0.contestant,CONTEST_STRING_STOPPED_CARING);
+    eContestantStatus[eContestResources8.contestant].resistant = TRUE;
+    SetContestantEffectStringID(eContestResources8.contestant,CONTEST_STRING_STOPPED_CARING);
 }
 
-// Slightly startles the POK�MON in front.
+// Slightly startles the POK�MON in front.
```
Diegoisawesome: Something happened with the encoding of the `é` here.

src/contest_effect.c
```diff
@@ -227,141 +227,141 @@ static void ContestEffect_StartlePrevMons2(void)
...
     }
-    SetContestantEffectStringID(shared192D0.contestant, CONTEST_STRING_ATTEMPT_STARTLE);
+    SetContestantEffectStringID(eContestResources8.contestant, CONTEST_STRING_ATTEMPT_STARTLE);
     if (numStartled == 0)
-        SetContestantEffectStringID2(shared192D0.contestant, CONTEST_STRING_MESSED_UP2);
+        SetContestantEffectStringID2(eContestResources8.contestant, CONTEST_STRING_MESSED_UP2);
 }
 
-// Shifts the JUDGE�s attention from others.
+// Shifts the JUDGE�s attention from others.
```
Diegoisawesome: A junk character was inserted here.

include/contest.h
Phlosioneer: Because it's a value, not a pointer.

src/contest_effect.c
Phlosioneer: Can we replace these? I've actually looked at the source file with a hex editor and it's not valid UTF-8.

Phlosioneer: Some more details:
The hex of the original comment is `50 4F 4B E9 4D 4F 4E`
50 = P
4F = O
4B = K
E9 = Start of 3-byte codepoint
4D = M
4F = O
4E = N

The correct hex for é is `C3 A9`. The confusion appears to be because the codepoint for é is 223, or 0xE9. But that by itself is not valid, as the highest-order bits would be set, indicating a multi-byte sequence.

Browsers are known to accept errors like this, and treat incorrectly-encoded UTF-8 as literal codepoint values (Also known as USC8).

include/contest.h
Phlosioneer: Because the `PLTT_BUFFER_SIZE` constant is used as part of a struct definition.

src/contest_link_80FC4F4.c
Phlosioneer: No, it fails to match without the cast and subtractions. Both are necessary.

src/contest.c
Phlosioneer: But someone looking at the arguments to the function will want to know that the argument doesn't matter at all, right?

Phlosioneer:
This PR is updated.

huderlem:
merge conflicts

Phlosioneer:
This PR is updated.

PR #679, opened 2019-04-08
Decompile some pokenav_unk_2

ultima-soul:
2 NONMATCHING


src/pokenav_unk_2.c
```diff
@@ -0,0 +1,109 @@
...
+
+void sub_81C9990(u32 a)
+{
+	struct UnknownPokenavSubstruct2 *ptr;
+	ptr = GetSubstructPtr(2);
+	ptr->field_4 = CreateLoopedTask(gUnknown_086201A0[a], 1);
+	ptr->callback = sub_81C99FC;
+}
+
+void *sub_81C99C0
```
ProjectRevoTPP: what happened here

garakmon:
I think this will match
```C
extern const u32 gUnknown_08620240[];
extern void sub_81CA0EC(const unsigned *, u16, u16);

void sub_81CA0C8(void)
{
    unsigned var = sub_81C9894() * 7;
    sub_81CA0EC(
        (gUnknown_08620240 + var),
        *(gUnknown_08620240 + var - 1) & 0xFFFF,
        *(gUnknown_08620240 + var - 1) >> 16
    );
}
```

PR #680, opened 2019-04-08
Pokenav unk 1 Decompiled

Phlosioneer:
About halfway done with the file.


Phlosioneer:
I'm taking a break from pokeemerald for a while, so I'm going to stop here. There's only one function left to decompile.

This file seems to be the callbacks to make the main menu UI work.

PR #687, opened 2019-04-14
remove some statics from src/data files

garakmon:


DizzyEggg:
Why? I'm of the opinion that if a variable/function can be static, it should. Unless you give a reason for removing statics, I am against this PR.

huderlem:
Shelving this PR for now--my fault for the knee-jerk proposal.  May revisit this in the future.

PR #696, opened 2019-04-24
Make installation documentation more straightforward

SatoMew:


garakmon:
Now that pokeemerald (and all of the pret repos I think) are templates, could you add a note about creating a github profile then clicking the "use this template" button.
Also maybe add a third note about how on macOS you might need to `echo "if [ -f ~/.bashrc ]; then . ~/.bashrc; fi" >> ~/.bash_profile` if the environment variables aren't getting set with new sessions.

Otherwise this should probably be merged asap.

huderlem:
Waiting for the modern work to land before coming back to this documentation update.

PikalaxALT:
Please resolve merge conflicts

PR #698, opened 2019-04-27
Document MON_DATA_10 and magic number in atk04_critcalc

hjk321:
Hello, I have determined the purpose of the unknown constant MON_DATA_10. It is not used as part of the actual BoxPokemon data structure, but is instead used in comparisons when determining whether to treat a field as encrypted data. All fields with values lower than 10 are unencrypted, anything above is encrypted. As such I propose it be renamed to MON_DATA_ENCRYPT_SEPARATOR in the interest of documentation.

I have also changed some code in atk04_critcalc at the suggestion of pret discord members to remove a magic number and replace it with an expression that makes much more sense in context.

Build passes. This is my first real pret contribution so if I did something that doesn't conform to standards, feel free to let me know.

### DIFF
--- a/src/battle_script_commands.c
+++ b/src/battle_script_commands.c
@@ -1277,7 +1277,7 @@ static void atk04_critcalc(void)
                 + 2 * (holdEffect == HOLD_EFFECT_STICK && gBattleMons[gBattlerAttacker].species == SPECIES_FARFETCHD);
 
     if (critChance > ARRAY_COUNT(sCriticalHitChance) - 1 )
-    critChance = ARRAY_COUNT(sCriticalHitChance) - 1;
+		critChance = ARRAY_COUNT(sCriticalHitChance) - 1;
 
     if ((gBattleMons[gBattlerTarget].ability != ABILITY_BATTLE_ARMOR && gBattleMons[gBattlerTarget].ability != ABILITY_SHELL_ARMOR)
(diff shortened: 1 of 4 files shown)


src/battle_script_commands.c
```diff
@@ -1276,8 +1276,8 @@ static void atk04_critcalc(void)
                 + 2 * (holdEffect == HOLD_EFFECT_LUCKY_PUNCH && gBattleMons[gBattlerAttacker].species == SPECIES_CHANSEY)
                 + 2 * (holdEffect == HOLD_EFFECT_STICK && gBattleMons[gBattlerAttacker].species == SPECIES_FARFETCHD);
 
-    if (critChance > 4)
-        critChance = 4;
+    if (critChance > ARRAY_COUNT(sCriticalHitChance) - 1 )
+    critChance = ARRAY_COUNT(sCriticalHitChance) - 1;
```
huderlem: Please fix this indentation error.  Everything else looks fine.

hjk321: Nice catch. Fixed!

DizzyEggg: You could also change if (critChance >= ARRAY_COUNT(sCriticalHitChance)), makes more sense this way imo

hjk321: @DizzyEggg That's what I thought too, but I had figured it wouldn't match. Turns out it does! I didn't realize the compiler worked that way. Thanks!

PR #699, opened 2019-04-28
[WIP] Add debug output if scaninc dependency fails

SBird1337:
Normally if a dependency injected by scaninc fails, the error message issued by `make` is very misleading, as it only talks about the root file. This PR adds fallback rules that issue more detailed errors about the actual file that was not found, as well as the root file that was the actual build target.


SBird1337:
Does not work, because make rules cannot be prioritized like that

PR #708, opened 2019-05-14
Base stats abilities as array

DizzyEggg:


src/data/battle_frontier/battle_pyramid_level_50_wild_mons.h
```diff
@@ -3,49 +3,49 @@ static const struct PyramidWildMon sLevel50WildMons_Round1[] =
     {
         .species = SPECIES_PLUSLE,
         .lvl = 35,
-        .abilityBit = 2,
+        .abilityNum = 2,
```
huderlem: `abilityNum = 2`?  Why is this the case?  This might indicate that the array approach is not good, after all.

DizzyEggg: Nah, it means that the ability is random(if it has two) instead of the specified one.

huderlem: Can you create a `define` or it while we're here?

PR #709, opened 2019-05-14
name some trainer card related functions

garakmon:
courtesy @Sewef / Citron#3400 on discord


src/pokemon.c
```diff
@@ -1879,7 +1879,7 @@ static const u16 sDeoxysBaseStats[] =
     90, // Sp.Defense
 };
 
-const u16 gUnknown_08329D54[] =
+const u16 gFacilityClassPicsByIndex[] =
```
huderlem: This array doesn't seem to contain pics, so I don't think the name is correct.  Plus the call below no longer makes sense with this name:
```
return FacilityClassToPicIndex(gFacilityClassPicsByIndex[arrId]);
```

garakmon: renamed to `gLinkPlayerFacilityClasses`

PR #723, opened 2019-06-03
add unobtrusive pretty printer

nicholatian:
this change introduces, through a Makefile macro in `comwrap.mk`, ANSI-based pretty printing of build output.

- it is disabled on Windows.
- it can be turned off with `NOCOLOR=1`, reverting to make’s default noisy behaviour.
- if any job command talks (through `stdout` or `stderr`), the pretty printer will send that through along with an echo of the noisy command. otherwise it’s silent.

Every job command is put onto a single line, with the action taken on the left hand side and the file being processed in bold after it.

I introduced a simple mix of colour coding + labeling for the different kinds of stuff taken care of during build. C compilation is blue, linkage and other adjusting things are red, data conversion + transmogrification are in yellow, and sound stuff is in purple/magenta. Sometimes those boundaries are intermixed, like when music is assembled, in which case the action is Assembling but it’s put out purple.

Using the comwrap in the Makefile is simple enough: it involves a macro `$(call)`, where `$(1)` is your command, and `$(2)` is what you’re doing (`comwrap.mk` provides macro constants for all of what’s done). For instance:

```make
$(call COMWRAP,$(OBJCOPY) -O binary $< $@,$(COMWRAP_OCPY))
$(call COMWRAP,$(FIX) $@ -p -t'$(TITLE)' -c$(GAME_CODE) -m$(MAKER_CODE) -r$(REVISION) --silent,$(COMWRAP_FIX))
```

There is only one gotcha to bear in mind, and it’s a limitation of Make itself: you can’t use commas as-written in the commands. Getting around this is easy though, `comwrap.mk` exports `$(COMMA)` for convenience, just use that instead and it’ll resolve fine.


huderlem:
I can't say I'm really a fan of this.  One platform(?) gets some colors at the expense of cluttering and obscuring the entire makefile and its accompanying rules.

nicholatian:
It runs on all \*nix platforms, so really only excluding Windows (I could implement that but I don’t think it’s worth it, comparatively speaking).

As far as readability goes, it’s hard to say it could be done any better without preprocessing the Makefiles. Right now it’s inline, but you could also do
```make
$(ROM): $(ELF)
	$(call COMWRAP,\
	$(OBJCOPY) -O binary $< $@\
	,$(COMWRAP_OCPY))
	# ...
```
to give it more readability.

huderlem:
My opinion is that this is still too obtrusive for the little benefit it provides, even after the above suggested formatting improvements.  It could be a nice addition for a personal project, but I anticipate that this would cause too many headaches to be worth merging upstream.

PR #726, opened 2019-06-12
Metatile Labels

garakmon:
Add file `include/constants/metatile_labels.h`
Name metatiles that are used with `MapGridSetMetatileIdAt()` or `setmetatile`


src/fldeff_misc.c
```diff
@@ -800,16 +801,16 @@ static void Task_SecretBasePCTurnOn(u8 taskId)
     {
     case 4:
     case 12:
-        MapGridSetMetatileIdAt(data[0], data[1], 548);
+        MapGridSetMetatileIdAt(data[0], data[1], 0x224);
```
huderlem: Looks like you forgot about the ones in this file?

garakmon: The secret base ones I didn't name because I wasn't exactly sure how to handle it.
I though about naming them as being a part of the `gSecretBase` primary tileset, I don't remember why I didn't.

PR #727, opened 2019-06-16
Delete inaccurate comment.

JaceCearK1:


DizzyEggg:
They don't, only Mew and Deoxys are checked. Did you look at the function?

huderlem:
The function should be renamed to something like `ShouldCheckObedienceByLevel()`

JaceCearK1:
> 
> 
> They don't, only Mew and Deoxys are checked. Did you look at the function?

The thing is that internally the obedience flag is never used on Pokémon other than Mew/Deoxys, that's why the function returns TRUE with any other Pokémon. The name @huderlem came up with sounds like a better name for the function, I think.

huderlem:
We were discussing this on Discord, and noticed that the Mew/Deoxys check was likely accidentally inverted by Game Freak.  If the check is inverted, then you're left with perfectly-functional system that allows forced obedience.  (Some unused script commands like `checkobedience` support this idea.)  For now, the function name change is sufficient, and maybe add a note about the possible bug (inverted Mew/Deoxys check)

huderlem:
Update: Forget what I said before--this check is an intended disobedience check for Mew/Deoxys, as a crude anti-cheat measure.  (I was misreading the code yesterday.)  Regardless, the function should still be renamed to `ShouldCheckObedienceByLevel` or similar.

JaceCearK1:
Alright, I kind of assumed this was common knowledge around Gen 3 hackers, since many people used the Wild Encounter cheats with an Action Replay back in the day, which made this anti-cheat measure occur.

Always explain the full story of everything you know, lesson learnt! 😄

PikalaxALT:
@JaceCearK1 Please either follow @huderlem's suggestion or close this PR.

JaceCearK1:
Since I don't know how to append another file change to a pull request, I'll just close it and someone else who's more "In the know" can make the changes. :)

PR #728, opened 2019-06-16
Make devkitARM optional

froggestspirit:
This change should not affect anyone currently using devkit. This should enable users to compile without devkit if the arm-none-eabi compiler is present.

Tested working on Windows by Kurausukun with devkit installed
Tested working on 64-bit Debian by FroggestSpirit without devkit after running "sudo apt-get install gcc-arm-none-eabi binutils-arm-none-eabi"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,9 @@ include $(DEVKITARM)/base_tools
 else
 PREFIX		:=	arm-none-eabi-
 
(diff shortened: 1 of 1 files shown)


froggestspirit:
This should fix the previous build error

huderlem:
This PR should be obsolete after modern work lands.

PR #735, opened 2019-06-30
Add Different Field Info to Wild Encounters JSON

garakmon:
So a `fields` object describes the names ("`type`") of each field, and has an "`encounter_rates`" array which is the length of the number of pokemon to expect and their encounter rates. The purpose of this is to make modifying the wild encounters as easy as possible. My idea is that modifying "`fields`" will update every encounter object accordingly (through tools like porymap). The requirement is that for each field type, there needs to be the same number of encounterable pokemon. However, if a person wants fewer encounters than the length of this array in certain instances for whatever reason, it could be done with trivial modifications to the source.

If these changes are admissible, then of course I'll do the same for pokeruby.

### DIFF
--- a/src/data/wild_encounters.json.txt
+++ b/src/data/wild_encounters.json.txt
@@ -1,5 +1,16 @@
 {{ doNotModifyHeader }}
+
 ## for wild_encounter_group in wild_encounter_groups
+{% if wild_encounter_group.for_maps %}
+## for wild_encounter_field in wild_encounter_group.fields
+## for encounter_rate in wild_encounter_field.encounter_rates
+{% if trackVar(encounter_rate, 100) %}
+#define ENCOUNTER_CHANCE_{{ upper(wild_encounter_field.type) }}_INDEX_{{ loop.index }} {{ encounter_rate }} {% else %}#define ENCOUNTER_CHANCE_{{ upper(wild_encounter_field.type) }}_INDEX_{{ loop.index }} ENCOUNTER_CHANCE_{{ upper(wild_encounter_field.type) }}_INDEX_{{ subtract(loop.index, 1) }} + {{ encounter_rate }}{% endif %}
+## endfor
+
+## endfor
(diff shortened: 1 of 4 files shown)


src/wild_encounter.c
```diff
@@ -105,7 +105,7 @@ static bool8 CheckFeebas(void)
         if (y >= gRoute119WaterTileData[3 * 2 + 0] && y <= gRoute119WaterTileData[3 * 2 + 1])
             route119Section = 2;
 
-        if (Random() % 100 > 49) // 50% chance of encountering Feebas
```
garakmon: put this comment back

src/data/wild_encounters.json.txt
```diff
@@ -1,5 +1,16 @@
 {{ doNotModifyHeader }}
+
 ## for wild_encounter_group in wild_encounter_groups
+{% if wild_encounter_group.for_maps %}
+## for wild_encounter_field in wild_encounter_group.fields
+## for encounter_rate in wild_encounter_field.encounter_rates
+{% if trackVar(encounter_rate, 100) %}
+#define ENCOUNTER_CHANCE_{{ upper(wild_encounter_field.type) }}_INDEX_{{ loop.index }} {{ encounter_rate }} {% else %}#define ENCOUNTER_CHANCE_{{ upper(wild_encounter_field.type) }}_INDEX_{{ loop.index }} ENCOUNTER_CHANCE_{{ upper(wild_encounter_field.type) }}_INDEX_{{ subtract(loop.index, 1) }} + {{ encounter_rate }}{% endif %} {{ setVarInt(wild_encounter_field.type, loop.index) }}
```
PikalaxALT: `_SLOT_` instead of `_INDEX_` perhaps?

garakmon: I agree this is better, it's the word I used in porymap

garakmon:
fixed

PR #738, opened 2019-07-01
Match unref_sub_8119094

DizzyEggg:
Gotta match these unused functions, am I right?

### DIFF
--- a/src/battle_intro.c
+++ b/src/battle_intro.c
@@ -205,7 +205,7 @@ static void BattleIntroSlide1(u8 taskId)
 
         if (gBattle_WIN0V & 0xFF00)
             gBattle_WIN0V -= 0x3FC;
-        
+
         if (gTasks[taskId].data[2])
             gTasks[taskId].data[2] -= 2;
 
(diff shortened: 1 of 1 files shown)


src/battle_intro.c
```diff
@@ -600,129 +600,18 @@ void sub_8118FBC(int bgId, u8 arg1, u8 arg2, u8 battlerPosition, u8 arg4, u8 *ar
...
+    int i, j, offset;
+
     DmaCopy16(3, gMonSpritesGfxPtr->sprites[battlerPosition] + BG_SCREEN_SIZE * arg3, (void *)BG_SCREEN_ADDR(0) + arg5, BG_SCREEN_SIZE);
     offset = (arg5 >> 5) - (arg7 << 9);
     for (i = arg1; i < arg1 + 8; i++)
     {
         for (j = arg0; j < arg0 + 8; j++)
         {
-            ((u16 *)BG_VRAM)[i * 32 + j + (arg6 * 0x400) + arg0] = offset | (arg4 << 12);
+            *((u16 *)(BG_VRAM) + (i * 32) + (j + (arg6 << 10))) = offset | (arg4 << 12);
```
PikalaxALT: Wonder if there's a combination like this that will fix that last remaining function in [asm/decompress.s](/pret/pokeemerald/blob/master/asm/decompress.s) 🤔

DizzyEggg: I attempted that one many times, but couldn't get it right, this one is way simpler.

PR #744, opened 2019-07-10
Remove unnecessary tutor learnset casts

PikalaxALT:


DizzyEggg:
Against it, harder for romhackers to expand to 64 tutor moves this way

PikalaxALT:
Disagree with @DizzyEggg's comment. Romhackers can replace the `1u` with `1lu` and the type of sTutorLearnsets with `u64`.

GriffinRichards:
You'd have to change the typecast in TUTOR_LEARNSET to expand to 64 anyway, having mismatched typecasts is just more confusing imo

PR #745, opened 2019-07-11
Use a constant instead of integer literals when checking for shinyness

Gudf:
Even though there are functions to determine whether a pokemon is shiny or not, those are not used every time the game needs to check for shininess, which makes it tedious to edit shiny rate since you need to edit every extra check that isn't done using IsMonShiny() or IsShinyOtIdPersonnality(), instead of the one in IsShinyOtIdPersonnality() only. 
This PR makes it so the value against which the games checks to determine shininess is defined in only one place, which makes editing shiny rate much more practical.

### DIFF
--- a/include/constants/pokemon.h
+++ b/include/constants/pokemon.h
@@ -82,4 +82,7 @@
 #define NUM_STATS 6
 #define NUM_BATTLE_STATS 8
 
+// Shiny odds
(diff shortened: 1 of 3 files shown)


DizzyEggg:
That's pretty cool, I like that

PR #749, opened 2019-07-22
Fix PREPARE_MOVE_BUFFER macro

Sewef:
Will take u16 move as expected


PikalaxALT:
The following change is required:
- Use a bitmask instead of a typecast

PR #762, opened 2019-08-02
Documenting MAPSEC use in pokenav match call data

GriffinRichards:
Wally and Prof. Birch are skipped. Wally has a variable location, and Prof Birch (despite being listed as .v1 = 0, which is Littleroot) is actually listed as location unknown in the pokenav. Rival is also skipped, as they dont have a location field and are also listed as unknown

### DIFF
--- a/include/pokenav.h
+++ b/include/pokenav.h
@@ -107,7 +107,7 @@ void sub_81C8234(void);
 
 // pokenav_match_call_data.c
 bool32 sub_81D17E8(u32 idx);
-u8 sub_81D16DC(u32 idx);
+u8 MatchCallMapSecGetByIndex(u32 idx);
 bool32 sub_81D1BF8(u32 idx);
 bool32 MatchCallFlagGetByIndex(u32 idx);
 u32 MatchCall_GetRematchTableIdx(u32 idx);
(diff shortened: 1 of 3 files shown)


DizzyEggg:
Could you rename the field as well? from v1 to mapSec?

GriffinRichards:
Whoops didn't name the commit. But yes updated to mapSec. I was hesitant because it's not totally clear to me how MatchCallStruct1, MatchCallStruct2 and MatchCallStruct3 actually treat .v1. Also I noticed I didn't update sWallyAdditionalData with MAPSEC constants, about to commit that

GriffinRichards:
Yeah I'm not convinced v1 in MatchCallStructCommon and MatchCallStruct2 at least should be mapSec. For MatchCallStruct2 it looks like v2 in MatchCallSubstruct2 is its mapsec, which is fetched from sWallyAdditionalData in v10. I'm going to revert the v1's that aren't directly assigned a MAPSEC.

GriffinRichards:
Tacking on a fix to the MatchCall##Func order. Thanks to MatchCallGetFunctionIndex, index != type, so 3/4 are swapped

PR #764, opened 2019-08-03
Document vars that are used and unnamed

GriffinRichards:
All remaining used and unnamed vars documented except VAR_0x402E, which seems to be read in a function that's never called.


include/constants/vars.h
```diff
@@ -113,7 +113,7 @@
 #define VAR_ROUTE101_STATE                   0x4060
 #define VAR_UNUSED_0x4061                    0x4061 // Unused Var
```
PikalaxALT: Could probably interpolate what these unused vars could have been for

include/constants/vars.h
```diff
@@ -94,7 +94,7 @@
 #define VAR_POKELOT_PRIZE_PLACE              0x404D
 #define VAR_UNUSED_0x404E                    0x404E // Unused Var
 #define VAR_LOTAD_SIZE_RECORD                0x404F
-#define VAR_0x4050                           0x4050
+#define VAR_LITTLEROOT_STATE                 0x4050
 #define VAR_ROUTE102_ACCESSIBLE              0x4051
```
PikalaxALT: `VAR_OLDALE_TOWN_STATE`

include/constants/vars.h
```diff
@@ -94,7 +94,7 @@
...
 #define VAR_UNUSED_0x4052                    0x4052 // Unused Var
```
PikalaxALT: `VAR_DEWFORD_TOWN_STATE`

include/constants/vars.h
PikalaxALT: Example: `VAR_ROUTE102_STATE` here

include/constants/vars.h
GriffinRichards: That's inconsistent with the other var names. Should I rename the rest as well?

include/constants/vars.h
GriffinRichards: That's definitely Littleroot

include/constants/vars.h
GriffinRichards: Yup, they clearly allocated a block of vars for route/town states

include/constants/vars.h
PikalaxALT: The comment's on the var currently named `VAR_ROUTE102_ACCESSIBLE`, which is linked to the dude checking his own footprints.

include/constants/vars.h
PikalaxALT: Sure. Also, find a different name for `VAR_SOOTOPOLIS_STATE` because the var currently named `VAR_RAYQUAZA_STATE` is logically the Sootopolis State variable.

PR #769, opened 2019-08-05
Pokenav unk 9

PikalaxALT:


src/pokenav_unk_9.c
```diff
@@ -1,68 +1,91 @@
...
+struct PokenavSub10
 {
-    sub_81CFB8C,
-    sub_81CFC2C,
-    sub_81CFC40
+    bool32 (*callback)(void);
+    u32 ltid;
+    u16 winid;
+    bool32 unkC;
+    u16 buff[0x400];
```
DizzyEggg: Change to u8 and use BG_SCREEN_SIZE constant

src/pokenav_unk_9.c
```diff
@@ -1,68 +1,91 @@
...
+static u32 sub_81D01DC(s32 state);
+static u32 sub_81D021C(s32 state);
+static void sub_81D02B0(s32 windowId, s32 val1, s32 val2);
+static void sub_81D024C(struct PokenavSub10 *ptr);
+static void sub_81D0288(struct PokenavSub10 *ptr);
+static void sub_81D0304(void);
+static void sub_81D035C(struct PokenavMonList *, u8 *);
 
-const struct BgTemplate gUnknown_086237B0 =
+static u32 (*const gUnknown_086235D8[])(s32 state) =
```
DizzyEggg: Why not use static const LoopedTask gUnknown_086235D8[]?

src/pokenav_unk_9.c
```diff
@@ -163,18 +186,541 @@ u32 sub_81CFA88(struct PokenavSub9 *structPtr)
...
+    FreePokenavSubstruct(10);
+}
+
+static u32 sub_81CFEB8(s32 state)
+{
+    struct PokenavSub10 * unk = GetSubstructPtr(10);
+    switch (state)
+    {
+    case 0:
+        InitBgTemplates(gUnknown_086237B0, NELEMS(gUnknown_086237B0));
```
DizzyEggg: Why NELEMS even exists, if we have ARRAY_COUNT?

PikalaxALT: As discussed in the Discord, `NELEMS` is a GameFreak macro which appears in FireRed and Emerald-J asserts. It is identical to `ARRAY_COUNT`, and I aliased it as such in `global.h`.

PR #777, opened 2019-08-13
Copy Berry Fix MB dism from FR

PikalaxALT:


DizzyEggg:
Can't the berry fix and pokeemerald both share the same headers?

PR #779, opened 2019-08-15
Deprecate build_tools.sh

PikalaxALT:


Diegoisawesome:
Looks good to me, is there any reason this isn't merged in yet?

build_tools.sh
```diff
@@ -1,12 +1,8 @@
...
-make -C tools/aif2pcm CXX=${1:-g++}
-make -C tools/ramscrgen CXX=${1:-g++}
-make -C tools/gbafix CXX=${1:-g++}
-make -C tools/mid2agb CXX=${1:-g++}
-make -C tools/mapjson CXX=${1:-g++}
-make -C tools/jsonproc CXX=${1:-g++}
+
+echo "This script is deprecated. Run \"make tools\" instead."
+for dname in tools/*; do
+  if [[ -f ${dname}/Makefile ]]; then
```
dannye: `[[` is not POSIX. You should either avoid `[[`or change `#!/bin/sh`

PR #787, opened 2019-08-23
Rename table cmds

GriffinRichards:
Prefix BattleAICmd_ to Cmd_, prefix atk##_ to Cmd_, and renames constants/states that use atk##


DizzyEggg:
I'd leave the macros as they are, they've become too long and generic(DONT_CHECK_STATUSES for example).

DizzyEggg:
All in all, I'm actually not sure if it's a good idea changing these. I think a better alternative would be renaming all of the commands

cmdXX_sthsth

GriffinRichards:
I think if we're doing that we should keep BattleAICmd and do something like BattleScrCmd for the others. Having multiple Cmd23_ for ex might be confusing. I'm also still not sure there's any benefit to including the table byte in the name

DizzyEggg:
I'm against long names, and don't see how multiple Cmd23_ would be confusing. They're all static and their purpose in the scripting system is in the filename. I.e anything named Cmd_X in a file called battle_scripting_commands.c is self-exploratory.

PikalaxALT:
@GriffinRichards please propose another alternative naming scheme or close this PR.

GriffinRichards:
Should've updated with a comment, the conversation continued on Discord. The consensus was more or less the current macro names are too vague without the atk## qualifiers, and no other reason for the atk## naming scheme was proposed, so current state for this PR is:
1. Leave the cmd names as is (Cmd_, i.e. short and no atk## qualifier)
2. Rename the macro constants to be clearer/more specific
I can probably get to this later this week if not next

GriffinRichards:
Update: Still not terribly happy with the flag names but it's better than the meaningless ones some of them had.
Old Name | New Name
------------ | -------------
ATK80_DMG_ | DMG_
ATK80_DMG_HALF_BY_TWO_NOT_MORE_THAN_HALF_MAX_HP | DMG_RECOIL_FROM_MISS
ATK4F_DONT_CHECK_STATUSES | SWITCH_IGNORE_ESCAPE_PREVENTION
ATK48_ | STAT_CHANGE
ATK48_ONLY_MULTIPLE | STAT_CHANGE_SKIP_NEXT_ANIM
ATK48_DONT_CHECK_LOWER | STAT_CHANGE_IGNORE_ABILITY
ATK49_ | MOVEEND_
ATK49_MOVE_END_ABILITIES | ATK49_CONTACT_ABILITIES

STAT_CHANGE might be a problem given that it implies a relation to the other STAT_CHANGE, and STAT_ANIM is already in use. STAT_CHANGE_ANIM might be too long. Another option is something like STATANIMFLAG.

GriffinRichards:
The existing `STAT_CHANGE` prefix used by `statbuffchange` was replaced with `STAT_BUFF`, and `STAT_CHANGE` is now used exclusively by `statchangeanimation`. This PR is now just awaiting re-review

GriffinRichards:
This is good to merge unless there's further review. All feedback given in Discord was addressed in `73b96fe` and `5fd44d1`

huderlem:
@DizzyEggg Can you do a final sign-off on this PR, please?

PR #796, opened 2019-09-04
Documented Mechadoll answers.

AsparagusEduardo:
Is the name format ok? ^^;


data/event_scripts.s
```diff
@@ -3595,139 +3595,139 @@ gContestPaintingTough3::
 	.string "The mighty tough\n"
 	.string "hyper POKéMON {STR_VAR_1}$"
 
-gUnknown_0827ECBC:: @ 27ECBC
+gTrickHouse_Mechadoll1_Q1A1:: @ 27ECBC
```
PikalaxALT: Perhaps choose names that describe the contents rather than (or in addition to) the use context?

AsparagusEduardo: Is there a maximum of characters that I shoud strive for?

PR #804, opened 2019-09-11
Fix gbafix extraneous parentheses warning

GriffinRichards:


PikalaxALT:
A note on this warning: It only crops up when using xcode's C compiler (default on Mac OS X).

PR #807, opened 2019-09-14
Document more of field_specials

GriffinRichards:


src/field_specials.c
```diff
@@ -63,26 +69,25 @@ EWRAM_DATA u8 gBikeCollisions = 0;
...
-static EWRAM_DATA u8 gUnknown_0203AB5E = 0;
-static EWRAM_DATA u16 gUnknown_0203AB60 = 0;
-static EWRAM_DATA u16 gUnknown_0203AB62 = 0;
-static EWRAM_DATA struct ListMenuItem *gUnknown_0203AB64 = NULL;
-static EWRAM_DATA u16 gUnknown_0203AB68 = 0;
-static EWRAM_DATA u16 gUnknown_0203AB6A = 0;
-static EWRAM_DATA u8 gUnknown_0203AB6C = 0;
-static EWRAM_DATA u8 gUnknown_0203AB6D = 0;
-static EWRAM_DATA u8 gUnknown_0203AB6E = 0;
+static EWRAM_DATA u8 gTutorMoveAndElevatorWindowId = 0;
```
huderlem: `static` symbols should use the `s` prefix

src/field_specials.c
```diff
@@ -1077,36 +1088,35 @@ static void PCTurnOnEffect_0(struct Task *task)
     task->data[3]++;
 }
 
-// enum pc location, 
-static void PCTurnOnEffect_1(s16 flag, s8 dx, s8 dy)
+static void PCTurnOnEffect_1(s16 pcOff, s8 dx, s8 dy)
 {
     u16 tileId = 0;
-    if (flag != 0)
+    if (pcOff != 0)
```
huderlem: `if (isPCTurnedOff)`

src/field_specials.c
```diff
@@ -577,10 +587,10 @@ void SpawnLinkPartnerEventObject(void)
                         linkSpriteId = EVENT_OBJ_GFX_RIVAL_MAY_NORMAL;
                     break;
             }
-            SpawnSpecialEventObjectParameterized(linkSpriteId, gUnknown_085B2B5C[j], 0xf0 - i, gUnknown_085B2B60[j][0] + x + 7, gUnknown_085B2B60[j][1] + y + 7, 0);
+            SpawnSpecialEventObjectParameterized(linkSpriteId, movementTypes[j], 0xf0 - i, coordOffsets[j][0] + x + 7, coordOffsets[j][1] + y + 7, 0);
```
huderlem: Should this `0xf0` be `240` instead?

src/field_specials.c
```diff
@@ -1327,16 +1337,16 @@ void BufferEReaderTrainerName(void)
 
 u16 GetSlotMachineId(void)
 {
-    static const u8 gUnknown_085B2B88[] = {12, 2, 4, 5, 1, 8, 7, 11, 3, 10, 9, 6};
-    static const u8 gUnknown_085B2B94[] = {0, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 5};
-    static const u8 gUnknown_085B2BA0[] = {3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 5, 5};
+    static const u8 sSlotMachineRandomSeed[] = {12, 2, 4, 5, 1, 8, 7, 11, 3, 10, 9, 6};
```
huderlem: pluralize

src/field_specials.c
```diff
@@ -2018,130 +2058,191 @@ bool32 sub_8139ED0(void)
...
-        { 0x15, 0x23 },
-        { 0xff, 0xff },
-        { 0xff, 0xff },
-        { 0x02, 0x04 },
-        { 0x07, 0x15 },
-        { 0x07, 0x15 },
-        { 0x0e, 0x1c },
-        { 0x0d, 0x70 },
-        { 0x07, 0x38 }
+    static const u8 sFrontierManiacStreakThresholds[][FRONTIER_MANIAC_MESSAGE_COUNT - 1] =
```
huderlem: `FRONTIER_MANIAC_MESSAGE_COUNT - 1` doesn't make sense here.  This data is independent from the `sFrontierManiacMessages` data.

src/field_specials.c
```diff
@@ -2783,34 +2822,34 @@ void sub_813A76C(void)
...
-        BattleFrontier_Lounge5_Text_264D01,
-        BattleFrontier_Lounge5_Text_264D6B,
-        BattleFrontier_Lounge5_Text_264DD7,
-        BattleFrontier_Lounge5_Text_264E33,
-        BattleFrontier_Lounge5_Text_264E8F,
-        BattleFrontier_Lounge5_Text_2648BE,
+void ShowNatureGirlMessage(void)
+{
+    static const u8 *const sNatureGirlMessages[] = {
+        [NATURE_HARDY]   = BattleFrontier_Lounge5_Text_26468D,
```
huderlem: Let's label these texts while we're here.

src/field_specials.c
```diff
@@ -2870,45 +2911,59 @@ void sub_813A854(void)
...
     u16 frontierFacilityId = VarGet(VAR_FRONTIER_FACILITY);
 
-    if (VarGet(VAR_FRONTIER_GAMBLER_PLACED_BET_F) == 1)
+    if (VarGet(VAR_FRONTIER_GAMBLER_STATE) == FRONTIER_GAMBLER_PLACED_BET)
     {
-        if (gUnknown_085B3104[var2] == (frontierFacilityId << 8) + battleMode)
+        if (sFrontierChallenges[challenge] ==  FRONTIER_CHALLENGE(frontierFacilityId, battleMode))
         {
-            if (a0 != 0)
+            if (won != FALSE)
```
huderlem: `if (won)`

src/field_specials.c
```diff
@@ -3216,52 +3254,53 @@ void sub_813ADD4(void)
...
+    moveIndex = VarGet(VAR_TEMP_D);
 
-    if (temp1 != 0)
+    if (moveTutor != 0)
     {
         i = 0;
         do
         {
-            if (gTutorMoves[i] == gUnknown_085B320C[temp2])
+            if (gTutorMoves[i] == sBattleFrontier_TutorMoves2[moveIndex])
```
huderlem: `gTutorMoves` should probably use the `TUTOR_MOVE_COUNT` constant in its definition.

src/field_specials.c
GriffinRichards: Resolved

src/field_specials.c
GriffinRichards: The messages in `sFrontierManiacMessages` are printed depending on whether or not the streaks in `sFrontierManiacStreakThresholds` have been achieved. So if the player has a streak `> sFrontierManiacStreakThresholds[facility][1]`, then `sFrontierManiacMessages[facility][2]` is shown, if they have a streak `> sFrontierManiacStreakThresholds[facility][0]` but `<= sFrontierManiacStreakThresholds[facility][1]` then `sFrontierManiacMessages[facility][1]` is shown, and if they have a streak `<= sFrontierManiacStreakThresholds[facility][0]` then `sFrontierManiacMessages[facility][0]` is shown. This is demonstrated by the last two lines

```for (i = 0; i < FRONTIER_MANIAC_MESSAGE_COUNT - 1 && sFrontierManiacStreakThresholds[facility][i] < winStreak; i++);

ShowFieldMessage(sFrontierManiacMessages[facility][i]);```

Maybe the relationship would be clearer if something like `#define FRONTIER_MANIAC_STREAK_COUNT   FRONTIER_MANIAC_MESSAGE_COUNT - 1` was added.

src/field_specials.c
GriffinRichards: Yes, `0xf0 - i` is the event object localId, which makes more sense in decimal.

src/field_specials.c
GriffinRichards: The first `if` block is to turn the pc off, which occurs if the pc is already on. So changed to `if (isPcTurnedOn)`

GriffinRichards:
All other changes without a follow-up comment were resolved. Documented the following additional functions while making review changes:
`SetPCBoxToSendMon`, `GetPCBoxToSendMon`, `ShouldShowBoxWasFullMessage`, and `IsDestinationBoxFull`.

PR #812, opened 2019-09-19
Document script_menu and LilycoveCity_Harbor

GriffinRichards:
Documents `script_menu` functions and data, including using IDs for the `multichoice` cmd. Also incidentally documented `LilycoveCity_Harbor` and some more of the scrollable multichoice in `field_specials`


GriffinRichards:
Also a note that merging would be slightly easier if this is merged before #778 as opposed to the other way around

huderlem:
merge conflict

huderlem:
This is good or merge once the merge conflict is fixed.

PR #813, opened 2019-09-21
use $(PRET) tools install

nicholatian:
Also use `clang` on macOS, because `gcc` can be really wonky with homebrew sometimes.

The main feature is the migration of tools out of the local directory they’re installed in and into a folder `/opt/pret/`. This makes it easy to move the tools around or use different versions for diagnostics. It also allows people to download precompiled binaries of the tools, which I have made available on the Cave of Origin for Linux and macOS using the [`grups` tool](https://code.caveoforig.in/fhtk/grups).

I’m open to seeing the tools moved to the pret GitHub organisation, and this is how `agbcc` is already handled, but for now they exist under the Fanhacker’s Toolkit because it’s an org I have control over. I have tested this with a clean repo using `make -j$(nproc) && make compare` and get an OK, using the tools downloaded from the Cave using the `install-all.sh` convenience script provided by grups. Let me know how it handles for you.


PikalaxALT:
Please update the include path for compilation with agbcc, to use the pret headers instead of the devkitPro headers.

Makefile
```diff
@@ -3,12 +3,26 @@ COMPARE ?= 0
 
 ifeq ($(CC),)
 HOSTCC := gcc
+# use system clang on macOS; GCC is wonky w/ homebrew sometimes
+ifneq ($(OS),Windows_NT)
+UNAME := $(shell uname -s)
+ifeq ($(UNAME),Darwin)
+HOSTCC := /usr/bin/clang
```
PikalaxALT: No. Just no.

Makefile
```diff
@@ -75,49 +102,33 @@ endif
 
 CPPFLAGS := -iquote include -Wno-trigraphs -DMODERN=$(MODERN)
 ifeq ($(MODERN),0)
-CPPFLAGS += -I tools/agbcc/include -I tools/agbcc
+CPPFLAGS += -I $(PRET)/include -I $(PRET)
```
PikalaxALT: Add `-nostdinc -undef` here

install-tools.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+# -*- coding: utf-8 -*-
+## Convenience script for installing all the tools at once
+
+REPO='https://pkg.caveoforig.in/fhtk/';
```
PikalaxALT: Would be better to have the repo point back to pret on github, or to some host maintained by pret keyholders.

util/install.sh
```diff
@@ -0,0 +1,26 @@
...
+	exit 2;
+fi
+
+if [ "$(id -u)" != '0' ]; then
+	_sudo='sudo ';
+else
+	_sudo='';
+fi
+
+_cwd="${PWD}";
```
PikalaxALT: use `pushd` and `popd` instead

nicholatian: thanks i forgot about those

Makefile
nicholatian: uh, yes. the `gcc` command is ambiguous asf on macOS. we should stick to the system compiler unless otherwise specified

PR #817, opened 2019-09-25
Document day_care.inc and related

GriffinRichards:


include/constants/daycare.h
```diff
@@ -0,0 +1,32 @@
+#ifndef GUARD_DAYCARE_CONSTANTS_H
+#define GUARD_DAYCARE_CONSTANTS_H
+
+// Egg parents
+#define EGG_MOTHER 0
```
huderlem: `EGG_MOTHER` and `EGG_FATHER` seem to be misnomers.  They are really just slots and have nothing to do with which parent it is.

GriffinRichards: The mother is inherently slot 0, see something like `DetermineEggSpeciesAndParentSlots` where only the the first daycare mon is used to determine species. It is sort of misleading though, because the indices are referred to regardless (like if there isn't an egg, or there's only one mon in daycare). Should they just be dropped altogether?

huderlem: The constants you've defined are only relevant for a single line of code, and they don't make sense outside of it, so they probably should be removed altogether.
```
BuildEggMoveset(&egg, &daycare->mons[parentSlots[1]].mon, &daycare->mons[parentSlots[0]].mon);
```

src/daycare.c
```diff
@@ -532,9 +536,9 @@ static void RemoveIVIndexFromList(u8 *ivs, u8 selectedIv)
 static void InheritIVs(struct Pokemon *egg, struct DayCare *daycare)
 {
     u8 i;
-    u8 selectedIvs[3];
+    u8 selectedIvs[INHERITED_IV_COUNT];
     u8 availableIVs[NUM_STATS];
-    u8 whichParent[ARRAY_COUNT(selectedIvs)];
+    u8 whichParent[INHERITED_IV_COUNT];
```
huderlem: This local array could use a better name. Maybe `whichParentIvs`

include/constants/daycare.h
huderlem: I guess they're also relevant in this code:
```
u8 ditto = parentSlots[EGG_FATHER];
parentSlots[EGG_FATHER] = parentSlots[EGG_MOTHER];
parentSlots[EGG_MOTHER] = ditto;
```

However, the constants are pretty misleading in general, especially in the scripts.

src/daycare.c
```diff
@@ -889,7 +888,7 @@ static bool8 _DoEggActions_CheckHatch(struct DayCare *daycare)
     }
 
     // try to trigger poke sex
```
huderlem: Let's update this comment while you're here

include/constants/daycare.h
GriffinRichards: Yeah outside of `DetermineEggSpeciesAndParentSlots` they don't make a ton of sense. They sort of do in `GetDaycareCompatibilityScore` but the parents are interchangeable there so it doesn't matter.

src/egg_hatch.c
```diff
@@ -388,13 +388,13 @@ void ScriptHatchMon(void)
     AddHatchedMonToParty(gSpecialVar_0x8004);
 }
 
-static bool8 sub_807158C(struct DayCare *daycare, u8 daycareId)
+static bool8 _CheckDaycareMonReceivedMail(struct DayCare *daycare, u8 daycareId)
 {
-    u8 nick[0x20];
+    u8 nick[32];
```
huderlem: `nickname`

src/egg_hatch.c
```diff
@@ -388,13 +388,13 @@ void ScriptHatchMon(void)
...
     struct DaycareMon *daycareMon = &daycare->mons[daycareId];
 
     GetBoxMonNick(&daycareMon->mon, nick);
```
huderlem: This should really be expanded to `GetBoxMonNickname`

GriffinRichards: Expanded the uses of nick in `egg_hatch.c` and `daycare.c`. `GetMonNickname` was taken by an essentially identical function in `party_menu.c`, so the less-used one was given `GetMonNickname2`.

GriffinRichards:
All current comments addressed

PR #818, opened 2019-09-25
Relabeled unknown functions in `pokemon_summary_screen.c`

arantonitis:
Just bits of documentation for some previously unknown functions.

### DIFF
--- a/src/pokemon_summary_screen.c
+++ b/src/pokemon_summary_screen.c
@@ -224,10 +224,10 @@ static void DrawExperienceProgressBar(struct Pokemon* mon);
 static void DrawContestMoveHearts(u16 move);
 static void LimitEggSummaryPageDisplay(void);
 static void ResetWindows(void);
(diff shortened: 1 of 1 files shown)


src/pokemon_summary_screen.c
```diff
@@ -224,10 +224,10 @@ static void DrawExperienceProgressBar(struct Pokemon* mon);
 static void DrawContestMoveHearts(u16 move);
 static void LimitEggSummaryPageDisplay(void);
 static void ResetWindows(void);
-static void sub_81C25E8(void);
-static void sub_81C2628(void);
-static void sub_81C2794(void);
-static void sub_81C27DC(struct Pokemon *mon, u16 a);
+static void Summary_PrintMonInfo(void);
```
huderlem: Let's remove the `Summary_` prefix from these function names.  It's verbose and doesn't serve much purpose, especially since these are static functions.

PR #820, opened 2019-09-29
Document fldeff_flash.c.

arantonitis:
I figured out the purpose of FlashStruct--to determine when cave or underwater palette transitions occur, based on map type.


PikalaxALT:
Duplicate of #803

PR #824, opened 2019-09-30
Fishing Groups

garakmon:


src/data/wild_encounters.json
```diff
@@ -8,25 +8,33 @@
           "type": "land_mons",
           "encounter_rates": [
             20, 20, 10, 10, 10, 10, 5, 5, 4, 4, 1, 1
-          ]
+          ],
+          "groups": {}
```
PikalaxALT: These dummy groups could probably be removed, and the manifest edited to query whether the key exists at all.

PR #826, opened 2019-10-04
Use MAX_LEVEL constant for stat calculation

GriffinRichards:


GriffinRichards:
Also weird note about this, while `MAX_LEVEL` is clearly what's intended there (and perhaps GF had some similar constant used here), editing the `MAX_LEVEL` would result in what might be unexpected scaling behavior for some people, i.e. if `MAX_LEVEL` were changed to 200, a level 200 pokemon would have the stats that a level 100 pokemon would have before, or similarly if it were changed to level 10, a level would have the stats of a level 100 pokemon and a level 5 would have the stats of a level 50. So it may be better to leave the 100 hard-coded so that the scaling remains intact if people change `MAX_LEVEL`?

GriffinRichards:
Seems like there's some agreement with the above comment, and that it makes more sense to leave the constant hard-coded to keep scaling intact.

PR #835, opened 2019-10-11
Document cable_club.inc

GriffinRichards:
Will be clarified in a few places while doing `cable_club.c`


data/maps/DewfordTown_PokemonCenter_2F/scripts.inc
```diff
@@ -1,19 +1,20 @@
 DewfordTown_PokemonCenter_2F_MapScripts:: @ 81FC615
-	map_script MAP_SCRIPT_ON_FRAME_TABLE, DewfordTown_PokemonCenter_2F_MapScript2_276C3B
-	map_script MAP_SCRIPT_ON_WARP_INTO_MAP_TABLE, DewfordTown_PokemonCenter_2F_MapScript2_276B6C
-	map_script MAP_SCRIPT_ON_LOAD, DewfordTown_PokemonCenter_2F_MapScript1_276BBE
-	map_script MAP_SCRIPT_ON_TRANSITION, PokemonCenter_2F_OnTransition
+	map_script MAP_SCRIPT_ON_FRAME_TABLE, CableClub_MapScript2_OnFrame
```
huderlem: `_MapScript2` `_MapScript1`, and `_MapScript` are redundant information, which should be removed from map script labels.  For example, `CableClub_OnFrameTable` or `CableClub_OnTransition`.  (This is how the rest of the codebase is doing it, too)

include/constants/vars.h
```diff
@@ -77,7 +77,7 @@
 #define VAR_REGICE_STEPS_2                   0x403C
 #define VAR_REGICE_STEPS_3                   0x403D
 #define VAR_ALTERING_CAVE_WILD_SET           0x403E
-#define VAR_ALWAYS_ZERO_0x403F               0x403F // This var is read and written, but is always zero.
+#define VAR_DISTRIBUTE_EON_TICKET            0x403F // This var is read and written, but is always zero. Presumably meant to be set externally by Mystery Gift distribution
```
huderlem: I believe the EON TICKET is only obtainable in emerald via record mixing with a R/S game.

include/dodrio_berry_picking.h
```diff
@@ -2,7 +2,7 @@
 #define GUARD_DODRIO_BERRY_PICKING_H
 
 void sub_802493C(u16 a0, void (*callback)(void));
-void sub_8027A5C(void);
-void sub_8027AAC(void);
+void DoesPartyHaveDodrio(void);
```
huderlem: `IsDodrioInParty()` seems like a better name

include/pokemon_jump.h
```diff
@@ -4,10 +4,10 @@
 #include "main.h"
 
 void sub_802A9A8(u16 monId, MainCallback callback);
-bool32 sub_802C908(u16 species);
-void sub_802C920(void);
+bool32 IsSpeciesAllowedInPokemonJump(u16 species);
+void DoesPartyHavePokemonJumpSpecies(void);
```
huderlem: Same pattern here.  `IsPokemonJumpSpeciesInParty()`

include/link.h
```diff
@@ -48,6 +48,28 @@
 #define EXTRACT_LINK_ERRORS(status) \
 (((status) & LINK_STAT_ERRORS) >> LINK_STAT_ERRORS_SHIFT)
 
+#define LINKTYPE_0x1111              0x1111  // trade
```
huderlem: What's the story behind all the unnamed `LINKTYPE_` constants here?  Was it because you weren't able to figure them out?

include/constants/vars.h
GriffinRichards: Updated comment to remove the speculation/add context. The meaning of the var is still the same, based on its usage in the source

include/link.h
GriffinRichards: Just documented the ones I did because their usage was obvious in the linkup functions of `cable_club.c`, and the berry blender type is the only link type used by `berry_blender`. I'll get to the rest, but probably outside this PR. I'd rather document the functions they're used in to be sure about their names (and most of those are outside the scope of `cable_club.inc`), so I just commented what they seem to be based on the numbering pattern and where they're used.

GriffinRichards:
All current comments addressed

PR #843, opened 2019-10-17
MIDI Parity with Fire Red

Kermalis:


huderlem:
Can you elaborate on what these changes actually are?  They appear to be no-ops, since the files already existed in matching form before this PR.

PR #850, opened 2019-10-21
Split some smaller files from event_scripts

GriffinRichards:
See comments in `event_scripts.s` about some other potential splits. Probably about 10-15 small files left in there


data/text/mart_attendant.inc
```diff
@@ -0,0 +1,10 @@
+gText_HowMayIServeYou:: @ 8272A21
```
huderlem: I'd rather this file be named `mart_clerk.inc`

PR #851, opened 2019-10-21
battle_anim_effects_3

Sewef:


Sewef:
OK for review

data/battle_anim_scripts.s
```diff
@@ -8014,7 +8014,7 @@ HyperFangInContest:
 
 Move_TRI_ATTACK:
 	loadspritegfx ANIM_TAG_TRI_FORCE_TRIANGLE
-	createsprite gUnknown_085CE2F8, ANIM_TARGET, 2, 16, 0
+	createsprite gTriForceTriangleSpriteTemplate, ANIM_TARGET, 2, 16, 0
```
huderlem: I think `TriAttack` would be more appropriate than `TriForce`

src/battle_anim_mons.c
```diff
@@ -966,7 +966,7 @@ void sub_80A6D60(struct BattleAnimBgData *unk, const void *src, u32 arg2)
     CopyBgTilemapBufferToVram(unk->bgId);
 }
 
-u8 sub_80A6D94(void)
+u8 GetPaletteOffset(void)
```
huderlem: `GetPaletteOffset()` needs a better name.  What is the offset for?  The battle background (etc.?)?

Sewef: This functions returns 1 for contest, else 2. According to `LoadCompressedPalette` it's a palette offset...
Welp, when I see the use in `LoadMoveBg`, I think `GetPaletteOffset` is kind of pointless anyway, I don' tknow how to label that

huderlem: I investigated this.  Its the palette number for the battle background.  So something like `GetBattleBgPaletteNum()` is fine.

include/constants/battle_anim.h
```diff
@@ -237,7 +237,7 @@
 #define ANIM_TAG_SPOTLIGHT                  (ANIM_SPRITES_START + 227)
 #define ANIM_TAG_LETTER_Z                   (ANIM_SPRITES_START + 228)
 #define ANIM_TAG_RAPID_SPIN                 (ANIM_SPRITES_START + 229)
-#define ANIM_TAG_TRI_FORCE_TRIANGLE         (ANIM_SPRITES_START + 230)
+#define ANIM_TAG_TRI_ATTACK_TRIANGLE         (ANIM_SPRITES_START + 230)
```
huderlem: fix this mis-alignment too

Sewef:
OK.

PR #863, opened 2019-10-31
Document party menu

GriffinRichards:
Incidentally documents some scripts as well (and renames one, Cozmos House), as well as some adjacent functions. Essentially documents fldeff_softboiled as well.


include/party_menu.h
```diff
@@ -4,93 +4,91 @@
...
 {
     MainCallback exitCallback;
-    TaskFunc unk4;
-    u8 unk8_0:4;
-    u8 mode:2;
-    u8 unk8_2:2;
+    TaskFunc task;
+    u8 menuType:4;
+    u8 layout:2;
+    u8 unused:2;
```
huderlem: No need to have this `unused` field defined.

src/battle_main.c
```diff
@@ -166,8 +167,8 @@ EWRAM_DATA static u32 sUnusedUnknownArray[25] = {0};
 EWRAM_DATA u32 gBattleTypeFlags = 0;
 EWRAM_DATA u8 gBattleTerrain = 0;
 EWRAM_DATA u32 gUnknown_02022FF4 = 0;
-EWRAM_DATA struct UnknownPokemonStruct4 gUnknown_02022FF8[3] = {0}; // what is it used for?
-EWRAM_DATA struct UnknownPokemonStruct4* gUnknown_02023058 = NULL; // what is it used for?
+EWRAM_DATA struct UnknownPokemonStruct4 gMultiPartnerParty[PARTY_SIZE / 2] = {0};
```
huderlem: I'd rather see a define for `MULTI_PARTY_SIZE = PARTY_SIZE / 2`, rather than see `PARTY / 2` everywhere in the code.

GriffinRichards: Sounds good. It gets sort of messy with the size of Battler Frontier parties, which are also 3 despite not being multi parties, because they both use the same "choose 3 mons from the party menu" logic. I defined both constants (`MULTI_PARTY_SIZE` and `FRONTIER_PARTY_SIZE`) for now, but if someone wanted to change the size of you'd have to handle them differently.

GriffinRichards: Also I added one instance of `PARTY_SIZE / 2` for `gBattlePartyCurrentOrder` because it literally is the party packed into 4 bits at a time

PR #866, opened 2019-11-01
Document Team Aqua/Magma scripts

GriffinRichards:
Documents the main areas Team Aqua/Magma are encountered and some adjacent things along the way. Couldn't get `io_reg.s` to process `#include` for the `gba/io_reg.h` constants


data/maps/AquaHideout_B1F/map.json
```diff
@@ -128,7 +128,7 @@
       "movement_range_y": 1,
       "trainer_type": "1",
       "trainer_sight_or_berry_tree_id": "4",
-      "script": "AquaHideout_B1F_EventScript_233AEC",
+      "script": "AquaHideout_B1F_EventScript_Grunt25",
```
huderlem: There aren't 25 grunts in this map. Why is the number so high?  This seems to be a recurring issue throughout this whole PR.  Grunt "numbers" should be local to the current map, not globally indexed for the whole game.

PR #876, opened 2019-11-05
Clean and split mystery event msg

GriffinRichards:
To keep the diff intact all changes were made in the first commit. The second commit is just the file split.


data/mystery_event_msg.s
```diff
@@ -13,332 +15,11 @@
...
-
-sText_MysteryGiftOldSeaMap_1:
-	.string "It appears to be for use at the\nLILYCOVE CITY port.\pWhy not give it a try and see what\nit is about?$"
-
-sText_MysteryGiftOldSeaMap_2:
-	.string "Thank you for using the MYSTERY\nGIFT System.$"
-
-sText_MysteryGiftOldSeaMap_3:
-	.string "Oh, I'm sorry, {PLAYER}.\nYour BAG's KEY ITEMS POCKET is full.\pPlease store something on your PC,\nthen come back for this.$"
+	.include "data/scripts/meventmsg_stamp_card.inc"
```
huderlem: Why is `msg` in the name of all of these?  These are full mystery event scripts, so I don't think `msg` is appropriate in the names.

GriffinRichards: Only for naming consistency with `mystery_event_msg.s`, which only contains those split files. I can rename that as well

PR #893, opened 2019-11-10
Document contest painting image processing effects

huderlem:
`contest_painting_effects.c` might have been an external GBA image effects library that Game Freak imported because there are plenty of unused effects, and missing enum values for image effects.


DizzyEggg:
This indeed looks like an external library. In this case, wouldn't it be better to rename variables that have contest in their names to more general names? like ContestPaintingContext  would become ImageContext?

huderlem:
> This indeed looks like an external library. In this case, wouldn't it be better to rename variables that have contest in their names to more general names? like ContestPaintingContext would become ImageContext?

Yes, I'll rename that before this is merged.

PR #898, opened 2019-11-12
Match ConvertImageProcessingToGBA

DizzyEggg:
That's one of the functions that could really use a comment or two.

### DIFF
--- a/src/image_processing_effects.c
+++ b/src/image_processing_effects.c
@@ -761,211 +761,53 @@ static u16 QuantizePixel_BlurHard(u16 *prevPixel, u16 *curPixel, u16 *nextPixel)
     return RGB2(red, green, blue);
 }
 
-/*
-void ConvertImageProcessingToGBA(struct ImageProcessingContext *arg0)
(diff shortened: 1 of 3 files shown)


src/image_processing_effects.c
```diff
@@ -761,211 +761,53 @@ static u16 QuantizePixel_BlurHard(u16 *prevPixel, u16 *curPixel, u16 *nextPixel)
...
+    u16 *src, *dest, *src_, *dest_;
+    u16 width, height;
 
-    if (arg0->var_16 == 2)
+    width = context->canvasWidth >> 3;
+    height = context->canvasHeight >> 3;
+    src_ = context->canvasPixels;
+    dest_ = context->dest;
+
+    if (context->var_16 == 2)
```
PikalaxALT: What other values can var_16 take?

DizzyEggg: var_16 is set up only once in the entire game in contest_painting.c to value 2. So no clue what it stands for.

PikalaxALT: Seems to be an indicator for whether the source image is unpacked 8bpp or unpacked 4bpp 🤔

PR #901, opened 2019-11-12
Document Battle Frontier lounges

GriffinRichards:


data/maps/BattleFrontier_Lounge3/scripts.inc
```diff
@@ -1,214 +1,218 @@
...
-	setvar VAR_0x8004, 20
+BattleFrontier_Lounge3_EventScript_RewardBet10:: @ 8261F8A
+	buffernumberstring 0, (BET_AMOUNT_10 * 2)
+	setvar VAR_0x8004, (BET_AMOUNT_10 * 2)
 	return
 
-BattleFrontier_Lounge3_EventScript_261F94:: @ 8261F94
-	buffernumberstring 0, 30
-	setvar VAR_0x8004, 30
+BattleFrontier_Lounge3_EventScript_RewardBet20:: @ 8261F94
```
huderlem: Should be `RewardBet15` instead of 20, right?

GriffinRichards: Whoops, yes. They're 5, 10, 15

PR #903, opened 2019-11-13
Document Trainer Hill

GriffinRichards:


include/constants/trainer_hill.h
```diff
@@ -1,27 +1,46 @@
...
 #define TRAINER_HILL_FUNC_SET_TAG 17
 
-// Values returned by the TrainerHillGetChallengeStatus
-// function.
+#define TRAINER_HILL_TEXT_INTRO        2
+#define TRAINER_HILL_TEXT_PLAYER_LOST  3
+#define TRAINER_HILL_TEXT_PLAYER_WON   4
+#define TRAINER_HILL_TEXT_AFTER        5
+
+#define NUM_TRAINER_HILL_TRAINERS     (NUM_TRAINER_HILL_FLOORS * 2)
```
huderlem: Is this relationship definitely supposed to be `num_floors * 2`, or does it just happen to be that for the existing data in the game?

include/global.h
```diff
@@ -391,11 +391,11 @@ struct BattleFrontier
     /*0xEBA*/ u16 field_EBA;
     /*0xEBC*/ u32 battlesCount;
     /*0xEC0*/ u16 field_EC0[16];
-    /*0xEE0*/ u8 field_EE0;
+    /*0xEE0*/ u8 trainerFlags;
     /*0xEE1*/ u8 opponentName[2][PLAYER_NAME_LENGTH + 1];
     /*0xEF1*/ u8 field_EF1[2][4];
-    /*0xEF9*/ u8 field_EF9_0:7;
-    /*0xEF9*/ u8 field_EF9_1:1;
+    /*0xEF9*/ u8 neverRead:7;
```
huderlem: I'd prefer `unk*` rather than `neverRead`, in general.

src/data/battle_frontier/trainer_hill.h
```diff
@@ -382,18 +383,18 @@ static const struct TrHillFloor gUnknown_0862609C_floors[] = {
...
 };
 
-static const struct TrHillFloor sDataTagNormal_floors[] =
+static const struct TrHillFloor sDataTagNormal_Floors[] =
 {
     [0] =
     {
-        .unk0 = 0x11,
-        .unk1 = 0x12,
+        .trainerNum1 = 17,
```
huderlem: Are these numbers supposed to be opponent constants, etc.?

include/constants/trainer_hill.h
GriffinRichards: Yeah there's always 2 trainers per floor. Even when E-Reader trainers are used, if there's an odd number of trainers then one is added to round it up.

include/global.h
GriffinRichards: Alright, I'll update that. I don't have a strong preference one way or the other, though I prefer `neverRead` slightly because it contains at least a little information.

src/data/battle_frontier/trainer_hill.h
GriffinRichards: I don't think so. The actual `TRAINER_` value for the trainer on each floor is just 1 and 2 each time (see the use of `gSaveBlock2Ptr->frontier.trainerIds` in `trainer_hill.c`. As far as I can tell those are some kind of identifier for each trainer used to determine what prize to give. If you mean the use of the name `trainerNum` is misleading here then I'm happy to rename it.

huderlem: it's fine for now

GriffinRichards: Yeah there's clearly some funky things going on with trainer hill. I think documenting more of E-Reader might clear it up.

PR #908, opened 2019-11-20
Turn some magic numbers into named constants

ketsuban:


GriffinRichards:
Looks like you forgot to include `constants/field_poison.h` in `field_control_avatar.c`. I don't know why on my screen at least it just says the check is pending, details show it failed.

PR #909, opened 2019-11-21
Document Apprentice

GriffinRichards:
Not overly a fan of the macro-heavy script. You'd probably need 4 files open at once to understand internally what's going on in `apprentice.inc`. It does have the benefit of not being crowded by function id constants and `setvar`s though. Open to thoughts on that, because if it seems good that's how I'll do the rest of Battle Frontier. Also moved data in a separate final commit so it can be diffed neatly.


asm/macros/battle_frontier.inc
```diff
@@ -32,4 +32,141 @@ special CallBattleTowerFunc
...
+setvar VAR_0x8006, \tobuff
+special CallApprenticeFunction
+.endm
+
+.macro apprentice_buffv  whichstringvar:req tobuff:req
+setvar VAR_0x8004, APPRENTICE_FUNC_BUFFER_STRING
+setvar VAR_0x8005, \whichstringvar
+copyvar VAR_0x8006, \tobuff
+special CallApprenticeFunction
+.endm
```
GriffinRichards: These could be consolidated into one macro by checking if `tobuff >= VARS_START`

src/record_mixing.c
```diff
@@ -1018,57 +1018,59 @@ static void Task_DoRecordMixing(u8 taskId)
...
-    var_2C = 0;
-    r8 = 0;
+    oldPlayerApprenticeSaveId = 0;
+    numOldPlayerApprentices = 0;
+    apprenticeSaveId = 0;
+    numMixApprentices = 0;
     for (i = 0; i < 2; i++)
     {
-        id = ((i + gSaveBlock2Ptr->playerApprentice.field_B2_1) % 3) + 1;
+        id = ((i + gSaveBlock2Ptr->playerApprentice.saveId) % 3) + 1;
```
huderlem: I think this `% 3` is supposed to be `APPRENTICE_COUNT - 1`

GriffinRichards: Yeah, `saveId` is used as an index into `gSaveBlock2Ptr->apprentices`. Updated

PR #910, opened 2019-11-21
Standardize Object Event naming

GriffinRichards:
NOTE: Should break porymap compatibility

SUMMARY:
This PR standardizes the nomenclature for object events to fit as part of the larger event group (e.g. warp events, coord events, etc). 

For the purposes of file organization, files/filepaths (and their include guards) retain the old `event_object` name scheme. The following files changed names:
`event_obj_lock` --> `event_object_lock`
`bg_event_constants` --> `event_bg`
`event_object_movement_constants` --> `event_object_movement`
`field_object_palette_*.pal` --> `event_object_palette_*.pal`

The following directories changed names:
`graphics/event_objects` --> `graphics/object_events`
`src/data/field_event_obj` --> `src/data/object_event`

The following replacements are made throughout the repo (note that if this PR is being reproduced with a script, file paths/include guards will need to be excluded/fixed to keep the names listed above)

| Old name | New name |
| ------------- | ------------- |
| EventObject | ObjectEvent |
| EventObj | ObjEvent |
| eventObject | objectEvent |
| eventObj | objEvent |
| EVENT_OBJECT | OBJECT_EVENT |
| EVENT_OBJ | OBJ_EVENT |
| EVENTOBJ | OBJEVENT |
| event object | object event |
| event_object | object_event |
| field_event_obj | object_event |
| mapObject | objectEvent |

PR #914, opened 2019-11-25
Document Frontier Util

GriffinRichards:
This started off as Battle Pike documentation but was aborted when it became clear it made more sense to document Frontier Util first, which this does. This will definitely conflict a little with #910.


data/maps/BattleFrontier_BattleArenaBattleRoom/scripts.inc
```diff
@@ -40,16 +39,14 @@ BattleFrontier_BattleArenaBattleRoom_EventScript_2574DC:: @ 82574DC
...
 	goto_if_eq BattleFrontier_BattleArenaBattleRoom_EventScript_25752E
 	applymovement 5, BattleFrontier_BattleArenaBattleRoom_Movement_257C08
 	applymovement 8, BattleFrontier_BattleArenaBattleRoom_Movement_257BEB
 	setvar VAR_TEMP_2, 1
-	setvar VAR_0x8004, FRONTIER_UTIL_FUNC_2
-	setvar VAR_0x8005, 6
+	setvar VAR_0x8004, FRONTIER_UTIL_FUNC_SET_DATA
+	setvar VAR_0x8005, FRONTIER_DATA_6
 	setvar VAR_0x8006, 1
 	special CallFrontierUtilFunc
```
huderlem: Why doesn't this one get a macro?

GriffinRichards: Not sure what `field_CA9_b` is. It's used in determining whether or not a battle should be saved but the battle buffer values are a pain to track down so I left it for later. When I'm actually documenting one of the scripts its in it might become clearer.

include/global.h
```diff
@@ -317,32 +326,27 @@ struct BattleFrontier
...
-    /*0xCB2*/ u16 curChallengeBattleNum; // In case of battle pyramid, the floor.
+    /*0xCA8*/ u8 challengeStatus;
+    /*0xCA9*/ u8 lvlMode:2;
+    /*0xCA9*/ u8 challengePaused:1;
+    /*0xCA9*/ u8 field_CA9_b:1;
+    /*0xCA9*/ u8 unused_CA9_c:1;
+    /*0xCA9*/ u8 unused_CA9_d:1;
+    /*0xCA9*/ u8 unused_CA9_e:1;
+    /*0xCA9*/ u8 unused_CA9_f:1;
+    /*0xCB0*/ u16 selectedPartyMons[MAX_FRONTIER_PARTY_SIZE];
```
huderlem: This field's offset is supposed to be `0xCAA`, not `0xCB0`.

PR #918, opened 2019-12-02
Sync weather names

GriffinRichards:
Syncing with pokefirered names. Also unusual/alternating weather --> abnormal weather


huderlem:
`WEATHER_CLOUDY` is kind of bad, even though it's apparently what the original name was.  It's used almost exclusively for maps that aren't exposed to the sky.

GriffinRichards:
Reverted to `WEATHER_SHADE` for now

huderlem:
Please add a comment to `weather.h` stating the original name of `WEATHER_SHADE` was more like `WEATHER_CLOUDY`

SatoMew:
Since not everyone uses Discord, I'll summarily state my disagreement with renaming `WEATHER_CLOUDY` to anything but the equivalent `WEATHER_OVERCAST` for the following reasons:

- it replaces the intended meaning of the weather effect with our own subjective interpretation based on the immediate graphical output
- it worsens the fragile internal consistency of the constants (e.g.: `WEATHER_SUNNY` is visually indistinguishable from `WEATHER_NONE` and it is not being renamed; `WEATHER_UNDERWATER_BUBBLES` and its unused counterpart `WEATHER_UNDERWATER` are not descriptors of weather conditions either)
- "cloudy" and "overcast" are both weather terms that implicitly contrast with "sunny" with and without some clouds (dim vs. bright)

GriffinRichards:
> Please add a comment to `weather.h` stating the original name of `WEATHER_SHADE` was more like `WEATHER_CLOUDY`

Added

PR #921, opened 2019-12-04
Use ST_OAM constants and SPRITE macros

GriffinRichards:
Seems like when OamData fields are explicitly set to 0 could be standardized as well


src/naming_screen.c
```diff
@@ -1949,37 +1949,169 @@ const struct OamData gOamData_858BFFC =
...
+    },
+    {
+        .x = -20,  
+        .y = -4,  
+        .shape = SPRITE_SHAPE(32x8),  
+        .size = SPRITE_SIZE(32x8),
+        .tileOffset = 5,     
+        .priority = 1
+    },
+    {.x =  12,
```
huderlem: small formatting mistake

PR #922, opened 2019-12-06
Document pokedex info page

Slawter666:


src/pokedex.c
```diff
@@ -2970,13 +2970,13 @@ void LoadInfoScreen(u8 taskId)
             break;
         case 6:
             {
-                u32 r3 = 0;
+                u32 paletteIndex = 0;
```
huderlem: This variable should be named `preservedPalettes`, not `paletteIndex`.  It looks like these are the palettes that don't get faded during the fade effect.

src/pokedex.c
```diff
@@ -4435,7 +4436,7 @@ bool16 HasAllMons(void)
     return TRUE;
 }
 
-void sub_80C09B0(u16 a)
+static void ResetOtherRegisters(u16 a)
```
huderlem: This function needs a more descriptive name.  Something like `ResetVideoRegisters()`.

PR #925, opened 2019-12-08
Document some pokenav match call

GriffinRichards:


huderlem:
1. Consider using a union instead of those defines for the multi-purpose fields.
2. Can some of those pokenav files be given descriptive names, yet, or are they still too hairy?

GriffinRichards:
1. Oh, yeah a union would be a better, I hate those defines.
2. Sort of. Their general purpose is clear to me but the distinction between some of them isnt. e.g. I know both 9 and 10 handle the ribbon section but I dont know how they divide it up.

PR #933, opened 2019-12-15
Document Battle Tent scripts

GriffinRichards:


data/maps/FallarborTown_BattleTentBattleRoom/scripts.inc
```diff
@@ -96,66 +97,69 @@ FallarborTown_BattleTentBattleRoom_EventScript_2009B3:: @ 82009B3
...
 
-FallarborTown_BattleTentBattleRoom_EventScript_200AF3:: @ 8200AF3
+FallarborTown_BattleTentBattleRoom_EventScript_WarpToLobbyWon:: @ 8200AF3
 	delay 60
 	frontier_set FRONTIER_DATA_CHALLENGE_STATUS, CHALLENGE_STATUS_WON
 	special LoadPlayerParty
 	warp MAP_FALLARBOR_TOWN_BATTLE_TENT_LOBBY, 255, 6, 6
 	waitstate
+
+@ Unreachable code block? The flow into the next block also doesnt make sense
```
huderlem: Why doesn't this make sense? It looks like it continues the script after warping into the lobby map.

data/maps/SlateportCity_BattleTentBattleRoom/scripts.inc
```diff
@@ -58,41 +61,40 @@ SlateportCity_BattleTentBattleRoom_EventScript_2099BE:: @ 82099BE
...
-	case 1, SlateportCity_BattleTentBattleRoom_EventScript_209A39
-
-SlateportCity_BattleTentCorridor_EventScript_209A1B:: @ 8209A1B
+	case 1, SlateportCity_BattleTentBattleRoom_EventScript_DefeatedOpponent
+SlateportCity_BattleTent_EventScript_WarpToLobbyLost:: @ 8209A1B
 	frontier_set FRONTIER_DATA_CHALLENGE_STATUS, CHALLENGE_STATUS_LOST
 	special LoadPlayerParty
 	warp MAP_SLATEPORT_CITY_BATTLE_TENT_LOBBY, 255, 6, 6
 	waitstate
+@ forced stop
```
huderlem: Are you sure these `warp` + `waitstate` combos stop the script?  From a quick look at the source code, I couldn't see why that would happen.

data/maps/SlateportCity_BattleTentCorridor/scripts.inc
```diff
@@ -1,121 +1,122 @@
...
-	case NO, SlateportCity_BattleTentCorridor_EventScript_208EB4
-	case YES, SlateportCity_BattleTentCorridor_EventScript_208FF1
-	case MULTI_B_PRESSED, SlateportCity_BattleTentCorridor_EventScript_208EB4
+	case NO, SlateportCity_BattleTentCorridor_EventScript_EnterBattleRoom
+	case YES, SlateportCity_BattleTentCorridor_EventScript_SwapMons
+	case MULTI_B_PRESSED, SlateportCity_BattleTentCorridor_EventScript_EnterBattleRoom
 
-SlateportCity_BattleTentCorridor_EventScript_208FF1:: @ 8208FF1
+SlateportCity_BattleTentCorridor_EventScript_SwapMons:: @ 8208FF1
 	fadescreen 1
```
huderlem: side note, `fadescreen` needs named parameters--not `1` or `0`.

data/maps/SlateportCity_BattleTentBattleRoom/scripts.inc
GriffinRichards: Not sure how, but I've seen enough of them that wouldn't make sense if execution continued. Take for example `VerdanturfTown_BattleTentBattleRoom_EventScript_WarpToLobbyLost`, where if it doesn't stop it continues right into `VerdanturfTown_BattleTentBattleRoom_EventScript_DefeatedOpponent`, which is entirely the opposite of what the previous script was doing.

data/maps/FallarborTown_BattleTentBattleRoom/scripts.inc
GriffinRichards: If it continues into the next block then every time the win streak is incremented there (which would apparently be occurring while the player is warping back the lobby) it would ask if the player is ready for the 2nd opponent. At that point they've already defeated all 3 opponents

GriffinRichards: It looks like it was leftover porting this from Arena. It matches a block there, and I think these are the only direct uses of Arena functions in Fallarbor Tent.

GriffinRichards:
Re: `warp` + `waitstate` stopping script execution, see the definition of `waitstate` in `scrcmd.c` which calls `ScriptContext1_Stop`, and look at something where script execution continues after the `waitstate` like `ShakeCamera`, which when the task is finished calls `EnableBothScriptContexts`. It's also easy enough to test and confirm that it stops execution (which I did just now) by following a `warp waitstate` with `special DoSoftReset` or `setflag <flag>` etc and seeing that they don't execute.

PR #941, opened 2019-12-20
Remove references to LASTRESULT

GriffinRichards:


asm/macros/event.inc
```diff
@@ -818,7 +818,7 @@
 	.byte 0x6d
 	.endm
 
-	@ Displays a YES/NO multichoice box at the specified coordinates, and blocks script execution until the user makes a selection. Their selection is stored in variable 0x800D (LASTRESULT); 0x0000 for "NO" or if the user pressed B, and 0x0001 for "YES".
+	@ Displays a YES/NO multichoice box at the specified coordinates, and blocks script execution until the user makes a selection. Their selection is stored in VAR_RESULT; 0 for "NO" or if the user pressed B, and 1 for "YES".
```
Diegoisawesome: `YES` and `NO` are also constants, like `FEMALE` down below

GriffinRichards: Yeah I figured this was equivalent to what I did there with `MALE (0)`, but I'll write it that way instead

GriffinRichards: Updated

PR #945, opened 2019-12-26
Give rival and Maxie parties better names

Fontbane:
Merry Christmas


src/data/trainer_parties.h
```diff
@@ -6781,15 +6781,15 @@ static const struct TrainerMonNoItemCustomMoves sParty_Wally1[] = {
     }
 };
 
-static const struct TrainerMonNoItemDefaultMoves sParty_Brendan1[] = {
+static const struct TrainerMonNoItemDefaultMoves sParty_BrendanRoute103Mudkip[] = {//Brendan1
```
GriffinRichards: I don't think these comments are necessary. The `TRAINER_*` constants should be updated to match the new party names, these would become the only party names different from their trainer constant.

GriffinRichards:
Seems good otherwise. The `RivalStarter` name convention isn't my favorite because I half expected e.g. `BrendanMudkip` to have Mudkip on its team, but that's how we've been naming them already. 

Also, up to you, but it'd make sense to re-name Tabitha and Shelly too. The Wallys could also be split into `WallyMauville` and `WallyVictoryRoad#`

PR #949, opened 2020-01-08
Document Battle Dome and add facility party constants

GriffinRichards:
There's a few misc things lumped in that were looked at while doing Battle Dome


data/battle_ai_scripts.s
```diff
@@ -2395,10 +2395,10 @@ AI_CV_Hail_ScoreDown1:
 AI_CV_Hail_End:
 	end
 
+@ BUG: Facade score is increased if the target is statused, but should be if the user is. Replace AI_TARGET with AI_USER
```
huderlem: neat

PR #953, opened 2020-01-11
Port some macro/script fixes from pokefirered

GriffinRichards:
NOTE: This PR makes a few potentially intrusive restorations to macro names, listed below.

| Old name | New name |
| ------------- | ------------- |
| `giveitem` | `additem` |
| `giveitem_std` | `giveitem` |
| `takeitem` | `removeitem` |
| `givedecoration` | `adddecoration` |
| `givedecoration_std` | `givedecoration` |
| `takedecoration` | `removedecoration` |
| `givemoney` | `addmoney` | 
|`takemoney` | `removemoney`|
|`givecoins` | `addcoins`|
|`takecoins`|`removecoins`|
|`givepcitem`|`addpcitem`|
|`finditem_std` | `finditem` |

Reasoning: `giveitem` and `giveitem_std` have poor naming distinction, and the former was a name change from the original `additem`. `additem` silently puts an item in the bag, `giveitem` puts an item in the bag and displays a message that the player received it. The same is true of the reasoning for the decoration macro name change. `take` --> `remove` is for english consistency with give/take and add/remove being opposites, and the money/coins macros are renamed to be consistent with the other add/remove macros (which are all silent). The `_std` prefix is then dropped entirely because it's unnecessary.


huderlem:
`goto_if_undefeated` doesn't sit right to me.  To me, the word "undefeated" implies at least one victory.  Would `goto_if_not_defeated` be an improvement?

GriffinRichards:
> `goto_if_undefeated` doesn't sit right with me. To me, the word "undefeated" implies at least one victory. Would `goto_if_not_defeated` be an improvement? The way it currently is, it reads as if the player has fought the trainer multiple times and never lost.

Sure. Technically undefeated is correct but it probably carries that connotation for most people. `not_defeated` would be less confusing then.

asm/macros/event.inc
```diff
@@ -469,14 +469,14 @@
 	.endm
 
 	@ Attempts to add quantity of item index to the player's Bag. If the player has enough room, the item will be added and VAR_RESULT will be set to TRUE; otherwise, VAR_RESULT is set to FALSE.
-	.macro giveitem index:req, quantity:req
+	.macro additem index:req, quantity=1
```
huderlem: This name change is a very intrusive change for people like Diego, as any existing `giveitem` calls will still compile, but have different behavior.  The name change should still happen, since the names are better, but it will cause some people to tear their hair out if you don't communicate this.

GriffinRichards: Addressed in Discord

GriffinRichards: Updating macros per discussion, dont merge until resolved

GriffinRichards: Macros updated. Should just be waiting on making sure people know about the change to avoid silent conflicts

PR #966, opened 2020-02-02
Use friendship event constants

GriffinRichards:


src/pokemon.c
```diff
@@ -5779,18 +5788,16 @@ void AdjustFriendship(struct Pokemon *mon, u8 event)
...
         if (friendship > 199)
             friendshipLevel++;
-        if ((event != 5 || !(Random() & 1))
-         && (event != 3
-          || ((gBattleTypeFlags & BATTLE_TYPE_TRAINER)
-           && (gTrainers[gTrainerBattleOpponent_A].trainerClass == TRAINER_CLASS_ELITE_FOUR
-            || gTrainers[gTrainerBattleOpponent_A].trainerClass == TRAINER_CLASS_LEADER
-            || gTrainers[gTrainerBattleOpponent_A].trainerClass == TRAINER_CLASS_CHAMPION))))
+
+        if ((event != FRIENDSHIP_EVENT_WALKING || !(Random() & 1))
```
huderlem: This boolean expression is pretty hard to understand.  Can you check if the following matches, please?
```
if ((event == FRIENDSHIP_EVENT_WALKING && (Random() & 1)
    || (event == FRIENDSHIP_EVENT_LEAGUE_BATTLE && IS_LEAGUE_BATTLE))
```

GriffinRichards: It doesn't, but I don't think it's logically the same. The boolean needs to allow all friendship events, but only allow those two if they pass the additional tests respectively. Adding `(event != FRIENDSHIP_EVENT_WALKING && event != FRIENDSHIP_EVENT_LEAGUE_BATTLE) || ` to the beginning of your suggestion makes it logically equivalent but it also doesnt match. Could add a comment briefly explaining it?

huderlem: ah, right

PR #967, opened 2020-02-04
Name Wally struct and use bag macros

segabl:


src/item_menu.c
```diff
@@ -445,7 +445,7 @@ EWRAM_DATA struct BagStruct gBagPositionStruct = {0};
 static EWRAM_DATA struct ListBuffer1 *sListBuffer1 = 0;
 static EWRAM_DATA struct ListBuffer2 *sListBuffer2 = 0;
 EWRAM_DATA u16 gSpecialVar_ItemId = 0;
-static EWRAM_DATA struct TempWallyStruct *gUnknown_0203CE80 = 0;
+static EWRAM_DATA struct TempWallyStruct *gTempWallyBag = 0;
```
GriffinRichards: `sTempWallyBag`, use `s` prefix for static data

segabl: Ah yeah, small oversight on my part

PR #979, opened 2020-02-15
Fix incorrect move relearner field

segabl:


segabl:
Nvm, saw griffin already did the same thing in #978

PR #981, opened 2020-02-17
Document battle animations

GriffinRichards:
Draft for now. Doing as much of the battle anim scripts and their functions as I can stand

Many of the sprite callbacks in this PR use the `Anim` prefix instead of the usual `SpriteCB_`, which could be standardized later. That's just what the pre-existing ones had


data/battle_anim_scripts.s
```diff
@@ -134,7 +136,7 @@ gBattleAnims_Moves::
 	.4byte Move_FOCUS_ENERGY
 	.4byte Move_BIDE
 	.4byte Move_METRONOME
-	.4byte Move_MIRROR_MOVE @ doesn't have an actual animation
+	.4byte Move_MIRROR_MOVE @ doesnt have an actual animation
```
huderlem: You introduced a typo?

src/battle_anim_effects_2.c
```diff
@@ -78,7 +78,7 @@ static void AnimTask_SplashStep(u8);
 static void AnimTask_GrowAndShrinkStep(u8);
 static void ThrashMoveMonStep(u8);
 static void ThrashMoveMon(u8);
-static void AnimTask_SketchDrawMon(u8);
+static void AnimTask_SketchDrawMonStep(u8);
```
huderlem: There is no reason to add the `Step` prefix if it's a single task without any additional steps.

src/battle_anim_effects_2.c
```diff
@@ -2998,29 +3001,29 @@ void AnimSpeedDust(struct Sprite *sprite)
     }
 }
 
-void sub_8105CB4(u8 taskId)
+void AnimTask_LoadMusicNotesPal(u8 taskId)
```
huderlem: `LoadMusicNotePals`

src/battle_anim_effects_2.c
```diff
@@ -2998,29 +3001,29 @@ void AnimSpeedDust(struct Sprite *sprite)
...
-    for (i = 0; i < 3; i++)
+    for (i = 0; i < NUM_MUSIC_NOTE_PAL_TAGS; i++)
         LoadPalette(&gMonSpritesGfxPtr->field_17C[i * 32], (u16)((paletteNums[i] << 4) + 0x100), 32);
 
     FREE_AND_SET_NULL(gMonSpritesGfxPtr->field_17C);
     DestroyAnimVisualTask(taskId);
 }
 
-void sub_8105D60(u8 taskId)
+void AnimTask_FreeMusicNotesPal(u8 taskId)
```
huderlem: `FreeMusicNotePals`

src/battle_anim_utility_funcs.c
```diff
@@ -681,7 +683,7 @@ static void sub_81175C4(u32 selectedPalettes, u16 color)
     }
 }
 
-void sub_8117610(u8 taskId)
+void AnimTask_BlendAllBattlerPalettesButAttacker(u8 taskId)
```
huderlem: `BlendNonAttackerPalettes`?

src/battle_anim_utility_funcs.c
```diff
@@ -999,9 +1005,9 @@ void sub_8117DD8(u8 taskId)
 void AnimTask_IsContest(u8 taskId)
 {
     if (IsContest())
-        gBattleAnimArgs[7] = 1;
+        gBattleAnimArgs[7] = TRUE;
```
huderlem: There is a define `ARG_RET_ID`  for index 7

src/battle_anim_utility_funcs.c
```diff
@@ -1016,14 +1022,14 @@ void sub_8117E94(u8 taskId)
 void AnimTask_IsTargetSameSide(u8 taskId)
 {
     if (GetBattlerSide(gBattleAnimAttacker) == GetBattlerSide(gBattleAnimTarget))
-        gBattleAnimArgs[7] = 1;
+        gBattleAnimArgs[7] = TRUE;
```
huderlem: `ARG_RET_ID`

src/battle_anim_utility_funcs.c
```diff
@@ -1036,6 +1042,7 @@ void sub_8117F30(u8 taskId)
     DestroyAnimVisualTask(taskId);
 }
 
+// TODO: Rename. Also used by Faint Attack
```
huderlem: Let's rename it now, instead of adding a TODO

src/normal.c
```diff
@@ -187,40 +187,41 @@ const struct SpriteTemplate gBattleAnimSpriteTemplate_8597370 =
...
     .oam = &gOamData_AffineNormal_ObjBlend_32x32,
     .anims = gDummySpriteAnimTable,
     .images = NULL,
     .affineAnims = gUnknown_08597348,
-    .callback = sub_8116458,
+    .callback = AnimHitSplatRandom,
 };
 
-const struct SpriteTemplate gBattleAnimSpriteTemplate_85973B8 =
+// TODO: Needs generic descriptive name, what distinguishes this hit splat
```
huderlem: Same as above.  Try not to leave TODOs if this can be renamed now.

src/normal.c
```diff
@@ -413,55 +415,68 @@ void sub_81159B4(struct Sprite *sprite)
     sprite->callback(sprite);
 }
 
-void sub_8115A04(u8 taskId)
+// Task data for AnimTask_BlendColorCycle
+#define tPalSelector   data[0] // arg0
```
huderlem: These `// arg*` comments don't seem helpful to me.  It's obvious when reading the function.

huderlem: These `// arg*` comments don't seem helpful to me.  It's obvious when reading the function.

src/battle_anim_effects_2.c
GriffinRichards: Sure. I just need a way to distinguish it from `AnimTask_SketchDrawMon`. I've been avoiding giving everything in the battle scripts file an `Init` prefix

src/battle_anim_utility_funcs.c
GriffinRichards: These are for me, theyll be resolved before the PR is marked ready

GriffinRichards:
Still plenty more to be done for battle anims but done with this for now.

huderlem:
This PR should also document the `setarg` battle script command.  For example: `setarg 7, -1` should really be something like `setarg ARG_RET_ID, 0xFFFF`, since the various handlers compare `ARG_RET_ID` to `0xFFFF`.

huderlem:
And in fact, maybe the better thing is to un-use `ARG_RET_ID` in many of the places your global find and replace changed, since that's not really what it is used for in those situations.

GriffinRichards:
Left a few of the `ARG_RET_ID` that are explicitly returns, `jumpreteq` makes it at least somewhat useful. Any more thoughts on merging the files? I'm a little hesitant because of how unwieldy it might make it

PR #985, opened 2020-02-23
Document decoration.c

mrgriffin:
Small changes around decorations:
* Use `decoration` for structure members (e.g. `decorDesk` => `decorationDesks`, `playerRoomDecors` => `playerRoomDecorations`).
* Use `decor` as the name for decoration parameters/variables.
* Use `category` as the name for decoration categories parameters/variables.
* Make `src/decoration_inventory.c` more consistent with the style in `src/decoration.c`.
* Attempt to name unknown variables, unnamed functions, and task data.
* Make variables and functions `static` where possible.
* Introduce an enum for the window IDs.
* (Hackily) use `BG_TILE_*_FLIP` in the code that constructs a sprite.

Definitely not confident in all my choices for names, any suggestions are welcome, particularly if they make the naming scheme more consistent with the rest of the code.

Document is probably a strong word too. Please suggest a better commit message!


data/specials.inc
```diff
@@ -31,7 +31,7 @@ gSpecials:: @ 81DBA64
 	def_special GetObjectEventLocalIdByFlag
 	def_special GetSecretBaseTypeInFrontOfPlayer
 	def_special SetSecretBaseOwnerGfxId
-	def_special sub_8129708
+	def_special ScrSpecial_PutAwayDecorationIteration
```
GriffinRichards: Prefix isn't needed where the name doesn't conflict with other functions

src/decoration.c
```diff
@@ -494,7 +520,7 @@ static u8 AddDecorationWindow(u8 windowIndex)
     if (windowIndex == 0)
     {
         template = sDecorationWindowTemplates[0];
-        template.width = GetMaxWidthInMenuTable(sDecorationMainMenuActions, ARRAY_COUNT(sDecorationMainMenuActions));
+        template.width = GetMaxWidthInMenuTable(sDecorationMainMenuActions, NELEMS(sDecorationMainMenuActions));
```
GriffinRichards: Please use `ARRAY_COUNT`. It's the more readable of the two macros (and the standard for pokeemerald).

include/global.h
```diff
@@ -928,16 +928,16 @@ struct SaveBlock1
...
+    /*0x271C*/ u8 playerRoomDecorations[DECOR_MAX_PLAYERS_HOUSE];
+    /*0x2728*/ u8 playerRoomDecorationPositions[DECOR_MAX_PLAYERS_HOUSE];
+    /*0x2734*/ u8 decorationDesks[10];
+    /*0x273E*/ u8 decorationChairs[10];
+    /*0x2748*/ u8 decorationPlants[10];
+    /*0x2752*/ u8 decorationOrnaments[30];
+    /*0x2770*/ u8 decorationMats[30];
+    /*0x278E*/ u8 decorationPosters[10];
+    /*0x2798*/ u8 decorationDolls[40];
+    /*0x27C0*/ u8 decorationCushions[10];
```
PikalaxALT: Any particular reason for the increased verbosity?

src/decoration.c
```diff
@@ -214,9 +241,9 @@ static const u8 *const sSecretBasePCMenuItemDescriptions[] =
 
 static const TaskFunc sSecretBasePC_SelectedDecorationActions[][2] =
 {
-   { sub_8127F68, sub_8127A8C },
-   { sub_812A3D4, sub_8127A8C },
-   { sub_8133DA0, sub_8127A8C },
+   { DecorationItemsMenuAction_AttemptPlace, DecorationItemsMenuAction_Cancel },
+   { DecorationItemsMenuAction_AttemptToss, DecorationItemsMenuAction_Cancel },
+   { DecorationItemsMenuAction_Trade, DecorationItemsMenuAction_Cancel },
```
PikalaxALT: Align the second column, or rearrange them like
```c
{
  pointer1,
  pointer2
}, {
  // ...
```

include/global.h
mrgriffin: The `SecretBase` struct uses "decoration", and I assumed that the whole word would be preferred to the abbreviated form. Happy to switch everything to "decor" if you prefer. Or could leave them mixed, but I think that comes across a bit weird?

src/decoration.c
```diff
@@ -78,10 +105,10 @@ EWRAM_DATA static u16 sDecorationsScrollOffset = 0;
 EWRAM_DATA u8 gCurDecorationIndex = 0;
 EWRAM_DATA static u8 sCurDecorationCategory = DECORCAT_DESK;
 EWRAM_DATA static u32 filler_0203a174[2] = {};
-EWRAM_DATA struct DecorationPCContext gDecorationContext = {};
+EWRAM_DATA static struct DecorationPCContext sDecorationContext = {};
 EWRAM_DATA static u8 sDecorMenuWindowIds[4] = {};
```
mrgriffin: I should introduce a `WINDOW_COUNT` to the enum, and use that in place of the 4 here.

include/global.h
huderlem: For what it's worth, I have never liked the `decor` abbreviation and would prefer `decoration`, in general.

PR #986, opened 2020-02-23
Sync Unown and qmark mon folders with pokefirered

GriffinRichards:


huderlem:
This deviates from the structure used by every normal pokemon.  Are we sure this is better?

GriffinRichards:
It deviates from the other pokemon either way, either they're organized by folder or they're organized by name suffix. I think it's easier to navigate with them in folders, and it gives some name consistency (e.g. all the icon files are `icon.png`). Castform should probably do the same. For the question marks, that folder already exists to hold the footprint and icon. It should hold the rest of the related images too.

PR #995, opened 2020-03-06
Ignores mGBA's first savestate file.

AsparagusEduardo:


GriffinRichards:
Why only the first savestate? `*.ss[0-9]` or `*.ss*` would ignore all of them

AsparagusEduardo:
Basically because I saw that only *.sa1 (VBA's first savestate) was in .gitignore.
I should really just add all of them.

PR #998, opened 2020-03-08
Slight item_menu.c documentation

AsparagusEduardo:


src/item_menu.c
```diff
@@ -2173,19 +2175,19 @@ void Task_WallyTutorialBagMenu(u8 taskId)
 
     if (!gPaletteFade.active)
     {
-        switch (data[8])
+        switch (tItemCount)
```
huderlem: `tItemCount` is not appropriate here.  `data[8]` is just some state tracker in this context.

PR #1002, opened 2020-03-11
Sync map header property names with pokefirered

GriffinRichards:
Syncing with pret/pokefirered#258.
The bike property name change is a bit superfluous, but the escape rope property also applies to Dig and should reflect that. 

NOTE: porymap also needs these changes, don't merge until it's ready


GriffinRichards:
Closing in favor of #1010

PR #1008, opened 2020-03-20
Index heal location data, use better names

GriffinRichards:
NOTE: Do not merge until a version of porymap containing huderlem/porymap#192 is released.


GriffinRichards:
Closing in favor of #1010

PR #1014, opened 2020-04-02
Re-export tileset anim pngs to use palettes

GriffinRichards:


PikalaxALT:
Stormy water looks weird -- palette only makes sense for Groudon's platform, not for Kyogre's

GriffinRichards:
> Stormy water looks weird -- palette only makes sense for Groudon's platform, not for Kyogre's

Looks like Groudon's uses palette 10 and Kyogre's uses palette 9. Resolved by splitting them into separate images

PR #1015, opened 2020-04-02
[Don't Merge Yet] Split voicegroups into individual files.

PlatinumMaster:


PlatinumMaster:
I will provide a script that makes migration for this easier if needs be. I just need to polish it.

PlatinumMaster:
Travis build failed due to the voice_groups.inc being out of order. Sorting algorithms...

PlatinumMaster:
I will open a new PR with proper changes, and instructions.

PR #1016, opened 2020-04-04
Split voicegroups and cry tables properly.

PlatinumMaster:
Here is a commit that contains the results of the instructions below.

Here is the script which is referred to below. [fixup script](https://gist.github.com/PlatinumMaster/af1dc8d7501ed0e02a687e0fa0f236fc)

```
Disclaimer: My script may not be perfect, but it is servicable. If you would like to improve it, I will have no problem with that.

1) Open sound/voice_groups.inc, and search for "gCryTable" and "gCryTable2". They should be next to each other. 
2) Starting from the ".align 2" above the first gCryTable, copy both tables into sound/cry_tables.inc (this needs to be created by you), then delete them from voice_groups.inc.
3) Download and run the fixup script, making sure it is placed in the pokeemerald root directory.

4) The pathway you follow depends on what you are achieving.
If you are ensuring that you pass the checksum comparison (also known as OK):
 - Open sound/voice_groups.inc, and find the line (.include "sound/voicegroups/voicegroup128.inc").
 - Right after that line, add this line: (.include "sound/cry_tables.inc").
 - Save changes.
 
Otherwise:
 - Open data/sound_data.s.
 - Under the line (.include "sound/voice_groups.inc"), add (.include "sound/cry_tables.inc").
 - Save changes.
 
5) If you are satisfied with your changes, delete voice_groups.bak from the sound folder.

You're done! 
```

This should match, as I have tested multiple times with clean copies.


PlatinumMaster:
Fixed the travis build. Issue seemed to be related to the devkitARM package being outdated.

PlatinumMaster:
Bumped DKP properly.

GriffinRichards:
Any reason this splits the cry tables into separate files? It's also confusing that `gCryTable2` is in a file called `gCryTable1.inc`

PlatinumMaster:
Well I split them to separate files for easier editing, however I could just leave them alone if needs be.

GriffinRichards:
They should be in a single file called `cry_tables.inc`. Being in separate files makes them harder to edit, these two tables will essentially always be edited together

PlatinumMaster:
If that's the case I can merge them.

PR #1022, opened 2020-04-13
Label GCN link flags field

GriffinRichards:


PikalaxALT:
See comment on pret/pokefirered#329

PR #1029, opened 2020-04-29
Fix -Wmissing-braces clang warning

GriffinRichards:
Silences `error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]` from building gbagfx with -Werror.


Sierraffinity:
It was suggested that the C version should be changed to gnu11 as the empty brace initializer is a GNU extension and not a default C11 feature. As it is right now, though, it will not complain about C11 violations.

GriffinRichards:
Yeah @jiangzhengwenjz suggested an alternate solution, I'm happy to close this if they'd like to implement it. Otherwise this is needed for some people to build the project

PR #1030, opened 2020-05-01
gfx: fix hpbar_anim.png visual palette

SBird1337:
fixes the hp bar animation image, s.t. its colors match what they actually look ingame.

right:
hpbar_anim
wrong:
hpbar_anim


SBird1337:
Ready for Review/Merge once CI has completed

Sierraffinity:
Would you also merge the incbins into one in graphics.c? The format should be `INCBIN_U8("file1", "file2");`

SBird1337:
I'm not fully sure about that last one, maybe they should just be one file, in the end they do share a palette and are addressed like one file, does it really make sense to have individual files?

Sierraffinity:
It does make sense to keep them as separate files, because they _don't_ all share the same palette. For example, take the status icons - they each have a different color but all the colors share the same slot.

PR #1031, opened 2020-05-02
Document minor berry yield bug

hondew:
Minor error in berry yield calculation skews towards higher results.

### DIFF
--- a/include/berry.h
+++ b/include/berry.h
@@ -1,6 +1,8 @@
 #ifndef GUARD_BERRY_H
 #define GUARD_BERRY_H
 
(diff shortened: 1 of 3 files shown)


include/berry.h
```diff
@@ -1,6 +1,8 @@
 #ifndef GUARD_BERRY_H
 #define GUARD_BERRY_H
 
+#define NUM_BERRY_STAGES 4
```
GriffinRichards, line 4: This should be moved to `include/constants/berry.h` and clarified, as berry trees have 6 stages of growth

hondew: Done.

PR #1036, opened 2020-05-09
add BUGFIX macro

ipatix:
The BUGFIX macro can be used to fix small bugs in the game if matching is not required. This, for example, is the case when using a modern compiler which might not like the bugs and will produce non-matching results anyway.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -26,4 +26,12 @@
 #define UNITS_METRIC
 #endif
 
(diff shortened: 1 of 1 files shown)


PikalaxALT:
Make this change in the Makefile instead.

ipatix:
I'll reopen once I get the time to fix this properly.

PR #1040, opened 2020-05-15
Fix building modern from clean

PikalaxALT:


ipatix:
Oh, I guess that's why I sometimes had problems with that. Thank you very much.

PR #1047, opened 2020-05-19
Make the code clearer

PokeCF:
The counter stops if the number is greater than or equal to 1000, which is more or less the same as greater than 999, but it makes the intent clearer that at 1000, the timer is set to max

### DIFF
--- a/src/play_time.c
+++ b/src/play_time.c
@@ -24,7 +24,7 @@ void PlayTimeCounter_Start(void)
 {
     sPlayTimeCounterState = RUNNING;
 
-    if (gSaveBlock2Ptr->playTimeHours > 999)
+    if (gSaveBlock2Ptr->playTimeHours >= 1000)
         PlayTimeCounter_SetToMax();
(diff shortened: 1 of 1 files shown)


atasro2:
`void PlayTimeCounter_SetToMax(void)
{
    sPlayTimeCounterState = MAXED_OUT;

    gSaveBlock2Ptr->playTimeHours = 999;
    gSaveBlock2Ptr->playTimeMinutes = 59;
    gSaveBlock2Ptr->playTimeSeconds = 59;
    gSaveBlock2Ptr->playTimeVBlanks = 59;

}`

i would like to disagree because of this, it would make more sense for GF to do the previous thing

aaaaaa123456789:
This PR is so trivial that it's an amazingly good candidate for [bikeshedding](https://shed.bike/).

That being said, it's also so trivial as to contribute basically nothing at all.

PikalaxALT:
What they said.

PR #1048, opened 2020-05-19
set Pointers to NULL, rather than 0

PokeCF:
NULL is for pointers, not 0.

Also, according to the disassembly, a more accurate decompilation would be to break when (*func)(ctx) is 1, not return true directly.

### DIFF
--- a/src/script.c
+++ b/src/script.c
@@ -26,9 +26,9 @@ void InitScriptContext(struct ScriptContext *ctx, void *cmdTable, void *cmdTable
     s32 i;
 
     ctx->mode = 0;
-    ctx->scriptPtr = 0;
+    ctx->scriptPtr = NULL;
     ctx->stackDepth = 0;
-    ctx->nativePtr = 0;
+    ctx->nativePtr = NULL;
     ctx->cmdTable = cmdTable;
     ctx->cmdTableEnd = cmdTableEnd;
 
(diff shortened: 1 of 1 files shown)


aaaaaa123456789:
Failed CI, so I assume this doesn't match. This is why the code is written as is and not using a `break`.

huderlem:
CI actually failed for non-code reasons:
```
145.13s$ sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"
Error: retrieving gpg key timed out.
The command "sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"" failed and exited with 1 during .
```

PokeCF:
So this is not related to the code itself, huh?

src/script.c
```diff
@@ -255,7 +255,7 @@ u8 *MapHeaderGetScriptTable(u8 tag)
         if (*mapScripts == tag)
         {
             mapScripts++;
-            return (u8 *)(mapScripts[0] + (mapScripts[1] << 8) + (mapScripts[2] << 16) + (mapScripts[3] << 24));
+            return (u8 *)(*mapScripts + (*(mapScripts + 1) << 8) + (*(mapScripts + 2) << 16) + (*(mapScripts + 3) << 24));
```
ketsuban, line 258: Why did you do this?

huderlem: Yes, please explain the reasoning behind this commit.  It is wholly unnecessary and makes the code more difficult to read..

PokeCF: Because it is accessing a memory location, not an array, which the original code was misleading about.

PokeCF: And you are already checking the value of the map scripts pointer already, so consistency too

PikalaxALT:
Proposed change is insignificant and undesirable. Closing.

PR #1050, opened 2020-05-20
Sync math util with pokefirered

GriffinRichards:


Sierraffinity:
Might be outside the scope of this PR, but the values used as arguments for these functions should really be using the fixed point macros.

GriffinRichards:
Updated. A few (e.g. the assignments to `b` and `c` in `AnimAirWaveProjectile`) will only match if the macro uses `<< 8` instead of `* 256`, which would seem to preclude using decimals elsewhere

Sierraffinity:
That's probably something like the team1 vs team2 thing, where two macros exist to use the same thing.

PR #1052, opened 2020-05-20
Decompile unused decompress function, as seen in ruby

Missingmew:
Refer to https://github.com/pret/pokeruby/pull/783 for details


huderlem:
tabs and `unsigned char`

PR #1068, opened 2020-06-03
Shop: Document some object event constants

SBird1337:
Some stray `16`'s should be `OBJECT_EVENTS_COUNT` - this breaks some possible modifications to this constant.


src/shop.c
```diff
@@ -839,9 +839,9 @@ static void BuyMenuDrawObjectEvents(void)
     u8 spriteId;
     const struct ObjectEventGraphicsInfo *graphicsInfo;
 
-    for (i = 0; i < 16; i++) // max objects?
+    for (i = 0; i < OBJECT_EVENTS_COUNT; i++) // max objects?
```
GriffinRichards, line 842: This comment can be removed now

PR #1069, opened 2020-06-03
Document linking

GriffinRichards:
Mix of new documentation and documentation ported from pokefirered. 
Still plenty more to do for link rfu and union room, but should reduce the work a good deal.
`cable_club.c`, `wireless_communication_status_screen.c`, and the supplementary union room files (`union_room_chat.c`, `union_room_battle.c`, `union_room_player_avatar.c`) are basically done.
Added all the missing chars from the charmap to `text.h`, mostly to use them in union room chat.


src/link_rfu_2.c
```diff
@@ -114,79 +179,83 @@ const char sUnref_082ED6E8[][15] = {
...
-const char gUnknown_082ED866[2] = {'*',0x00};
-const char gUnknown_082ED868[8] = "NOWSLOT";
-const char gUnknown_082ED870[12] = "           ";
-const char gUnknown_082ED87C[12] = "CLOCK DRIFT";
-const char gUnknown_082ED888[12] = "BUSY SEND  ";
-const char gUnknown_082ED894[12] = "CMD REJECT ";
-const char gUnknown_082ED8A0[12] = "CLOCK SLAVE";
-const char gUnknown_082ED8A8[3][8] = {
+
+static const char sASCII_PokemonSioInfo[] = "PokemonSioInfo";
```
nullableVoidPtr, line 197: Some of these string constants can be inlined.

GriffinRichards: Why should we? pokefirered does this, but `sASCII_PokemonSioInfo` for example is used more than once. Seems just as reasonable to collect all these debug strings in one place rather than scatter them

common_syms/link.txt
```diff
@@ -3,15 +3,15 @@ gLinkDebugSeed
 gLocalLinkPlayerBlock
 gLinkErrorOccurred
 gLinkDebugFlags
-gFiller_03003074
+gLinkFiller1
```
PikalaxALT, line 6: Is it possible to remove some of these fillers and let section alignment handle things?

GriffinRichards: Ah, `gLinkFiller6` and 7 can be yeah. The others would presumably at least require file splits

src/link_rfu_2.c
PikalaxALT: iirc agbcc does optimize the assembly of multiply-referenced C strings

GriffinRichards: I don't see a reason we should make that change

PR #1076, opened 2020-06-13
Move Pokemon cries to own directory.

surskitty:
sound/direct_sound_samples seemed egregiously large, and the Pokemon cries were already named understandably. I think it makes more sense to put them in their own directory.


GriffinRichards:
This doesn't match. The Makefile needs to be updated to reflect these changes

PR #1082, opened 2020-06-24
Allow flag/var counts to be updated alongside values

GriffinRichards:
The main reason for this is to make adding new trainers at the cost of save space easier for beginners; it now only requires increasing `MAX_TRAINERS_COUNT`. `FLAGS_COUNT` and `VARS_COUNT` are now calculated from existing constants


include/constants/flags.h
```diff
@@ -1573,7 +1573,7 @@
 #define FLAG_UNUSED_0x91F                           (SYSTEM_FLAGS + 0xBF) // Unused Flag
 
 // Daily Flags
-#define DAILY_FLAGS_START                           0x920
+#define DAILY_FLAGS_START                           (FLAG_UNUSED_0x91F + 1)
```
PikalaxALT, line 1576: Maybe make this relative to SYSTEM_FLAGS, and/or force alignment? idk whether the daily flags are reset by clearing that part of the array directly

GriffinRichards: Yeah they're cleared directly (`ClearDailyFlags`). I'll force alignment, otherwise itll round down and potentially clear some system flags

PR #1087, opened 2020-07-05
fix aggressive loop optimizations

ipatix:
Previously, aggressive loop optimizations with a new compiler were not
possible due to undefined behaviour at end of arrays.

A macro "UBFIX" is added to allow ifdefs for fixes which resolve
undefined behavior. For example newer GCC versions will detect various
bugs in the original game code and will otherwise not compile with -Werror.

### DIFF
--- a/src/slot_machine.c
+++ b/src/slot_machine.c
@@ -384,7 +384,7 @@ static const u16 gSlotPayouts[];
 static const u8 *const gUnknown_083EDCE4;
 static const u8 *const gUnknown_083EDCDC;
 static const u32 gReelTimeGfx[];
-static const struct SpriteSheet gSlotMachineSpriteSheets[];
+static const struct SpriteSheet gSlotMachineSpriteSheets[22];
 static const struct SpritePalette gSlotMachineSpritePalettes[];
 static const u16 *const gUnknown_083EDE20;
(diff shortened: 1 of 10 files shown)


src/slot_machine.c
```diff
@@ -6727,15 +6727,12 @@ static const struct SpriteSheet gSlotMachineSpriteSheets[] =
...
-    {0, 0, 0, 0, 0, 2, 19}, 
-    {0, 0, 0, 0, 0, 3, 20}, 
-    {0, 0, 0, 0, 0, 3, 21}, 
-    {0, 0, 0, 0, 0, 0,  0}
+    // the data for these sheets is determined at runtime
+    { .data = NULL, .size = 0x200, .tag = 18 },
+    { .data = NULL, .size = 0x200, .tag = 19 },
+    { .data = NULL, .size = 0x300, .tag = 20 },
+    { .data = NULL, .size = 0x300, .tag = 21 },
+    { .data = NULL, .size = 0x0, .tag = 0 },
```
PikalaxALT, line 6735: Terminator element does not need explicit field initializers.

ipatix: Well, yes. I mainly wrote it like that due to readability.
Should I remove the last line?

ipatix:
Updated.

PR #1090, opened 2020-07-09
Made documentation more readable

Ryuhouji:
-Snip- Finished the correct PR, updated 'delay' macro and made the documentation more digestible.

### DIFF
--- a/asm/macros/event.inc
+++ b/asm/macros/event.inc
@@ -274,15 +274,17 @@
 	.2byte \functionId
 	.endm
 
-	@ Blocks script execution until a command or ASM code manually unblocks it. Generally used with specific commands and specials. If this command runs, and a subsequent command or piece of ASM does not unblock state, the script will remain blocked indefinitely (essentially a hang).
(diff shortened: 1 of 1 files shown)


asm/macros/event.inc
```diff
@@ -315,7 +317,8 @@
 	.byte 0x2d
 	.endm
 
-	@ Sets the values of variables 0x8000, 0x8001, and 0x8002 to the current hour, minute, and second. In FRLG, this command sets those variables to zero.
+	@ Sets the values of variables 0x8000, 0x8001, and 0x8002 to the current hour, minute, and second. In FRLG, 
+        @ this command sets those variables to zero.
```
PikalaxALT, line 321: Indent `.s`, `.inc` files with tab character, not spaces.

GriffinRichards, line 321: Indentation should be tabs in this file

Ryuhouji: Updated with fix, I used the github editor, which doesn't automatically fix mistakes like vscode does. My bad.

PR #1105, opened 2020-07-20
Changed mon front and back coordinates from hex to decimal so they're easier to edit.

AsparagusEduardo:


huderlem:
Changing `.size` to decimal makes it much harder, since the high and low nybbles are the dimensions.

AsparagusEduardo:
I'll revert it then, one sec.

huderlem:
It's possible that `.size` is actually a bitfield with two 4-bit values, but I haven't ever attempted to see if it matches that way.

GriffinRichards:
> It's possible that `.size` is actually a bitfield with two 4-bit values, but I haven't ever attempted to see if it matches that way.

I believe I've tried that in the past and it doesn't match

GriffinRichards:
Check this over again, you missed a number of `y_offset`s that are still hex values

PR #1107, opened 2020-07-20
Change -mtune to -mcpu

PokeCodec:
ARM's own website says -mtune and -march should be replaced with -mcpu instead, so I did for the modern compilers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ OBJ_DIR := build/emerald
 LIBPATH := -L ../../tools/agbcc/lib
 else
 CC1              = $(shell $(CC) --print-prog-name=cc1) -quiet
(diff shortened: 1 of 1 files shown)


DizzyEggg:
According to this link https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu

As long as you’re not cross compiling, the simplest and easiest way to get the best performance on Arm with both GNU compilers and LLVM-compilers is to use only -mcpu=native and actively avoid using -mtune or -march.

As lon as you're not cross compiling, but that's exactly what we're doing! There was a reason pika decided to use this flags, and I think we should keep them as they are.

PokeCodec:
we are not using -mcpu=native, we are using -mcpu=arm7tdmi

PikalaxALT:
Closing due to the absence of evidence that this change is necessary for compatibility.

PR #1104, opened 2020-07-19
Reduce "Implicit conversion" warnings

PokeCodec:
Modern compilers tend to complain about this implicit casting, so I added an explicit cast in these areas.

PR #1110, opened 2020-07-22
Change (s32) to (uintptr_t) in agb flash

PokeCodec:
to allow for better portability


PikalaxALT:
There's no real reason why this PR should be separate from #1104.

PokeCodec:
If you want it to be part of it, sure

PR #1114, opened 2020-07-27
Document some battle_util.c

ExpoSeed:


src/battle_util.c
```diff
@@ -259,7 +259,7 @@ void ResetSentPokesToOpponentValue(void)
         gSentPokesToOpponent[(i & BIT_FLANK) >> 1] = bits;
 }
 
-void sub_803F9EC(u8 battler)
+void OpponentSwitchInResetSentPokesToOpponent(u8 battler)
```
GriffinRichards, line 262: Could you explain this name? What it means at a glance is not clear

ExpoSeed: That function is called when an opponent switches in a new mon

When called, the function will clear the previous battler's `gSentPokesToOpponent` data and set the correct bits for the mons on the field (it does the same thing as `ResetSentPokesToOpponentValue` but for only one flank instead of both flanks)

I was struggling with a good name for that function, so if you have other ideas, that may work

Also I realize I forgot to add Value to the end of the function name like the other functions have

GriffinRichards: Thanks. Yeah it's a little wordy, but it's fine for now

PR #1117, opened 2020-07-30
Mostly match

PokeCodec:
This is a draft, but basically all but one asm line differs. And this will remain a draft until that ONE line is fixed


GriffinRichards:
This is filled with needless indentation

PR #1123, opened 2020-08-06
Generate pokemon data from JSON using inja templates

rroelke:
I recently began to work on an edit of Emerald and found it very difficult to edit Pokemon data.

The purpose of this pull request is to centralize where Pokemon data is defined, and present it in a human-readable and human-writable format.

To achieve this goal, we centralize Pokemon data in a JSON data structure (```data/pokemon_species.json```) and use inja templates to generate source code from the JSON.  The JSON data is very easy to edit, and doing so will re-generate all of the pokemon data tables at the beginning of the next build.

The JSON data consists of two arrays: ```species``` and ```unown_forms```.  The ```species``` array is more comprehensive, listing all attributes of a pokemon species; whereas the ```unown_forms``` array only indicates graphical data.

The templates are generally straightforward, though some special consideration has to be given for the ```OLD_UNOWN_*``` species, which do not consistently show up in tables.  When these need special handling, it is typically done by checking that the species name starts with the ```OLD_UNOWN_``` string.  But in general, most templates begin with a table entry for ```SPECIES_NONE```; generate entries for all pokemon species using a loop template; write an entry for ```SPECIES_EGG```; and then write an entry for each unown form using another loop template.


rroelke:
The Makefile dependencies aren't quite right yet - a clean parallel build fails.  My makefile-ese is a bit weak so I could use an assist on that.

include/constants/species.h.template
```diff
@@ -0,0 +1,41 @@
+#ifndef GUARD_CONSTANTS_SPECIES_H
```
rroelke, line 1: Needs a ```{{ doNotModify }}```

GriffinRichards:
This seems like the kind of thing to me that's better left as a fork so people can choose whether or not to use it. The data we have in json is specifically so it can be used by tools, and I'm not sure how reasonable it would be to have a pokemon editor tool given the kinds of changes people are likely to make. Without tooling I'm not sure everyone would agree it's easier to edit inja templates and json than how it's laid out now.
The point is moot anyway if the build doesn't succeed and match

Ryuhouji:
I agree that we don't need this to be forced down peoples' throats. The json changes already made were enough to cause many people large headaches, myself included. I don't even know what inja is. I don't see what's so hard to edit about the existing system, anyway.

Toby313:
Can you explain why it is "very difficult to edit Pokemon data"? Editing pokemon data is the first thing I did in pokeemerald because of how easy it is, even though I had zero programming experience.

rroelke:
I would quantify difficulty of editing pokemon data as "number of files one has to edit".  If your goal is to change a learnset, then this tooling doesn't help you because that metric doesn't change.  The use case I built this for is "incrementally add new pokemon species".  Adding a new species requires editing ~25 files of source.  Given that this tooling reduces it to 1 that's a pretty clear improvement in difficulty.

I'll close this since it seems that what you expect people to use this repository for are not what I intend to use it for.

The build doesn't succeed at the moment because the Make dependencies don't generate the new source files before trying to compile the C files.  I had hoped that there would be more interest, and someone would help to fix that.

PR #1127, opened 2020-08-09
Glyph match

PokeCodec:
Good news: I was able to force the compiler to use ip instead of sl. Problem is that because I was not able to 100% fully decompile it perfectly, I had to use an NONMATCHING hack "asm(::"":::"sl");"


PokeCodec:
@GriffinRichards plz check when you can

PR #1129, opened 2020-08-10
Switch to syntax-unified

PokeCodec:
We have been writing all our asm in unified syntax. However, by default, gcc outputs the old syntax. -masm-unified-syntax ensures that the asm it outputs are syntax unified


PokeCodec:
Note: I did not change the asm for copyglyph since I have a different pr for that anyway, so accept the copyglyph pr first, and then I will fix and squash this one.

GriffinRichards:
I'm not really convinced any of this is necessary

PokeCodec:
considering syntax unified is the standard, it makes sense that ALL the asm is in the newer syntax, because having some in the legacy syntax but others in the unified syntax isn't good style.

kiliwily:
This pull request need to be fixed.

PokeCodec:
I merged all changes into one convenient PR

PR #1132, opened 2020-08-12
Apply bool8 type to u8 that act as bools

PokeCodec:
some fields, like textprinter->Japanese and autoscroll act as boolean values, and should be labeled as such, instead of using 1 or 0, true and false work, so I replaced the 1 and 0 being assigned to those u8 as true and false, with ! replacing == FALSE.

### DIFF
--- a/gflib/text.c
+++ b/gflib/text.c
@@ -21,7 +21,7 @@ static u16 gLastTextFgColor;
 static u16 gLastTextShadowColor;
 
 const struct FontInfo *gFonts;
-u8 gUnknown_03002F84;
+bool8 gUnknown_03002F84;
 struct Struct_03002F90 gUnknown_03002F90;
 TextFlags gTextFlags;
 
@@ -205,7 +205,7 @@ bool16 AddTextPrinter(struct TextPrinterTemplate *printerTemplate, u8 speed, voi
             CopyWindowToVram(gTempTextPrinter.printerTemplate.windowId, 2);
         gTextPrinters[printerTemplate->windowId].active = 0;
     }
-    gUnknown_03002F84 = 0;
+    gUnknown_03002F84 = FALSE;
     return TRUE;
 }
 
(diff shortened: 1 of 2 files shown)


gflib/text.h
```diff
@@ -354,10 +354,10 @@ struct KeypadIcon
...
 
 typedef struct {
-    bool8 canABSpeedUpPrint:1;
-    bool8 useAlternateDownArrow:1;
-    bool8 autoScroll:1;
-    bool8 forceMidTextSpeed:1;
+    bool8 canABSpeedUpPrint:TRUE;
+    bool8 useAlternateDownArrow:TRUE;
+    bool8 autoScroll:TRUE;
+    bool8 forceMidTextSpeed:TRUE;
```
GriffinRichards, line 360: These are numbers of bits, not boolean values

PokeCodec: the usage of these variables say otherwise, but ok. are you sure?

GriffinRichards: The `:1` indicates these fields are 1 bit in size. They're bools by definition of being 1 bit, but these 1's don't have anything to do with truth values

gflib/text.c
```diff
@@ -205,15 +205,15 @@ bool16 AddTextPrinter(struct TextPrinterTemplate *printerTemplate, u8 speed, voi
...
+    gUnknown_03002F84 = FALSE;
     return TRUE;
 }
 
 void RunTextPrinters(void)
 {
     int i;
 
-    if (gUnknown_03002F84 == 0)
+    if (gUnknown_03002F84)
```
GriffinRichards, line 216: This is incorrect (and presumably why the build doesn't match)

gflib/text.h
PokeCodec: Ok

PokeCodec:
all fixed!

PR #1137, opened 2020-08-14
Create Enum for Colors

PokeCodec:
I could define the different colors and document them via defines, but enums are easier and modern. Also I made slight changes to CopyGlyphtoWindow


gflib/text.h
```diff
@@ -269,6 +269,13 @@
 
 #define TEXT_SPEED_FF 0xFF
 
+enum ColorMode 
+{
+    FgColor,
+    ShadowColor,
+    BgColor
```
GriffinRichards, line 276: Use the naming convention of `CONSTANT_NAME`. The enum name also isn't necessary, it should be obvious what they're for

PokeCodec:
Done

gflib/text.h
PokeCodec: Done

PR #1140, opened 2020-08-16
Match battle script commands

PokeCodec:
Although it may appear unrelated changes are in this PR, they are necessary for the matching. The reason why the first function was not matching was because this is the only function in which the problem wasn't the function itself, but the data it is accessing.

When I inputted the structs into Ghidra, Ghidra showed that the size of the struct in the actual Emerald Rom is too big for the current struct declaration. Adjusting it allowed me to get rid of implicit casts that caused the nonmatching in C. The only function affected by this change other than Cmd_unknown32 was the money function.

### DIFF
--- a/include/constants/battle.h
+++ b/include/constants/battle.h
@@ -186,7 +186,7 @@
 #define HITMARKER_x4000000              (1 << 26)
 #define HITMARKER_CHARGING              (1 << 27)
 #define HITMARKER_FAINTED(battler)      (gBitTable[battler] << 28)
-#define HITMARKER_FAINTED2(battler)     (1 << (28 + battler))
+#define HITMARKER_FAINTED2(battler)     ((1 << 28) << battler)
 
 // Per-side statuses that affect an entire party
 #define SIDE_STATUS_REFLECT          (1 << 0)
--- a/src/battle_script_commands.c
+++ b/src/battle_script_commands.c
@@ -3489,7 +3489,6 @@ static void Cmd_getexp(void)
     }
 }
 
-#ifdef NONMATCHING
 static void Cmd_unknown_24(void)
 {
     u16 HP_count = 0;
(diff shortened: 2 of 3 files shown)


include/constants/battle.h
```diff
@@ -186,7 +186,7 @@
 #define HITMARKER_x4000000              (1 << 26)
 #define HITMARKER_CHARGING              (1 << 27)
 #define HITMARKER_FAINTED(battler)      (gBitTable[battler] << 28)
-#define HITMARKER_FAINTED2(battler)     (1 << (28 + battler))
+#define HITMARKER_FAINTED2(battler)     ((1 << 28) << battler) //(1 << (28 + battler))
```
PikalaxALT, line 189: Evaluate `1 << 28` as an integer literal i.e. `0x10000000`

include/global.h
```diff
@@ -97,6 +97,7 @@
 #define T2_READ_16(ptr) ((ptr)[0] + ((ptr)[1] << 8))
 #define T2_READ_32(ptr) ((ptr)[0] + ((ptr)[1] << 8) + ((ptr)[2] << 16) + ((ptr)[3] << 24))
 #define T2_READ_PTR(ptr) (void*) T2_READ_32(ptr)
+#define T2_READ_U8_PTR(ptr) (u8*) T2_READ_32(ptr)
```
PikalaxALT, line 100: Not necessary, just cast the result of the preceding macro to `(u8 *)` in source

src/battle_script_commands.c
```diff
@@ -3489,11 +3489,10 @@ static void Cmd_getexp(void)
     }
 }
 
-#ifdef NONMATCHING
 static void Cmd_unknown_24(void)
 {
-    u16 HP_count = 0;
     s32 i;
+    u16 HP_count = 0;
```
PikalaxALT, line 3495: Is switching the declaration order necessary to match?

include/constants/battle.h
GriffinRichards: Why? None of the other macros do, and bit shifts are easier to read

PokeCodec:
Done!

src/battle_script_commands.c
PokeCodec: I fixed it, so

PR #1150, opened 2020-08-24
Match trainer hill

PokeCodec:


GriffinRichards:
Looks like you included an assembly file suffixed `_OG` by mistake

PokeCodec:
My bad. I removed it.

PR #1154, opened 2020-08-26
Battle intro match

PokeCodec:
This is a draft I have decided to put up because despite having matching asm, the function is not matching. Hopefully someone can help me in this.


GriffinRichards:
Help requests can be kept to #asm2c-help, open a PR once it matches

PR #1160, opened 2020-08-28
add MAX_LEVEL_UP_MOVES macro

kiliwily:


GriffinRichards:
These arrays also depend on this value, it'd probably be helpful to have those use this constant too. They're not exact (move relearner uses 25, daycare 50), so you could do something like `#define MAX_RELEARNER_MOVES (MAX_LEVEL_UP_MOVES > 25 ? MAX_LEVEL_UP_MOVES : 25)`

https://github.com/pret/pokeemerald/blob/master/src/move_relearner.c#L157
https://github.com/pret/pokeemerald/blob/master/src/move_relearner.c#L160
https://github.com/pret/pokeemerald/blob/master/src/daycare.c#L34

kiliwily:
Done.

PR #1174, opened 2020-09-04
Make it easier to compile on modern compilers as well as replacing constants with macros where applicable

PokeCodec:
One common error that happens when using other toolchains is the need for an stdlib to be present even if the game does not need to do so. For this reason I added -nostdinc.

With this added, the asm needed for modern compilers is no longer needed: RESET_ALL works in agb main.

Finally, to reduce warnings, I put an if !MODERN directive on the infamous "Break after return".

In addition to these, I made small changes including rewriting the reset function so it the code reads that it is checking for "A, B, START, SELECT", as we have learned is the combination used to erase save data.

I have tested these changes locally and found only improvements.

Thank you!

### DIFF
--- a/src/battle_ai_script_commands.c
+++ b/src/battle_ai_script_commands.c
@@ -533,9 +533,7 @@ static u8 ChooseMoveOrAction_Doubles(void)
                 if (i == (sBattler_AI ^ BIT_FLANK) && bestMovePointsForTarget[i] < 100)
                 {
                     bestMovePointsForTarget[i] = -1;
-                    #ifndef NONMATCHING
                     mostViableMovesScores[0] = mostViableMovesScores[0]; // Needed to match.
-                    #endif
                 }
             }
         }
--- a/src/battle_ai_script_commands.c
+++ b/src/battle_ai_script_commands.c
@@ -533,7 +533,9 @@ static u8 ChooseMoveOrAction_Doubles(void)
                 if (i == (sBattler_AI ^ BIT_FLANK) && bestMovePointsForTarget[i] < 100)
                 {
                     bestMovePointsForTarget[i] = -1;
+                    #ifndef NONMATCHING
                     mostViableMovesScores[0] = mostViableMovesScores[0]; // Needed to match.
+                    #endif
                 }
             }
         }
--- a/src/link_rfu_3.c
+++ b/src/link_rfu_3.c
@@ -638,7 +638,10 @@ static u8 GetConnectedChildStrength(u8 maxFlags)
                 if (maxFlags == flagCount + 1)
                 {
                     return gRfuLinkStatus->strength[i];
+
+                #if !(MODERN)//Because some modern compilers will throw a warning
                     break; // This break is needed to match
+                #endif
                 }
                 flagCount++;
             }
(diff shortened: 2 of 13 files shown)


src/main.c
```diff
@@ -85,27 +85,7 @@ void EnableVCountIntrAtLine150(void);
 
 void AgbMain()
 {
-#if MODERN
```
atasro2, line 88: These was a reason this was added you can't just say it isn't needed anymore if one of your modern builds doesn't blow up because of it and oh hey i just read the comment so `RegisterRamReset **may** crash if it resets IWRAM.`

src/battle_ai_script_commands.c
```diff
@@ -533,7 +533,9 @@ static u8 ChooseMoveOrAction_Doubles(void)
                 if (i == (sBattler_AI ^ BIT_FLANK) && bestMovePointsForTarget[i] < 100)
                 {
                     bestMovePointsForTarget[i] = -1;
+                    #ifndef NONMATCHING
```
atasro2, line 536: the original purpose of these fake matches were to avoid ifdefs like this, this is just defeating the point

src/link_rfu_3.c
```diff
@@ -631,14 +628,17 @@ static u8 GetConnectedChildStrength(u8 maxFlags)
...
-        for (i = 0; i < 4; i++)
+        for (i = 0; i < RFU_CHILD_MAX; i++)
         {
             if (flags & 1)
             {
                 if (maxFlags == flagCount + 1)
                 {
                     return gRfuLinkStatus->strength[i];
+
+                #if !(MODERN)//Because some modern compilers will throw a warning
```
atasro2, line 639: a warning or two doesn't hurt and won't blow everything up

src/link_rfu_3.c
```diff
@@ -925,8 +925,7 @@ void RecordMixTrainerNames(void)
 {
     if (gWirelessCommType != 0)
     {
-        s32 i;
-        s32 j;
+        int i, j;
```
atasro2: changing the types for these loop iterators is useless

src/main.c
```diff
@@ -130,22 +110,24 @@ void AgbMain()
         SetMainCallback2(NULL);
 
     gLinkTransferringData = FALSE;
-    gUnknown_03000000 = 0xFC0;
+    gUnknown_03000000 = sizeof(struct Link);
```
atasro2, line 113: where did this come from?

src/main.c
PokeCodec: Not a single compiler I tried from 9.3 onwards has had this issue. Not one.

atasro2: you can't rule it out

atasro2: what do you know maybe someone using 9.1 will have this issue and we can't do anything about it

src/main.c
PokeCodec: We are in a link related function.

Not to mention this is in the beginning of ewram, and the link struct is this size, so I am certain this is where the number came from

atasro2: this is the main loop of the game but ok and that address is very much in IWRAM

src/main.c
PokeCodec: 9.1 Works. Want me to test any other compiler?

PR #1175, opened 2020-09-04
Adjustments for Modern so compiling on other toolchains is easier

PokeCodec:
One common error that happens when using other toolchains is the need for an stdlib to be present even if the game does not need to do so. For this reason I added -nostdinc.

Additionally, the asm needed for modern compilers is no longer needed: RESET_ALL works in agb main.


PikalaxALT:
Closing without merge because the changes proposed are counterproductive.

PR #1185, opened 2020-09-11
Created more enums, made more macros, and documented some functions

PokeCodec:
Until I manage to finish fixing the fakematches, I thought I'd send in a PR with some more documentation done with macros.

PR #1188, opened 2020-09-13
Created more enums, made more macros, and documented some functions

PokeCodec:
This is a PR using only the changes that were both beneficial AND do not rely on proprietary knowledge


aaaaaa123456789:
This is as poisonous as #1185 and the author has all the credibility of a used car dealer. I'd recommend closing it.

PR #1189, opened 2020-09-13
Undo PokeCodec's PRs and port changes

aaaaaa123456789:
Undo most of PokeCodec's PRs after the debate in chat. Some harmless or completely superseded PRs have been left alone, as there is not much benefit in attempting to undo them.

Thanks to @Kurausukun for porting back matched functions from those PRs.

Reverts #1104, #1108, #1115, #1118, #1119, #1124, #1126, #1127, #1132, #1136, #1137, #1139, #1140, #1144, #1148, #1149, #1150, #1153, #1155, #1177, #1179, #1180, #1181, #1182 and #1183.

Matched functions (from list below):

* [x] _InitContestMonPixels
* [x] AnimFallingFeather
* [x] AnimTask_CreateSurfWave
* [x] AnimTask_CreateSurfWave_Step1
* [x] BattleIntroOpponent1SendsOutMonAnimation
* [x] BerryCrushCommand_BeginNormalPaletteFade
* [x] CheckBagHasSpace
* [x] CheckFocusPunch_ClearVarsBeforeTurnStarts
* [x] Cmd_forcerandomswitch
* [x] Cmd_unknown_24
* [x] CopyGlyphToWindow
* [x] CreateFrontierBrainPokemon
* [x] CreateWaterPulseRingBubbles
* [x] DrawResultsTextWindow
* [x] GetConnectedChildStrength
* [x] GetFrontierOpponentClass
* [x] GetTypeEffectivenessPoints
* [x] GLYPH_COPY
* [x] HandleSendFailure
* [x] HasNoMonsToSwitch
* [x] InitObjectEventStateFromTemplate
* [x] InitPaintingMonOamData
* [x] ModifyStatByNature
* [x] MovementAction_StoreAndLockAnim_Step0
* [x] Phase2_FrontierLogoWave_Func4
* [x] PrintMonWeight
* [x] ReceiveDaycareMailData
* [x] RotateAuroraRingColors_Step
* [x] SetSearchRectHighlight
* [x] sub_806F160
* [x] sub_80C6104
* [x] sub_8181C2C
* [x] sub_81BF2B8
* [x] sub_81C79BC
* [x] sub_81D5F58
* [x] TrendyPhraseIsOld
* [x] TryPutLinkBattleTvShowOnAir
* [x] UpdateSaveAddresses

PR #1191, opened 2020-09-13
Remove unsanitary code, and make final changes

pi1024e:
Please let me know if you have questions, problems, or concerns.

This is an attempt to undo those changes I made under the influence of the code, while keeping documentations and matchings. If I missed anything, please let me know.


GriffinRichards:
This is already being handled by #1189

PR #1192, opened 2020-09-15
Revert Style changes

PokeCodec:
Reverted the unnecessary changes that were style related and had no effect on matching.


GriffinRichards:
This contains about as many style changes as reversions (and leaves most of the reverting out), and still includes essentially unnecessary changes pointed out as having come directly from the source (e.g. those to `src/crt0.s`). Please leave the reversions to the PR already open for this purpose. If there are changes from this PR you'd like to include (the bike macro for example is fine) please make them separate from these unnecessary changes

PR #1194, opened 2020-09-16
Complete TMHM Learnset description

ExpoSeed:
Egg forgot to write the last word when commenting this


PikalaxALT (approved):
You and Egg go together so well, you even complete each other's

PR #1195, opened 2020-09-17
Update Audio Macros

Kurausukun:
Information: Our current audio macros preclude the definition of root key and pan (applicable when used in drumsets, aka "keysplit_all") for GB channels. However, this functionality does exist in m4a--Pokemon Emerald just happens not to use it. However, for the sake of people working with custom music, I believe these values should be modifiable by users.

Additionally, given the recent events that have occurred with this repo, I would like to make something fully clear: I have indeed looked at the GBA SDK before, which contains notes on how these sounds are defined; therefore, I cannot blame you if you reject this PR on the grounds that the information is "tainted." However, this information is readily available and matches the same macros used for directsound--it's hardly information that one would need the source to know.


GriffinRichards:
While you're editing these sound macros it might be nice to flag their arguments with `:req`, especially so that people who have added new uses of them will know they need to be updated

Kurausukun:
Done--all arguments are now required.

PR #1196, opened 2020-09-18
rename instruments (again)

Kurausukun:
This should be the last large overhaul of instrument names--the rest will just be whenever new things get found.

Once again, in the interest of transparency: Some of the presets were found due to looking at the original source (the music files had notes, and some of them included the exact source of the samples--mostly ones that came from the Trinity). However, it's not like the names of these presets are copyrighted or anything, so I don't think there's an issue.


GriffinRichards:
Is the use of `unused` consistent in these? It seems like some kept or added the unused prefix and some dropped it (e.g. `unused_acid_bass` → `trinity_30303_mega_bass`)

Kurausukun:
Some of them were ones I previously thought were unused, but actually turned out to be used in SFX, etc.

GriffinRichards:
Ah alright. Re: "if you want, I can just remove all of the unused prefixes", no that's fine, I was just curious what the reasoning was.

PR #1197, opened 2020-09-18
Changed casts to volatile

MeatLoaf3:
devkitArm wouldn't copy the bits properly in some instances. Casting as volatile solved the issue


PikalaxALT:
Just to clarify, this is to fix an undocumented issue in the `modern` build?

MeatLoaf3:
Yes

PR #1202, opened 2020-09-20
Match sub_819A080

pi1024e:
I, Rose, would like to propose using the match Shoomer wrote for this one unused function so that others can focus on the last nonmatchings. If you want to close without merge, I find no fault in you. This is my only contribution.


aaaaaa123456789:
The problem with this change is that it has an "area of effect". The non-matching is at the end, but _the whole function_ has changed. How much of this is strictly necessary for matching?

PikalaxALT:
There is a strong chance that this conflicts with #1189

pi1024e:
> The problem with this change is that it has an "area of effect". The non-matching is at the end, but _the whole function_ has changed. How much of this is strictly necessary for matching?

It's in the end of a loop, where most changes are anyway

aaaaaa123456789:
And this is why we've been rejecting all your "removals of unsanitary changes" PRs. They all keep many style changes that are not necessary.

pi1024e:
All changes here were necessary.

aaaaaa123456789:
What happens if you remove the cast on ilne 2040?

pi1024e:
Everything is fine.

aaaaaa123456789:
Then that change wasn't strictly necessary, was it? And this is the game I don't want to have to play on every PR.

pi1024e:
Only needed changes are here

pi1024e:
You can’t fire me; I quit this project.

PR #1205, opened 2020-09-25
Use constants for item digits

GriffinRichards:
Items are artificially limited at 99. There's a constant to change this, but people often have trouble tracking down why it's not printing correctly.


PikalaxALT (approved):
Looks good to me. I'll assume this is a comprehensive implemetation?

GriffinRichards:
To the best of my knowledge yeah

PR #1206, opened 2020-09-25
Comments for script command tables 

ExpoSeed:
TODO: update `event.inc` so that script commands unused in emerald are noted


ExpoSeed:
Done. Documenting `cmdD8` and `cmdD9` can be done in a separate PR

PR #1208, opened 2020-09-27
Fix UB

MeatLoaf3:


GriffinRichards:
This isn't the correct fix. The case for the target partner already exists, and correctly checks the move history rather than the moves directly

PR #1209, opened 2020-09-27
Set NULL for modern compilers

MeatLoaf3:
Freeing memory is not enough. On modern compilers, the pointer must be set to NULL too.


aaaaaa123456789:
What's the point of this? That null is a dead store.

PR #1211, opened 2020-09-28
[Leak-informed] Resolve fakematch CopyGlyphToWindow.

ProjectRevoTPP:
The function in the OG source was used to discover the required fixes for this function. I did my best to remove any unnecessary changes, although I do have a few disclaimers:

1] While not specifically mentioning if Game Freak did or didn't use an inline, it looks better as one due to usage of temporaries. I tried to test a macro but I can't deal with lvalue errors, so here we are.
2] In addition to an inline possibly being required, there are at least 2 in-condition assignment of variables being required to match. Reverting any of these causes an error because it doesn't match.
3] Doing the u32* array instead of the struct members works while the struct doesn't. The reason is using the struct requires a local array assignment in order to iterate over the loop, and regalloc does not match if you do this. Suck it, Rose.

Discord:
Revo#7090

### DIFF
--- a/gflib/text.c
+++ b/gflib/text.c
@@ -462,40 +462,37 @@ u8 GetLastTextColor(u8 colorType)
     }
 }
 
-#define GLYPH_COPY(fromY_, toY_, fromX_, toX_, unk)                                                               \
-    {                                                                                                             \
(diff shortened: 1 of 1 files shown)


PikalaxALT:
The initial comment does not use the template, which is now required for all PRs going forward.

PR #1212, opened 2020-09-28
[Leak informed] Resolve fakematch CopyGlyphToWindow.

ProjectRevoTPP:
## Description
The function in the OG source was used to discover the required fixes for this function. I did my best to remove any unnecessary changes, although I do have a few disclaimers:

1] While not specifically mentioning if Game Freak did or didn't use an inline, it looks better as one due to usage of temporaries. I tried to test a macro but I can't deal with lvalue errors, so here we are.
2] In addition to an inline possibly being required, there are at least 2 in-condition assignment of variables being required to match. Reverting any of these causes an error because it doesn't match.
3] Doing the u32* array instead of the struct members works while the struct doesn't. The reason is using the struct requires a local array assignment in order to iterate over the loop, and regalloc does not match if you do this. Suck it, Rose.

## Type of changes

- [X] Decompilation (matching, fixing nonmatching, fakematching, etc.)
- [x] Documentation (naming symbols, commenting, etc.)
- [ ] Style (code style refactors, typo, etc.)
- [ ] Other: 

## Checklist

- [X] I am a member of the [pret Discord server](https://discord.gg/d5dubZ3).
- [X] `make compare` and `make compare modern` on my local machine outputs OK.
- [X] My code follows the code style of this project.
- [X] If I am fixing a bug or undefined behavior in the modern build, I have documented the bug and tested the fix locally.
- [X] All my usage, if any, of the leaked source code has been disclosed in pret's server.

## **Discord contact info**
Revo#7090

### DIFF
--- a/gflib/text.c
+++ b/gflib/text.c
@@ -464,7 +464,6 @@ u8 GetLastTextColor(u8 colorType)
 
 inline static void GLYPH_COPY(u8 *windowTiles, u32 widthOffset, u32 j, u32 i, u32 *ptr, s32 width, s32 height)                                           //
 {
-    // WHYYYYYYYYY is dummyX needed aaaaaaaaaaaaaaaaaaaaa
     u32 xAdd, yAdd, r5, bits, toOrr, dummyX;
     u8 *dst;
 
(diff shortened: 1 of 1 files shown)


gflib/text.c
```diff
@@ -462,104 +462,88 @@ u8 GetLastTextColor(u8 colorType)
...
-        }                                                                                                         \
+inline static void GLYPH_COPY(u8 *windowTiles, u32 widthOffset, u32 j, u32 i, u32 *ptr, s32 width, s32 height)                                           //
+{
+    // WHYYYYYYYYY is dummyX needed aaaaaaaaaaaaaaaaaaaaa
+    u32 xAdd, yAdd, r5, bits, toOrr, dummyX;
+    u8 *dst;
+
+    xAdd = j + width;
+    yAdd = i + height;
+    dummyX = j; // :masuda:
```
PikalaxALT, line 473: discord emotes don't render in c comments

PikalaxALT (approved):

PR #1216, opened 2020-10-01
Replace POKEMON_SLOTS_NUMBER

GriffinRichards:
`POKEMON_SLOTS_NUMBER` is an easy to miss constant that just duplicates `NUM_SPECIES`. If people want to differentiate the two they still can in `DEX_FLAGS_NO`, which was its only usage

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -2,7 +2,6 @@
 #define GUARD_DATA_H
 
 #include "constants/moves.h"
-#include "constants/species.h"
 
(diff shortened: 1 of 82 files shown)


PikalaxALT:
The required form is missing. Can't allow any exceptions, now, can we?

huderlem:
Yes, we obviously can.  Especially since the PR template was ham-fisted into the repo.

PR #1217, opened 2020-10-02
rewrite MultiBootWaitCycles as naked function

ipatix:
## Description

Before, when compiling MultiBootWaitCycles with O3 and MODERN=1, you
might have run into problems during optimizations when the compiler tried
to optimize the function, even if declared NOINLINE.

When rewriting this function as NAKED function, this no longer happens
as the optimizer will treat it as black box and compilation will resume.

## Type of changes

- [ ] Decompilation (matching, fixing nonmatching, fakematching, etc.)
- [ ] Documentation (naming symbols, commenting, etc.)
- [ ] Style (code style refactors, typo, etc.)
- [x] Other: Improve compatibility with modern compilers.

## Checklist

- [x] I am a member of the [pret Discord server](https://discord.gg/d5dubZ3).
- [x] `make compare` and `make compare modern` on my local machine outputs OK.
- [x] My code follows the code style of this project.
- [ ] If I am fixing a bug or undefined behavior in the modern build, I have documented the bug and tested the fix locally.
- [ ] All my usage, if any, of the leaked source code has been disclosed in pret's server.

## **Discord contact info**

iTrash#7602

### DIFF
--- a/src/multiboot.c
+++ b/src/multiboot.c
@@ -1,3 +1,4 @@
+#include "global.h"
 #include "gba/gba.h"
 #include "multiboot.h"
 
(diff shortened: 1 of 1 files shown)


MeatLoaf3:
What is wrong with the optimization of multiboot, if it does the same opertaions?

SBird1337:
the current implementation will fail on an O3 modern compilation

ipatix:
The error that's shown is that a definition of `MultiBootWaitCyclesLoop` already exists, which is most likely due to either the optimizer trying to inline the function or it might duplicate the function due to other analyzation passes.

PikalaxALT (approved):
Approved pending local test of the changes as per the checklist.

ipatix:
So what kind of local testing still needs to be done?

PR #1227, opened 2020-10-07
Document gUnknown_03006298

kageru:
Just a simple `sed` replace I did while working on a romhack and wanted to contribute upstream.
`make [modern] compare` runs fine locally.
 **Discord contact info**
I’m on the pret server. kageru#7213.

### DIFF
--- a/src/battle_tower.c
+++ b/src/battle_tower.c
@@ -46,7 +46,7 @@ EWRAM_DATA const struct BattleFrontierTrainer *gFacilityTrainers = NULL;
 EWRAM_DATA const struct FacilityMon *gFacilityTrainerMons = NULL;
 
 // IWRAM common
-u16 gUnknown_03006298[MAX_FRONTIER_PARTY_SIZE];
+u16 gFrontierTempParty[MAX_FRONTIER_PARTY_SIZE];
 
 // This file's functions.
 static void InitTowerChallenge(void);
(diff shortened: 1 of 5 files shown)


src/battle_tower.c
```diff
@@ -2224,11 +2224,11 @@ static void GetApprenticeMultiPartnerParty(u16 trainerId)
...
     }
 
-    gUnknown_03006298[0] = validSpecies[Random() % count];
+    gFactoryOpponentTeam[0] = validSpecies[Random() % count];
     do
     {
-        gUnknown_03006298[1] = validSpecies[Random() % count];
-    } while (gUnknown_03006298[0] == gUnknown_03006298[1]);
+        gFactoryOpponentTeam[1] = validSpecies[Random() % count];
+    } while (gFactoryOpponentTeam[0] == gFactoryOpponentTeam[1]);
```
GriffinRichards, line 2231: `FactoryOpponent` might be misleading, considering that it's used here for a non-opponent outside the factory

kageru: Hm, true. It’s more like the team of an NPC in various battle facilities (just mainly factory). How would you call that? `gNpcFrontierTeam`?

kageru: I renamed it to `gFrontierNpcTeam`. Let me know if that’s better.

GriffinRichards: What about something like `gFrontierTempParty`? In any case we tend to use "party" over "team"

kageru: Sounds good to me.

PR #1231, opened 2020-10-16
Add comments noting dead code

MeatLoaf3:
Hey all, Meatloaf here!

asm("svc 2") will never execute due to the above statement already handling this case, so that has been noted. Additionally, the case 0: statement is redundant due to the if statement already handling that case.

Changed !ctx->scriptPtr to ctx->scriptPtr == NULL in script.c, line 84
## Description

I only added comments and changed a single statement to reflect this
## **Discord contact info**

MEATLOAF#4302

### DIFF
--- a/src/script.c
+++ b/src/script.c
@@ -60,7 +60,7 @@ void StopScript(struct ScriptContext *ctx)
 
 bool8 RunScriptCommand(struct ScriptContext *ctx)
 {
-    if (ctx->mode == 0) 
+    if (ctx->mode == 0)
         return FALSE;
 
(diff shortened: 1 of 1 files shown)


GriffinRichards:
This seems unnecessary. There's a lot of dead code in the repo, I don't think we need comments on every instance of it. Especially when it's already clear because the redundant statement is only a few lines away

MeatLoaf3:
You actually do document dead code though. See: //unused

GriffinRichards:
Marking an entire function as unused seems more reasonable than marking every redundant case and conditional

PR #1232, opened 2020-10-16
[LEAK INFORMED] Match FadeOutBody

Kurausukun:
## Description
Matched by looking at the logic used in not the emerald leak but the agb bootrom leak, which contains an uncompiled (albeit very old) version of m4a.

Full credit for discovering the matching goes to [atasro2](https://github.com/atasro2).

## **Discord contact info**
Kurausukun#9923

### DIFF
--- a/src/m4a.c
+++ b/src/m4a.c
@@ -659,7 +659,6 @@ void FadeOutBody(struct MusicPlayerInfo *mplayInfo)
             mplayInfo->fadeOI = 0;
         }
     }
-
     else
     {
         if ((s16)(mplayInfo->fadeOV -= (4 << FADE_VOL_SHIFT)) <= 0)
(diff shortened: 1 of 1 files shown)


src/m4a.c
```diff
@@ -643,40 +643,27 @@ void FadeOutBody(struct MusicPlayerInfo *mplayInfo)
...
-        mplayInfo->fadeOV += (4 << FADE_VOL_SHIFT);
-
-        if ((u16)(mplayInfo->fadeOV & mask) >= (64 << FADE_VOL_SHIFT))
+        if ((u16)(mplayInfo->fadeOV += (4 << FADE_VOL_SHIFT)) >= (64 << FADE_VOL_SHIFT))
         {
             mplayInfo->fadeOV = (64 << FADE_VOL_SHIFT);
             mplayInfo->fadeOI = 0;
         }
     }
+
```
GriffinRichards: Unnecessary newline

PR #1234, opened 2020-10-19
[LEAK INFORMED] The Bug that Never Was; remove BUG: comment, as the t…

MeatLoaf3:
…emplate is of no array

It turns out the template is standalone. Whether or not this was a good design choice or poor is different entirely, or whether the correct method or not was called is another issue entirely. Nevertheless, template is not an array.

## Description

Changed sWindowTemplates from an array to a singular variable
Now we are sending the address of the struct instead of the array containing it

## **Discord contact info**
MEATLOAF#4302

### DIFF
--- a/src/roulette.c
+++ b/src/roulette.c
@@ -454,7 +454,7 @@ static const struct BgTemplate sBgTemplates[] =
 };
 static const struct WindowTemplate sWindowTemplate =
 {
-      .bg = 0;
+      .bg = 0,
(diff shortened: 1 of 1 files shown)


GriffinRichards:
This is worse. The bug still exists, and now requires more changes to fix. `InitWindows` loops until `bg` is `0xFF` or after 32 attempts, changing this from an array to the address of a single struct (which are equivalent) doesn't affect this

PR #1236, opened 2020-10-21
[STYLE] Change "inline static void" to "static inline void"

MeatLoaf3:
I propose "static inline void" because EVERY OTHER inline function in pret's codebase has "static inline void", not "inline static void". When I did a search for "static inline" every other function with static inline came up, except the GLYPH_COPY one. GLYPH_COPY is the only function to use "inline static", which is stylistically inconsistent.

## Description

Swap the words "Static" and "inline"
## **Discord contact info**
MEATLOAF#4302

### DIFF
--- a/gflib/text.c
+++ b/gflib/text.c
@@ -462,7 +462,7 @@ u8 GetLastTextColor(u8 colorType)
     }
 }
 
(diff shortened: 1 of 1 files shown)


MeatLoaf3:
Updated comment

MeatLoaf3:
I changed the PR to reflect this is a stylistic change, unrelated to the leaks

PikalaxALT:
Requester is blacklisted

PR #1239, opened 2020-10-23
[BUG-FIX COMMENT] Comment in bug fix for match_call records

MeatLoaf3:
The issue isn't that the cases are looking at the wrong data, but the wrong case labels were used. This is because in the original code, the enum values for pike and factory were accidentally swapped. This doesn't show up in pret, but is the reason why the bug happens. Switching the case values is the fix to this issue.

In commented out lines, fixed the cases in match_call.c
## **Discord contact info**
MEATLOAF#4302

### DIFF
--- a/src/match_call.c
+++ b/src/match_call.c
@@ -1847,7 +1847,7 @@ static u16 GetFrontierStreakInfo(u16 facilityId, u32 *topicTextId)
         }
         *topicTextId = 2;
         break;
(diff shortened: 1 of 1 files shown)


MeatLoaf3:
The fixes are commented in, so that the build can still match. The fix is just for those who want to make their own builds.

PikalaxALT:
Requester is blacklisted

PR #1245, opened 2020-11-04
Fix battle factory on gcc 10

### LINKED ISSUE #1243: Battle factory softlocks the game when building with modern
As the title says, when building the current master with `make modern`, the game softlocks when entering Pokemon selection in the battle factory. The music keeps playing, and menu sounds are still being played whenever A or B is pressed, but the game doesn’t respond to inputs beyond that.
The selection menu is also visibly broken, see https://i.kageru.moe/k36PRH.png
When building without `modern`, everything works as intended.

To quickly get a fresh save to factory to reproduce the issue, you can modify littleroot as follows:

```diff
--- a/data/maps/LittlerootTown/scripts.inc
+++ b/data/maps/LittlerootTown/scripts.inc
@@ -407,6 +407,8 @@ LittlerootTown_EventScript_GoSaveBirchTrigger:: @ 81E811F

 LittlerootTown_EventScript_TownSign:: @ 81E8151
 	msgbox LittlerootTown_Text_TownSign, MSGBOX_SIGN
+        warpteleport MAP_BATTLE_FRONTIER_BATTLE_FACTORY_LOBBY, 255, 9, 10
+        waitstate
 	end

 LittlerootTown_EventScript_BirchsLabSign:: @ 81E815A
```

Talking to the town sign in littleroot will then warp you to the factory.

For questions, I’m kageru#7213 on pret discord.

Edit: Tested on mgba 0.8.3 and built with gcc 10.2.0 from devkitARM.

GriffinRichards:
Fixes #1243

### DIFF
--- a/src/battle_factory_screen.c
+++ b/src/battle_factory_screen.c
@@ -1145,7 +1145,11 @@ static void CB2_InitSelectScreen(void)
         LoadPalette(gFrontierFactorySelectMenu_Pal, 0, 0x40);
         LoadPalette(gUnknown_0861046C, 0xF0, 8);
         LoadPalette(gUnknown_0861046C, 0xE0, 10);
+        #if MODERN
+        if (sFactorySelectScreen && sFactorySelectScreen->fromSummaryScreen)
+        #else
         if (sFactorySelectScreen->fromSummaryScreen == TRUE)
+        #endif
             gPlttBufferUnfaded[228] = sFactorySelectScreen->unk2A4;
         LoadPalette(gUnknown_0861039C, 0x20, 4);
         gMain.state++;
@@ -1167,7 +1171,11 @@ static void CB2_InitSelectScreen(void)
         SetVBlankCallback(Select_VblankCb);
         BeginNormalPaletteFade(0xFFFFFFFF, 0, 0x10, 0, RGB_BLACK);
         SetGpuReg(REG_OFFSET_DISPCNT, DISPCNT_OBJ_ON | DISPCNT_BG0_ON | DISPCNT_BG1_ON | DISPCNT_OBJ_1D_MAP);
+        #if MODERN
+        if (sFactorySelectScreen && sFactorySelectScreen->fromSummaryScreen)
+        #else
         if (sFactorySelectScreen->fromSummaryScreen == TRUE)
+        #endif
         {
             Select_SetWinRegs(88, 152, 32, 96);
             ShowBg(3);
@@ -1181,7 +1189,11 @@ static void CB2_InitSelectScreen(void)
         gMain.state++;
         break;
     case 5:
+        #if MODERN
+        if (sFactorySelectScreen && sFactorySelectScreen->fromSummaryScreen)
+        #else
         if (sFactorySelectScreen->fromSummaryScreen == TRUE)
+        #endif
             sFactorySelectScreen->cursorPos = gLastViewedMonIndex;
         Select_InitMonsData();
         Select_InitAllSprites();
@@ -3170,7 +3182,11 @@ static void CB2_InitSwapScreen(void)
         gMain.state++;
         break;
     case 5:
+        #if MODERN
+        if (sFactorySwapScreen && sFactorySwapScreen->fromSummaryScreen)
+        #else
         if (sFactorySwapScreen->fromSummaryScreen == TRUE)
+        #endif
             sFactorySwapScreen->cursorPos = gLastViewedMonIndex;
         gMain.state++;
         break;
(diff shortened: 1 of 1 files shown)


src/battle_factory_screen.c
```diff
@@ -1241,7 +1229,9 @@ static void CB2_InitSelectScreen(void)
     }
 }
 
-static void Select_InitMonsData(void)
+// This should be called at the top of CB2_InitSelectScreen
+// sFactorySelectScreen is dereferenced prior to being allocated otherwise
+static NOINLINE void Select_InitMonsData(void)
```
jiangzhengwenjz, line 1234: don't think it's a good fix becase UB is still there. Usually in modern compilers, if a pointer has been dereferenced, it means it's safe to remove NULL check afterwards in optimization (because if it's really NULL, then it's UB and compiler can generate any code; otherwise it's safe)
But this is never guaranteed to work.

aaaaaa123456789: This obviously isn't a fix; it's a hack. It should probably be marked as such.
(For what it's worth, passing `-fno-delete-null-pointer-checks` to the compiler would achieve the same thing in a much safer way.)

GriffinRichards: I'm happy to stick with my original solution. The current one was suggested by @PikalaxALT, if they feel differently

aaaaaa123456789: If you're talking about the fix from c59a81846262, then yes, that looks like a proper fix.

kageru: I think adding `-fno-delete-null-pointer-checks` to the compiler flags should be done regardless of which solution we pick here. The optimization is based on an assumption that simply isn’t true for the GBA, so I think we should disable it to prevent bugs like this in the future.

aaaaaa123456789: The assumption is completely true for the GBA, as address 0 isn't readable by non-BIOS code. And even if it was, the value contained there is completely useless for all user code that isn't explicitly trying to test something about GBA hardware/emulation (which is something that would normally be done in straight-up assembly, anyway).
Dereferencing a null pointer in C code for the GBA is an error 100 out of 100 times.

kageru: My understanding was that the value is useless but it doesn’t segfault (or generally crash) the way it does on most other hardware. Is that wrong?

aaaaaa123456789: There's not such thing as a segmentation fault in the GBA, but that's not the issue with null pointers. The null pointer assumption is "a null pointer is used as a sentinel value and it doesn't point to anything". Which is as true in the GBA as in any other modern platform; the fact that there is actually an interrupt vector at address 0 (which only the hardware can access) doesn't change this.
Contrast this with the GBC, where there's a specific instruction (opcode 0xC7) to call address 0, and where writes to address 0 are often used to control banking hardware — in this case address 0 is accessed intentionally.

kageru: `man gcc` says
```
-fdelete-null-pointer-checks
    Assume that programs cannot safely dereference null pointers, and that no code or data
    element resides at address zero.  This option enables simple constant folding
    optimizations at all optimization levels.  In addition, other optimization passes in GCC
    use this flag to control global dataflow analyses that eliminate useless checks for null
    pointers; these assume that a memory access to address zero always results in a trap, so
    that if a pointer is checked after it has already been dereferenced, it cannot be null.
```
which means it’s not allowed to point to anything *and* reading should trigger some kind of exception somewhere that interrupts the program. If I understand https://mgba.io/2020/01/25/infinite-loop-holy-grail/ right, the latter is not necessarily given.

aaaaaa123456789: Dereferencing a null pointer is undefined behavior. A program that does so is wholly invalid.

That being said, what that optimization flag is saying is, null pointer checks can be eliminated from _all_ programs if attempting to access address 0 stops execution. You're correct in that this never happens in the GBA — accessing memory will always result in a value, meaningful or not.
But correct C programs cannot dereference null pointers anyway. With correct code, this optimization is always valid, because that's something that is invalid anyway. And that's why it's enabled by default.

Any C program attempting to access address zero has a bug. The obvious exceptions are programs that _intentionally_ access address zero for some purpose, knowing that they run in an environment where that is valid and meaningful. That environment is not the GBA — writing to 0 does nothing, and reading from 0 reads garbage. And thus, this flag is perfectly valid for correct code.
Of course, the problem here is that what Game Freak wrote is _not_ correct code, and thus there's a bug when a more recent compiler attempts to optimize it out. They would've probably caught it themselves if they had used GCC 10, but GCC 10 didn't exist back then.

kageru: I was never trying to say that a null dereference can be correct code. Sorry if I didn’t make that clear enough.
It’s just that, as you said, GameFreak’s code isn’t correct code, so disabling that optimization can prevent similar issues for bugs that we haven’t discovered yet (and that gcc might not optimize *yet*).

aaaaaa123456789: If leaving the flag off helps finding bugs, why would you want to turn it on? Concealing bugs is never the right approach.

PR #1246, opened 2020-11-06
Simplify RAM reset adaptation for modern

PikalaxALT:
## Description

When compiling with newer versions of gcc, AgbMain preserves more registers than when compiled with agbcc. This results in the vanilla RegisterRamReset call crashing, due to the return address on the stack being overwritten with 0. The current fix injects a manual loop into AgbMain to clear IWRAM. The proposed solution instead moves the RegisterRamReset to crt0 when compiling with modern gcc. It's more efficient and less of an eyesore this way.

## **Discord contact info**

PikalaxALT#5823

### DIFF
--- a/src/crt0.s
+++ b/src/crt0.s
@@ -93,7 +93,7 @@ Init: @ 8000204
 	str r0, [r1]
 	.if MODERN
 	mov r0, #255 @ RESET_ALL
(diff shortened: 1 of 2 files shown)


Sierraffinity:
Looks good to me.

PR #1251, opened 2020-11-15
Document bugged IV inheritance difference

ExpoSeed:
Documented more of GameFreak's stupidity.

ExpoSeed#5613


GriffinRichards:
As a side note we should really add a bugfix target and put these in conditionals rather than commenting them with instructions. For some reason I thought we already had this

PR #1252, opened 2020-11-16
Missing inc files in asm/*.s and data/*.s will raise an actual error message.

luckytyphlosion:
This was supposed to include more changes to account for missing *.h files, but I haven't been able to work on that. Given that this error is very nasty and anti-user friendly, I've decided to make a PR for this now.

This fix generally follows the same principle used in the fix in pokecrystal, found [here](https://github.com/pret/pokecrystal/commit/ea8aaeb8e3a85fa2303fe52493d840de9222f3a2). [This commit](https://github.com/pkmncoraldev/polishedcoral/commit/d064efa02c2e8eb0b38d12968f7b125617e724c4#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52), which displays the changes for a more modern pokecrystal, was also referenced.

## **Discord contact info**
luckytyphlosion#1166

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -300,7 +300,6 @@ $(DATA_ASM_BUILDDIR)/%.o: $(DATA_ASM_SUBDIR)/%.s
 else
 define DATA_ASM_DEP
 $1: $2 $$(shell $(SCANINC) -I include -I "" $2)
-	#echo "hello"
 	$$(PREPROC) $$< charmap.txt | $$(CPP) -I include | $$(AS) $$(ASFLAGS) -o $$@
 endef
 $(foreach src, $(REGULAR_DATA_ASM_SRCS), $(eval $(call DATA_ASM_DEP,$(patsubst $(DATA_ASM_SUBDIR)/%.s,$(DATA_ASM_BUILDDIR)/%.o, $(src)),$(src))))
--- a/Makefile
+++ b/Makefile
@@ -137,8 +137,7 @@ ASM_SRCS := $(wildcard $(ASM_SUBDIR)/*.s)
 ASM_OBJS := $(patsubst $(ASM_SUBDIR)/%.s,$(ASM_BUILDDIR)/%.o,$(ASM_SRCS))
 
 # get all the data/*.s files EXCEPT the ones with specific rules
-REGULAR_DATA_ASM_SRCS := $(wildcard $(DATA_ASM_SUBDIR)/*.s)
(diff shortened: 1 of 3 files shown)


luckytyphlosion:
Also ideally the error message would print out where the bad include is, but I would need to figure out how to do that.

GriffinRichards:
Nice. Not having the location is a minor problem given these files are often included into a larger parent file (e.g. `event_scripts.s` or `tilesets.s`) but showing the problematic include path is significantly better. Is there a roadblock for doing this with the C files or have you just not touched that yet

Makefile
```diff
@@ -136,6 +136,10 @@ C_ASM_OBJS := $(patsubst $(C_SUBDIR)/%.s,$(C_BUILDDIR)/%.o,$(C_ASM_SRCS))
 ASM_SRCS := $(wildcard $(ASM_SUBDIR)/*.s)
 ASM_OBJS := $(patsubst $(ASM_SUBDIR)/%.s,$(ASM_BUILDDIR)/%.o,$(ASM_SRCS))
 
+# get all the data/*.s files EXCEPT the ones with specific rules
+REGULAR_DATA_ASM_SRCS := $(wildcard $(DATA_ASM_SUBDIR)/*.s)
+REGULAR_DATA_ASM_SRCS := $(filter-out $(DATA_ASM_SUBDIR)/maps.s $(DATA_ASM_SUBDIR)/map_events.s, $(REGULAR_DATA_ASM_SRCS))
```
GriffinRichards, line 141: Also no need for this double assignment

luckytyphlosion: Fixed.

luckytyphlosion:
> Is there a roadblock for doing this with the C files or have you just not touched that yet
Just haven't touched it yet. The fix isn't one-to-one between inc and h files so I didn't touch it yet.

Makefile
```diff
@@ -275,23 +278,33 @@ endif
...
 endif
 
-$(DATA_ASM_BUILDDIR)/%.o: $(DATA_ASM_SUBDIR)/%.s $$(data_dep)
+ifeq ($(NODEP),1)
+$(DATA_ASM_BUILDDIR)/%.o: $(DATA_ASM_SUBDIR)/%.s
 	$(PREPROC) $< charmap.txt | $(CPP) -I include | $(AS) $(ASFLAGS) -o $@
+else
+define DATA_ASM_DEP
+$1: $2 $$(shell $(SCANINC) -I include -I "" $2)
+	#echo "hello"
```
GriffinRichards, line 303: Forgot to remove this

luckytyphlosion: Fixed.

GriffinRichards (approved):

PR #1254, opened 2020-11-17
Use more constants in Battle Tower code

kageru:
Introduced a new constant for the number of tower records stored (via record mixing) and used that in the relevant loops.
Also use the party size constants in some more places where they should be used.

kageru#7213 on pret discord.

### DIFF
--- a/src/battle_tower.c
+++ b/src/battle_tower.c
@@ -999,7 +999,7 @@ static bool8 ChooseSpecialBattleTowerTrainer(void)
         return FALSE;
 
     winStreak = GetCurrentBattleTowerWinStreak(lvlMode, battleMode);
-    for (i = 0; i < 5; i++)
+    for (i = 0; i < BATTLE_TOWER_RECORD_COUNT; i++)
     {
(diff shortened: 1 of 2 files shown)


src/battle_tower.c
```diff
@@ -1324,15 +1324,15 @@ void PutNewBattleTowerRecord(struct EmeraldBattleTowerRecord *newRecordEm)
...
+    for (i = 0; i < BATTLE_TOWER_RECORD_COUNT; i++)
     {
         k = 0;
         for (j = 0; j < TRAINER_ID_LENGTH; j++)
         {
             if (gSaveBlock2Ptr->frontier.towerRecords[i].trainerId[j] != newRecord->trainerId[j])
                 break;
         }
-        if (j == 4)
+        if (j == BATTLE_TOWER_RECORD_COUNT - 1)
```
GriffinRichards, line 1335: `j` is looping over `TRAINER_ID_LENGTH`, not `BATTLE_TOWER_RECORD_COUNT`

kageru: You’re right, my bad. I thought this was an operation specifically on the last iteration of the loop. I’ll fix that.

PR #1255, opened 2020-11-23
Name object event palettes and tags

GriffinRichards:
e.g. `OBJ_EVENT_PAL_TAG_8` --> `OBJ_EVENT_PAL_TAG_BRENDAN` and `gObjectEventPalette17` --> `gObjectEventPal_May`


ShinyDragonHunter:
I feel for NPC palettes, they should be named with the colour in the palette. So if a palette has grey, blue and pink, it’d be named `gObjectEventPaletteGreyBluePink`. That way, it’d be easier for a person to tell them apart.

GriffinRichards:
> I feel for NPC palettes, they should be named with the colour in the palette. So if a palette has grey, blue and pink, it’d be named `gObjectEventPaletteGreyBluePink`. That way, it’d be easier for a person to tell them apart.

They each have 9 non-standard colors and they overlap. There's not a reasonable way to do that. For example npc palette 1 has various shades of purple and blue, but npc palette 4 does as well (and they're not the same shades).

If people want to see all the colors they can simply open the palette

ShinyDragonHunter:
That’s why you’d be labelling them with each colour. Some may have the same colours but with different shades but the person would still be able to tell them about because I doubt any palette has 3 colours the same but with different shades

GriffinRichards:
Still sounds like a cumbersome and potentially misleading generalization

ShinyDragonHunter:
Well, it’s similar to how pokecrystal named NPC palettes. But it’s entirely your call, it was honestly just a suggestion ☺️

PR #1187, opened 2020-09-13
fakematch fixes, begin documenting Berry Crush

kazblox:

PR #1265, opened 2020-12-11
fakematch fixes, begin documenting Berry Crush (based on #1187)

Kurausukun:
## Description
Credit to Kaz (kazblox) for the PR this was based on, I just merged it with master and fixed conflicts.

## **Discord contact info**
Kurausukun#9923


src/menu_specialized.c
```diff
@@ -880,39 +880,34 @@ s32 GetBoxOrPartyMonData(u16 boxId, u16 monId, s32 request, u8 *dst)
...
-            boxMon = GetBoxedMonPtr(boxId, monId);
-            gender = GetBoxMonGender(boxMon);
-            level = GetLevelFromBoxMonExp(boxMon);
-        }
+        boxMon = GetBoxedMonPtr(box, mon);
+        gender = GetBoxMonGender(boxMon);
+        level = GetLevelFromBoxMonExp(boxMon);
+    }
 
         if ((species == SPECIES_NIDORAN_F || species == SPECIES_NIDORAN_M) && !StringCompare(dst, gSpeciesNames[species]))
```
GriffinRichards, line 912: Fix indentation

src/overworld.c
```diff
@@ -2908,7 +2908,11 @@ static void ZeroObjectEvent(struct ObjectEvent *objEvent)
     memset(objEvent, 0, sizeof(struct ObjectEvent));
 }
 
-static void SpawnLinkPlayerObjectEvent(u8 linkPlayerId, s16 x, s16 y, u8 a4)
+#define linkGender(obj) obj->singleMovementActive
```
GriffinRichards, line 2911: Why is something called `linkGender` being used to access a field called `singleMovementActive`?

src/overworld.c
```diff
@@ -2923,8 +2927,8 @@ static void SpawnLinkPlayerObjectEvent(u8 linkPlayerId, s16 x, s16 y, u8 a4)
     linkPlayerObjEvent->movementMode = MOVEMENT_MODE_FREE;
 
     objEvent->active = 1;
-    objEvent->singleMovementActive = a4;
-    objEvent->range.as_byte = 2;
+    linkGender(objEvent) = gender;
+    linkDirection(objEvent) = 2;
```
GriffinRichards, line 2931: Is this a direction or a range? If it's a direction it should use the `DIR` constants

src/overworld.c
```diff
@@ -2947,7 +2951,7 @@ static void sub_80877DC(u8 linkPlayerId, u8 a2)
     {
         u8 objEventId = gLinkPlayerObjectEvents[linkPlayerId].objEventId;
         struct ObjectEvent *objEvent = &gObjectEvents[objEventId];
-        objEvent->range.as_byte = a2;
+        linkDirection(objEvent) = a2;
```
GriffinRichards, line 2954: Let's name `a2` while we're here

GriffinRichards (changes requested):

kazblox:
GriffinRichards: **The game reuses the direction and range variables as gender and direction variables when in Link Mode.** Hence the names linkGender and linkDirection.

GriffinRichards:
Does it not match with a union? If not please put a comment at the define that the field is being re-used, it's confusing at a glance

kazblox:
It could, but it would create useless cruft just to make space for two miniscule members, especially when they're far away from each other, which means I would have to create two unions of "link1" and "link2." It's better to just leave them as defines, maybe I should move them to the event object header itself so that people don't get lost.

Or should I introduce the unions anyway? How would this benefit?

GriffinRichards:
Just a comment stating that they're being re-used is fine then

GriffinRichards (approved):

PR #1267, opened 2020-12-11
refactor m4a.c/m4a_1.s constant/variable/type naming

ipatix:
## Description
Intentionally I just wanted to go through `CgbSound` and try to understand it. I was kind of bothered of how the code looked so I tried to make it more readable. I took the chance to go over some more of the m4a code to fix variable and constant names.


include/gba/m4a_internal.h
```diff
@@ -168,22 +195,22 @@ struct SoundInfo
     u8 freq;
 
     u8 mode;
-    u8 c15;
+    u8 c15;          // periodically counts from 14 downto 0 (15 states)
```
GriffinRichards, line 198: Typo, `downto`

include/gba/m4a_internal.h
```diff
@@ -270,8 +297,8 @@ struct MusicPlayerTrack
     u8 lfoDelay;
     u8 lfoDelayC;
     u8 priority;
-    u8 echoVolume;
-    u8 echoLength;
+    u8 pseudoEchoVolume;
+    u8 pseudoEchoLength;
```
GriffinRichards, line 301: What is the distinction here between `echoVolume` and `pseudoEchoVolume`?

src/m4a.c
```diff
@@ -963,243 +967,233 @@ void CgbSound(void)
...
                     *nrx0ptr = 0;
-                    *nrx1ptr = channels->le;
-                    if (channels->le)
-                        channels->n4 = -64;
+                    *nrx1ptr = channels->length;
+                    if (channels->length)
+                        channels->n4 = 0xC0;
                     else
-                        channels->n4 = -128;
+                        channels->n4 = 0x80;
```
GriffinRichards, line 1003: Is there a reason for these hex value changes? Should these be constants?

ipatix: Making a define for those would indeed be nicer. I wrote it in hex since it's a bitmap written to a register and thus is easier to read.

include/gba/m4a_internal.h
ipatix: I think that is the official terminus

ipatix: I think that is the official terminology from the SDK and this effect isn't a real echo, so pseudo echo seems more appropriate to me.

include/gba/m4a_internal.h
ipatix: oops, I probably wrote too much vhdl. Gonna get fixed..

src/m4a.c
```diff
@@ -1161,43 +1155,40 @@ void CgbSound(void)
...
-            {
                 *nrx3ptr = (*nrx3ptr & 0x08) | channels->frequency;
-            }
             channels->n4 = (channels->n4 & 0xC0) + (*((u8*)(&channels->frequency) + 1));
             *nrx4ptr = (s8)(channels->n4 & mask);
         }
 
-        if (channels->modify & 1)
+        /* 4. apply envelope & volume to HW registers */
+        if (channels->modify & CGB_CHANNEL_MO_VOL  )
```
ipatix, line 1191: Oops, I seem to have accidently pasted a few spaces too much.

PR #1273, opened 2020-12-15
Introduce NUM_ABILITY_SLOTS

ExpoSeed:
may seem trivial, but I was trying to refactor `abilityHidden` in PE to be an array of length 3 rather than a length 2 array + some other field, and this constant is a necessary prerequisite


GriffinRichards:
I'd rather not start introducing unused constants with the intention of someone using it for a theoretical change. This doesn't do anything in the repo, and if someone needs this it would take 10 seconds at best to add it

ExpoSeed:
It's not so simple with the branches and their bevy of merge conflicts

The constraint is that BE needs it to be u16 while PE needs it to be length 3, and there's no way to cleanly do both as is without this constant

Plus, magic numbers should be constants in general

GriffinRichards:
To restate from Discord: 
This isn't a magic number. It's the size of a single array, and its name makes what the size is self-evident. This array isn't looped over. All this does is move where you need to change the size of the array. If you need this constant in your branches to help with merge conflicts between them, then add it there. It doesn't add anything of value here.

ExpoSeed:
a few things;

fair point about not being looped over, but i don't think the size of `abilities` is any old array, it's a part of mon data that should be grouped with the rest of the constants like it for consistency. introducing the constant does have a bit of value to it, and i'd rather have a meaningful constant defined somewhere with other constants than just a random 2 popping up

and as i explained in discord, introducing the constant in egg branches just moves the conflict elsewhere, which gets us nowhere

i agree it's not much value but i don't think it's entirely worthless either

alternatively the collection of constants could be moved somewhere more sensible. stuff like `MAX_MON_MOVES` and `NUM_STATS` being in `global.h` and not `constants/pokemon.h` is something i always found odd

GriffinRichards:
`MAX_MON_MOVES` and `NUM_STATS` are in `constants/global.h` because they get used by `global.h`. 

I disagree that this adds even a little value. It's not a random 2 popping up, it's the size of an array named `abilities`, naturally one would assume that means it holds 2 abilities. The value doesn't appear anywhere else. All this does is move where the value needs to be changed to a different file. This is a trivial change for no good reason.

PR #1274, opened 2020-12-16
Improve INSTALL.md [READY FOR REVIEW, BUT DON'T MERGE]

luckytyphlosion:
## Description
Do not merge this PR without notifying me, since the pull request made in pret/agbcc needs to be merged for this to work, plus the merge date needs to be specified before merging.

The most controversial part of this PR is the inclusion of legacy instructions for WSL1. These instructions handle installing on (modified, non-contributor) repos which do not have dkPless builds set up. People have give counterarguments to not including legacy installation instructions, stating that they should just update their repo, or follow the instructions on the hack repo. Neither of these are desirable, for the following reasons:
- The entire purpose of this PR was because the old instructions are absolutely atrocious and barely help, even for advanced users. There would be no hope for an average user to be able to follow
- Updating their repo isn't desirable either, as they would have to know how to use git merge, plus they would need to know how to merge in a specific commit from another repo. This is also not possible if the repo doesn't have the pokeemerald commits (e.g. created a new repo from the source files).

This is a hypothetical example scenario in my mind illustrating issues that this would cause:
- Someone not skilled in programming makes a repo for a hack before this PR (referred to as pokehack)
- Because the current INSTALL.md is absolutely terrible, they rely on third-party guides or assistance to set up the repo
- Now, the new INSTALL.md comes in at pret/pokeemerald, featuring dkPless installations for WSL1.
- The third-party guides are changed so that they're for the dkPless installations instead.
- Now, another person wants to contribute to pokehack, or perhaps build the repo themselves.
- Looking at the INSTALL.md in pokehack is again useless. This time, they cannot refer to the third-party guides, because those ones require a dkPless repo
- Merging the dkPless changes may be too complicated for them: already, it would require knowledge of git merge, and they would also need to know how to merge one specific commit.

Note that, they would still be able to build pokehack because I've included instructions which use msys2, but this would not be as desirable as msys2 is slower than WSL1. It may also be fine if another third-party guide is created for legacy installations. I will also say, at some point if these legacy instructions become too burdonsome to maintain (which I do not expect other maintainers to do), they can be removed once the majority of modified bases use the dkP-less installations.

If the maintainers still disagree about the legacy instructions, then I will remove them. I still believe the removal of them is a net negative to the installation instructions.

Other than that, I propose that when the INSTALL.md is changed (other than minor typos), we should require that each supported OS should be tested (preferably from scratch). I think the failure to do this for the current INSTALL.md is what caused it to slip past people who would have otherwise prevented the change from being merged.
- [x] Windows 10 (WSL1)
- [x] Windows (msys2)
- [x] Windows (Cygwin)
- [ ] macOS
- [ ] Debian/Ubuntu based
- [ ] Other Linux distros

## **Discord contact info**
luckytyphlosion#1166


luckytyphlosion:
Also, the INSTALL.md references my fork of pokeemerald and agbcc. This is for people testing the INSTALL.md if it works, and it will be changed back to pret/pokeemerald and pret/agbcc once this PR is approved for merging.

GriffinRichards:
You should consider numbering critical steps by OS. With all the notes this a lot of text, and people are going to skim it looking for the things they need to do and may miss something

INSTALL.md
```diff
@@ -1,82 +1,393 @@
...
+- `libpng-dev`
+
+Then, follow the instructions [here](https://devkitpro.org/wiki/devkitPro_pacman) to install devkitPro's pacman installer. As a reminder, the goal is to configure an existing pacman installation to recognize devkitPro's repositories.
+
+Once devkitPro pacman is configured, run the following commands:
+```bash
+sudo pacman -Sy
+sudo pacman -S gba-dev
+```
+Note: the last command will ask for the selection of packages to install. Just press Enter to install all of them, followed by entering Y to proceed with the installation.
```
GriffinRichards, line 245: The macOS instructions are missing a similar note about `sudo dkp-pacman -S gba-dev`

INSTALL.md
```diff
@@ -1,82 +1,393 @@
...
 
-## Debug info
+For example, if you want to store pokeemerald (and agbcc) in **C:\Users\\_\<user>_\Desktop\decomps**, enter this command:
+```bash
+cd Desktop/decomps
+```
+Note that the directory **must exist** in Windows. If you want to store pokeemerald in a dedicated folder that doesn't exist (e.g. the example provided above), then create the folder (e.g. using Windows Explorer) before executing the `cd` command.
 
-To build **pokeemerald.elf** with enhanced debug info:
+If this works, then proceed to [Installation](#Installation).
```
GriffinRichards, line 123: All of the "proceed to Installation" links with a capital `I` in `Installation` don't appear to link correctly (Those with a lowercase `i` appear to work correctly)

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
+These instructions explain how to set up the tools required to build **pokeemerald**, which assembles the source files into a ROM.
 
-[xcode]: https://developer.apple.com/library/archive/technotes/tn2339/_index.html
-[terminal]: https://docs.microsoft.com/windows/terminal/get-started
-[wsl]: https://docs.microsoft.com/windows/wsl/install-win10
+If you run into trouble, ask for help on Discord or IRC (see [README.md](README.md)).
 
-Independently from the specific OS, make sure that the `gcc`, `g++`, `make`, `git`, and `libpng-dev` packages or their equivalents are installed and accessible to the development tools that are used by the project (this means that, for example, on Windows, the packages have to be installed in the WSL environment). The package names and installation methods may vary with each OS.
+## Windows
+Windows has instructions for building with three possible terminals, if users encounter unexpected errors in following instructions for one of the terminals. These instructions are:
```
GriffinRichards, line 8: Not sure I understand the wording here, of `if users encounter unexpected errors in following instructions for one of the terminals.` Doesn't sound like a complete thought. What exactly do you mean?

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
+- [Windows (msys2)](#windows-msys2) (Second fastest)
+- [Windows (Cygwin)](#windows-cygwin) (Slowest)
 
-Install the devkitARM toolchain of devkitPro as per [the instructions on their wiki](https://devkitpro.org/wiki/devkitPro_pacman). On Windows, follow the Linux instructions inside WSL as any steps about the Windows installer do not apply.
+Unscientific benchmarks suggest **msys2 is 2x slower** than WSL1, and **Cygwin is 5-6x slower** than WSL1.
 
-**Debian-based distro users:** This applies to Debian, Ubuntu, and similar distros, including in WSL. If necessary, install the `libarchive13`, `pkg-config`, and `gdebi-core` packages to be able to install devkitPro.
+> Note for advanced users: **WSL2** is an option and is even faster than **WSL1** if files are stored on the WSL2 file system, but some tools may have trouble interacting with the WSL2 file system over the network drive. For example, tools which use Qt versions before 5.15.2 such as [porymap](https://github.com/huderlem/porymap) may [have problems with parsing the `\\wsl$` network drive path](https://bugreports.qt.io/browse/QTBUG-86277).
```
GriffinRichards, line 15: This will be outdated for porymap soon as it moves to Qt6. Are there are any other notable tools for which this is true?

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
-	git clone https://github.com/pret/agbcc
+2. Once the process finishes, restart your machine.
 
-	cd ./agbcc
-	./build.sh
-	./install.sh ../pokeemerald
+3. The next step is to choose and install a Linux distribution from the Microsoft Store. The following instructions will assume Ubuntu as the Linux distribution of choice.
 
-	cd ../pokeemerald
+    > Note for advanced users: Advanced users can pick a preferred Linux distribution, but setup instructions may differ. Debian should work with the given instructions, but has not been tested.
```
GriffinRichards, line 38: `Note for advanced users: Advanced users can pick...`
Redundant

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
+3. Update WSL/Ubuntu before continuing. Do this by running the following command:
 
-To build **pokeemerald.gba** with your changes:
+    ```bash
+    sudo apt update && sudo apt upgrade
+    ```
+    > Note: These commands will likely take a long time to finish.
 
-	make
+> *Note: If the repository you plan to build was created before 2020/XX/YY (e.g. modifications of pokeemerald that haven't updated) then follow the [legacy WSL1 instructions](docs/legacy_WSL1_INSTALL.md). These repositories can be identified by the [older revision](https://github.com/pret/pokeemerald/blob/571c598/INSTALL.md) of the INSTALL.md*
```
GriffinRichards, line 64: `was created before 2020/XX/YY (e.g. modifications of pokeemerald that haven't updated)`
How about `has not been updated since 2020/XX/YY` (also, replacing the date of merge when done. note its 2021 now)

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
+    The last command will ask for the selection of packages to install. Just press Enter to install all of them, followed by entering Y to proceed with the installation.
+
+### Choosing where to store pokeemerald (Linux)
+At this point, you can choose a folder to store pokeemerald (and agbcc) into. If so, you'll have to account for the modified folder path when changing directory to the pokeemerald folder.
+
+If this works, then proceed to [Installation](#installation). Otherwise, ask for help on Discord or IRC (see [README.md](README.md)).
+
+## Installation
+
+**Windows users:** Consider adding an exception for the `pokeemerald` and/or `decomps` folder in Windows Security using [these instructions](https://support.microsoft.com/help/4028485). This prevents Microsoft Defender from scanning them which might improve performance while building.
```
GriffinRichards, line 333: Could we make this visually similar to the quoted notes? As a bold-faced title under installation it makes it look like the following steps are installation for windows users

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
+For example, say you want to store pokeemerald (and agbcc) in **~/Desktop/decomps**. If pokeemerald and/or agbcc hasn't been installed yet, then enter the following command to **change directory** to the desired folder:
+```bash
+cd Desktop/decomps
+```
+Note that the directory **must exist** in the folder system. If you want to store pokeemerald in a dedicated folder that doesn't exist (e.g. the example provided above), then create the folder (e.g. using Finder) before executing the `cd` command.
+
+> Note 1: If the path has spaces, then the path must be wrapped with quotations, e.g. `cd "Desktop/decomp folder"`  
+> Note 2: If pokeemerald is already installed, then run the following command instead (for the provided example path:
+> ```bash
+> cd Desktop/decomps/pokeemerald
```
GriffinRichards, line 291: Why are we telling people to change directory into pokeemerald here (same in msys2)? If someone continues to agbcc installation from here then `./install.sh ../pokeemerald` won't work as intended

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
+
+### Debian/Ubuntu-based distributions
+Run the following command to install the necessary packages:
+```bash
+sudo apt install build-essential binutils-arm-none-eabi git libpng-dev
+```
+Then proceed to [Choosing where to store pokeemerald (Linux)](#choosing-where-to-store-pokeemerald-linux).
+
+### Other distributions
+_(Specific instructions for other distributions would be greatly appreciated!)_
```
GriffinRichards, line 307: Can this be clarified now (by you or anyone available)

INSTALL.md
```diff
@@ -1,95 +1,508 @@
...
 
 To build **pokeemerald.elf** with enhanced debug info:
+```bash
+make DINFO=1
+```
 
-	make DINFO=1
+## devkitARM's C compiler
 
+This project supports the `arm-none-eabi-gcc` compiler included with devkitARM. If devkitARM (a.k.a. gba-dev) has already been installed as part of the platform-specific instructions, simply run:
```
GriffinRichards, line 429: Should installing `gcc-arm-none-eabi` be listed as an alternative here? Its size isn't going to be relevant to everyone, and it only takes a single command to install

GriffinRichards (commented):
The links will also need to be updated when it's ready, obviously

INSTALL.md
luckytyphlosion: There are multiple instructions in case a user tries following one set of instructions, but is unable to complete the installation. They're there as backups (and also in the case of msys2, there for Windows 7 and 8).

INSTALL.md
luckytyphlosion: I tested using a bunch of tools in the WSL1 network drive, and apparently only porymap has that Qt issue.
Graphics tools tested:
- Microsoft Paint
- Paint.net
- GIMP
- GraphicsGale
- Usenti

Emulators:
- no$gba (source file inlining isn't normally detected, but if you copy the file to the same directory as the elf it'll work. I'm not sure if this is an issue with the wsl drive or just the source inlining detection failing in general. **This should probably be tested**)
- Bizhawk

Misc:
- Tilemap Studio

Note that while I didn't do any real tests, I tried using GitHub Desktop with a repository in the WSL1 Drive, and it seemed slower (although I didn't even do a comparison with the Windows Drive). This should be looked at as well.

INSTALL.md
luckytyphlosion: I think my original thought was to mention "advanced users" first, so non-advanced users could quickly skim through the notes. I guess this can be removed though.

INSTALL.md
luckytyphlosion: Maybe I'll just drop the date and make a note about older repositories in general. I'll think of something later. Saying "updated" might be misleading for modifications of pokeemerald which update past the date, but don't sync with pokeemerald.

INSTALL.md
luckytyphlosion: Yeah, I'll do this when I go through the INSTALL.md again.

INSTALL.md
luckytyphlosion: I suppose, but it isn't a necessity and the INSTALL.md has dragged on long enough. I might do it in the further future when I have more time.

INSTALL.md
luckytyphlosion: This is just a note for people reading the INSTALL.md who may want to contribute instructions to other distributions. I haven't written anything there because I don't have any Linux distribution and I'm not an expert in Linux either.

INSTALL.md
luckytyphlosion: This was supposed to be to a note to resolve the whole "need to `cd` twice" issue. It could be resolved by mentioning that this method is only to be used after agbcc and pokeemerald is installed, or maybe a separate section for "navigating to pokeemerald when you open the Terminal".

INSTALL.md
GriffinRichards: Ok. I'm not sure we need the line after the comma (surely this isn't the only reason we include instructions for other terminals) but if you want to keep it it should be `terminals in case` instead of `terminals, if`. The latter was confusing to read.

INSTALL.md
GriffinRichards: You could still mention it first, i.e. `Note for advanced users: You can pick` instead of `Note for advanced users: Advanced users can pick`

INSTALL.md
LOuroboros: "Windows has instructions for building with three possible terminals, providing 3 different options in case the user stumbled upon unexpected errors." would be an alternative way to word the thought, and it's easier to understand as well.

luckytyphlosion: Went with Lunos's suggestion

SatoMew:
I found a more modern link for downloading the Xcode Command Line Tools:
https://developer.apple.com/xcode/resources/

SatoMew:
There's also a new URL for the linked Microsoft Support article about Windows Security exclusions:
https://support.microsoft.com/topic/811816c0-4dfd-af4a-47e4-c301afe13b26

Makefile
```diff
@@ -37,6 +36,33 @@ MAKER_CODE  := 01
...
+
+ROM_NAME := pokeemerald.gba
+ELF_NAME := $(ROM_NAME:.gba=.elf)
+MAP_NAME := $(ROM_NAME:.gba=.map)
+OBJ_DIR_NAME := build/emerald
+
+MODERN_ROM_NAME := pokeemerald_modern.gba
+MODERN_ELF_NAME := $(MODERN_ROM_NAME:.gba=.elf)
+MODERN_MAP_NAME := $(MODERN_ROM_NAME:.gba=.map)
+MODERN_OBJ_DIR_NAME := build/modern
```
GriffinRichards, line 64: This (and the resulting split of `tidy`) is unnecessary. After setting `ROM_NAME`, just check if `MODERN` is 1 and append modern if it is (`ROM_NAME  := $(ROM_NAME)_modern`), and overwrite `OBJ_DIR_NAME` in the same conditional.

berry_fix/payload/Makefile
```diff
@@ -1,36 +1,51 @@
 TOOLCHAIN := $(DEVKITARM)
 COMPARE ?= 0
 
-ifeq ($(CC),)
-HOSTCC := gcc
-else
-HOSTCC := $(CC)
+# don't use dkP's base_tools anymore
```
GriffinRichards, line 4: Are any of these makefile changes actually necessary? I was under the impression this was so we could build without dkP, but this is already possible now. We do it in the CI workflow

GriffinRichards:
Making the notes sections collapsible is nice, the install is very long. macOS, Debian/Ubuntu, and other Linux distros are still unchecked for having been tested. Is that still the case?

luckytyphlosion:
> Making the notes sections collapsible is nice, the install is very long. macOS, Debian/Ubuntu, and other Linux distros are still unchecked for having been tested. Is that still the case?

Debian/Ubuntu was tested by Lunos on a VM, macOS has not been tested. I do not have access to either so I can't test it myself (and apparently none of us have access to a macOS system which suffers from the recent agbcc issues).

berry_fix/payload/Makefile
luckytyphlosion: luckytyphlosion — Today at 9:07 PM
[...]it's so people don't need to install arm-none-eabi-gcc
right now the c preprocessor used is arm-none-eabi-cpp which only comes with arm-none-eabi-gcc
so we need to use gcc -E

PR #1275, opened 2020-12-22
[LEAK INFORMED] Correct Berry Crush Structs and Eliminate Non/Fakematchings

Kurausukun:
## Description
Most of the Berry Crush structs in our repo are horribly wrong, and lead to hacks for matching and generally uglier code. This PR corrects the structs using info from the leaks, but without any documentation from the leaks in an effort not to change more than necessary.

## **Discord contact info**
Kurausukun#9923


src/berry_crush.c
```diff
@@ -77,37 +87,22 @@ struct BerryCrushGame_5C
...
-        struct BerryCrushGame_Player unk1C[5];
-        u8 fillerBC[20];
-    } as_five_players;
+    u32 unk00;
+    u16 unk04;
+    u16 unk06;
+    u16 unk08;
+    u16 unk0A;
+    u16 unk0C[2][5];
+    u8 filler20[2][8];
```
GriffinRichards, line 105: This is in use, so it should be `unk` not `filler`.

src/berry_crush.c
```diff
@@ -2341,130 +2329,126 @@ static u32 sub_80232EC(struct BerryCrushGame *r4,  __attribute__((unused)) u8 *r
...
+            r4->unk98[r7].unkE = r4->unk28;
+            ++r4->unk98[r7].unk1B;
+            if (r4->unk98[r7].unk1B > 2)
+                r4->unk98[r7].unk1B = 0;
+        }
+        else
+        {
+            r4->unk98[r7].unk1C = 0;
         }
+
```
GriffinRichards, line 2373: Unnecessary newline

src/berry_crush.c
```diff
@@ -40,34 +40,44 @@
...
 {
-    u16 unk0;
-    u16 unk2;
-    union
-    {
-        u8 as_2d_bytes[2][8];
-        u16 as_hwords[8];
-    } unk4;
-    u8 unk14[12];
+    u8 unk0[PLAYER_NAME_LENGTH + 1 + 4];
```
GriffinRichards, line 43: Do you happen to know why this name field has an additional 4 bytes?

src/berry_crush.c
Kurausukun: done

src/berry_crush.c
Kurausukun: aaaaand done

PR #1280, opened 2020-12-27
Add fixes for undefined behavior with modern compiler

Sierraffinity:
## Description
Adds fixes for three null dereferences and one division by zero that the modern compiler detected and inserted undefined opcodes (which crashed My Boy!)

## **Discord contact info**
Sierraffinity#4131

### DIFF
--- a/src/fieldmap.c
+++ b/src/fieldmap.c
@@ -797,15 +797,33 @@ bool8 CameraMove(int x, int y)
 struct MapConnection *sub_8088950(u8 direction, int x, int y)
 {
     int count;
-    struct MapConnection *connection;
     int i;
-    count = gMapHeader.connections->count;
-    connection = gMapHeader.connections->connections;
+    struct MapConnection *connection;
+    const struct MapConnections *connections = gMapHeader.connections;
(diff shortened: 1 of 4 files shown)


src/fieldmap.c
```diff
@@ -797,15 +797,33 @@ bool8 CameraMove(int x, int y)
...
-    connection = gMapHeader.connections->connections;
+    struct MapConnection *connection;
+    const struct MapConnections *connections = gMapHeader.connections;
+    // UB: Multiple possible null dereferences
+#ifdef UBFIX
+    if (connections != NULL)
+    {
+        count = connections->count;
+        connection = connections->connections;
+        if (connection != NULL)
```
mateon1, line 809: Seems like you could switch both of these to `if (... == NULL) return NULL;` which reduces the catastrophic right indents.

This would also prevent the need for duplicating code, as you can just insert
```c
#ifdef UBFIX
    if (connections == NULL) return NULL;
#endif
    count = connections->count;
    connection = connections->connections;
#ifdef UBFIX
    if (connection == NULL) return NULL;
#endif
```
... into the old code.

PR #1289, opened 2021-01-04
Change pointer to array subscript.

YuYuShima:
It just reads better.

## **Discord contact info**

YuYuShima#0776


GriffinRichards:
Loop bound is incorrect to start, these changes don't fix that

PR #1295, opened 2021-01-07
Use constants for IV-related values

sphericalice:
Introduces three constants: `MAX_PER_STAT_IVS`, `MAX_IV_MASK` and `USE_RANDOM_IVS`.

## Description
 The first, `MAX_PER_STAT_IVS`, is self-explanatory and like `MAX_PER_STAT_EVS` is simply the maximum value a given stat's IVs can reach. `MAX_IV_MASK` is the same value as `MAX_PER_STAT_IVS`, but is used instead to make explicit the bit masking being used in `pokemon.c`'s function `CreateBoxMon`. Finally, the `USE_RANDOM_IVS` constant gives a name to the flag in various `CreateMon` functions where, when provided with a value of 32 (or greater, although I think the game always uses 32 for this), the game will generate random IVs for the Pokémon instead of making each stat use the value provided.

This PR also catches an instance where `MAX_TOTAL_EVS` should be used, but currently is not.

## **Discord contact info**
Spherical Ice#4683


src/pokemon.c
```diff
@@ -4314,12 +4314,12 @@ void SetBoxMonData(struct BoxPokemon *boxMon, s32 field, const void *dataArg)
...
-        substruct3->defenseIV = (ivs >> 10) & 0x1F;
-        substruct3->speedIV = (ivs >> 15) & 0x1F;
-        substruct3->spAttackIV = (ivs >> 20) & 0x1F;
-        substruct3->spDefenseIV = (ivs >> 25) & 0x1F;
+        substruct3->hpIV = ivs & MAX_PER_STAT_IVS;
+        substruct3->attackIV = (ivs >> 5) & MAX_PER_STAT_IVS;
+        substruct3->defenseIV = (ivs >> 10) & MAX_PER_STAT_IVS;
+        substruct3->speedIV = (ivs >> 15) & MAX_PER_STAT_IVS;
+        substruct3->spAttackIV = (ivs >> 20) & MAX_PER_STAT_IVS;
+        substruct3->spDefenseIV = (ivs >> 25) & MAX_PER_STAT_IVS;
```
GriffinRichards, line 4322: These should also use `MAX_IV_MASK`

PR #1302, opened 2021-01-12
Add local label support within preproc.

luckytyphlosion:
## Description
The prefix chosen was `$` since it's a valid gnu assembler name character and isn't used by other directives. `.` would have been preferable but it would require extra coding.

Semantics:
- Local labels are scoped by a non-local (or global) label. That is, local labels will become undefined when a non-local label is used, which means that the names of local labels can be reused as long as a non-local label is defined.
- Local labels can be referenced before they are declared, like normal labels. If a local label is referenced but not defined within its scope, then preproc will report an error.
- Since this is a preproc extension, local labels **will not work in macros**. If you need this advanced support, I would suggest looking at [my fork of PikalaxALT/easyaspi's agbcc + binutils](https://github.com/luckytyphlosion/agbcc/tree/new_layout_with_libs), which implements true local label support within gnu assembler.
- Internally, preproc will transform a local label into the concatenation of the last non-local label and the local label. E.g. for this input:
    ```arm
    NonlocalLabel1:
        .byte 10
        .word $localLabel @ forward reference
    $localLabel:
        .byte 20
        .word $localLabel @ backwards reference

    NonlocalLabel2:
        .hword 30
        .word $localLabel @ forward reference
    $localLabel:
        .hword 40
        .word $localLabel @ backwards reference
    ```
    preproc will transform it to:
    ```arm
    NonlocalLabel:
        .byte 10
        .word NonlocalLabel$localLabel @ forward reference
    NonlocalLabel$localLabel:
        .byte 20
        .word NonlocalLabel$localLabel @ backwards reference

    NonlocalLabel2:
        .hword 30
        .word NonlocalLabel2$localLabel @ forward reference
    NonlocalLabel2$localLabel:
        .hword 40
        .word NonlocalLabel2$localLabel @ backwards reference
    ```
    This means that you can also reference a local label outside of its scope, but this hasn't been tested. What is guaranteed is that the local label substitution (e.g. `.word $localLabel` -> `.word NonlocalLabel2$localLabel`) will only happen if the `$` is directly after a non-word character (anything that isn't alphanumeric, `_`, `.`, or `$`).
- Per file, local labels cannot be defined before a non-local label is defined. Part of this behaviour is a bit weird and not really designed for since it represents an edge case. For example:
    ```arm
    NonlocalLabel1:
        .byte 10
    $localLabel1:
        .byte 20

        .include "some_data.inc"

    @ in file "some_data.inc"
        .word $localLabel2
    ```
    would not be valid, however:
    ```arm
    @ in file 
    NonlocalLabel1:
        .byte 10
        .word $localLabel1
    $localLabel2:
        .include "more_data.inc"
    @ in file "more_data.inc"
    NonlocalLabel2:
        .byte 20

    @ back to the base file
    $localLabel1:
        .byte 30
    ```
    is (probably) valid. However, most users will probably never feel the need for cross-file local labels.
- Tests were done in another, private repository, but if needed I could refactor some scripts to use local labels to show their correctness. The following tests were done:
    - defining a label before it's referenced, then defining the label
    - defining a label before it's referenced, and not defining it (should error)
    - defining multiple of the same label before it's referenced, and not defining it (should error)
    - defining multiple different labels before they're referenced, and not defining them (should error)
    - defining multiple different labels multiple times before they're referenced, and not defining them (should error)
    - defining local label as global (should error)
    - defining local label before a non-local label is defined (should error)
    - defining just the local label prefix (should error)
    - referencing a local label at the end of a file without a newline
    - validating local labels at the end of the file

## **Discord contact info**
luckytyphlosion#1166

### DIFF
--- a/tools/preproc/asm_file.cpp
+++ b/tools/preproc/asm_file.cpp
@@ -27,6 +27,8 @@
 #include "utf8.h"
 #include "string_parser.h"
 
+static int asmFileIdentifier;
+
 AsmFile::AsmFile(std::string filename) : m_filename(filename)
 {
     FILE *fp = std::fopen(filename.c_str(), "rb");
@@ -55,7 +57,9 @@ AsmFile::AsmFile(std::string filename) : m_filename(filename)
     m_pos = 0;
     m_lineNum = 1;
     m_lineStart = 0;
-
+    m_curNonLocalLabel = std::string{"_.L"}.append(std::to_string(asmFileIdentifier));
+    asmFileIdentifier++;
+    m_nonLocalLabelFound = false;
     RemoveComments();
 }
 
@@ -66,6 +70,10 @@ AsmFile::AsmFile(AsmFile&& other) : m_filename(std::move(other.m_filename))
     m_size = other.m_size;
     m_lineNum = other.m_lineNum;
     m_lineStart = other.m_lineStart;
+    m_curNonLocalLabel = other.m_curNonLocalLabel;
+    m_nonLocalLabelFound = other.m_nonLocalLabelFound;
+    m_localLabels = other.m_localLabels;
+    m_identified = false;
 
     other.m_buffer = nullptr;
 }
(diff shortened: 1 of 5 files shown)


GriffinRichards:
When you say it will not work in macros do you mean a label in a macro definition or a label given as an argument to a macro? If the latter I'm struggling to see a use case for this given that the repo's scripts are almost all macro

luckytyphlosion:
I mean label in a macro definition. So:
```arm
  .macro my_macro
$localLabel:
  .byte 10
  .endm
```
would not work when the macro is actually used.

coltongit:
So you're saying with this it's possible to add labels in macro definitions within the macros themselves? Would this PR still be relevant in today's pokeemerald?

GriffinRichards:
The above merge conflict was trivial and unrelated; seems like this doesn't build on current pokeemerald. Do you still have any interest in this?

tustin2121:
What purpose is this for that isn't already possible with GAS's built-in local labels `0` through `9`, and referencing them via `0b` (backward reference) or `0f` (forward reference)? https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_5.html#SEC48

luckytyphlosion:
> What purpose is this for that isn't already possible with GAS's built-in local labels `0` through `9`, and referencing them via `0b` (backward reference) or `0f` (forward reference)? https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_5.html#SEC48

gas local labels suck because they can't be named

PR #1308, opened 2021-01-17
Get rid of Task_data_OP macro

PokeCodec:
This macro is so annoying. Can we please do away with it?

## Description

Removed Task_data_OP macro

## **Discord contact info**

Shoomer#2702

### DIFF
--- a/src/m4a.c
+++ b/src/m4a.c
@@ -315,7 +315,7 @@ void MPlayExtender(struct CgbChannel *cgbChans)
 
 void MusicPlayerJumpTableCopy(void)
 {
-    asm("swi 0x2A");
+    asm("svc 0x2A");
 }
 
 void ClearChain(void *x)
--- a/src/m4a.c
+++ b/src/m4a.c
@@ -315,7 +315,7 @@ void MPlayExtender(struct CgbChannel *cgbChans)
 
 void MusicPlayerJumpTableCopy(void)
 {
-    asm("svc 0x2A");
+    asm("swi 0x2A");
 }
 
 void ClearChain(void *x)
--- a/src/main.c
+++ b/src/main.c
@@ -298,7 +298,7 @@ void InitIntrHandlers(void)
 
     REG_IME = 1;
 
-    EnableInterrupts(0x1);
+    EnableInterrupts(INTR_FLAG_VBLANK);
 }
 
 void SetVBlankCallback(IntrCallback callback)
--- a/src/main.c
+++ b/src/main.c
@@ -298,7 +298,7 @@ void InitIntrHandlers(void)
 
     REG_IME = 1;
 
-    EnableInterrupts(INTR_FLAG_VBLANK);
+    EnableInterrupts(0x1);
 }
 
 void SetVBlankCallback(IntrCallback callback)
--- a/src/task.c
+++ b/src/task.c
@@ -136,32 +136,17 @@ void TaskDummy(u8 taskId)
 {
 }
 
-#define TASK_DATA_OP(taskId, offset, op)                    \
-{                                                           \
-    u32 tasksAddr = (u32)gTasks;                            \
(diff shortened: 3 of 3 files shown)


luckytyphlosion:
```c
if (dataElem <= TASK_SPACE)
{
    gTasks[taskId].data[dataElem] = value;
    gTasks[taskId].data[dataElem + 1] = value >> 16;
```
The `TASK_SPACE` define is NOT okay, because it conflicts with the orthodox style of bounds checking, as in:
```c
if (index < arrayLength)
```
Also, just noticed this now, `TASK_SPACE` isn't even defined correctly.
```c
#define TASK_SPACE NUM_TASKS - 2 // So we can reserve space for the last array element
```
but `dataElem` is used as an index to the task **data**:
```c
gTasks[taskId].data[dataElem] = value;
```
Please, be more vigilant in ensuring your fixes do not have semantic errors and bad style, because you are annoying me more and more with these mistakes.

luckytyphlosion:
> [9:18 PM] luckytyphlosion: @Shoomer there is still an issue in your PR. gonna try something different, this time you need to figure out what the issue is
[9:18 PM] luckytyphlosion: there are 6 new lines added
[9:18 PM] luckytyphlosion: it should not be hard
[9:18 PM] luckytyphlosion: if you can't figure it out then you'll have to wait for griffin's approval

> [9:32 PM] Shoomer: let me guess: TAIL_SENTINEL?

> [9:43 PM] luckytyphlosion: TAIL_SENTINEL is fine
[9:43 PM] luckytyphlosion: and shouldn't have been removed
[9:43 PM] luckytyphlosion: that's not the issue
[9:43 PM] luckytyphlosion: @Shoomer explain your thought process as to why you thought TAIL_SENTINEL was the issue
[9:46 PM] Shoomer: everything else seemed fine, because it matches that way, and casting to s16 is what I did because data is an array of s16. By process of eminination, I found task-sentinel to be the only thing that would have been wrong.
[9:46 PM] Shoomer: SetTaskFuncWithFollowupFunc is probably the issue
[9:46 PM] luckytyphlosion: :morphon:
[9:46 PM] luckytyphlosion: please think
[9:47 PM] luckytyphlosion: that's not a good mentality to have
[9:47 PM] luckytyphlosion: of assuming something is wrong without knowing why

> {ed note: context is explaining the aforementioned issue}
[9:52 PM] luckytyphlosion: variable name {ed note: `taskNum`} doesn't make sense
[9:52 PM] luckytyphlosion: taskNum has zero uses across the repo
[9:52 PM] luckytyphlosion: it's not a name used as an index to task data
[9:53 PM] luckytyphlosion: which shows that shoomer has no understanding of the stuff that she's actually editing

> [9:53 PM] luckytyphlosion: this is more like
[9:54 PM] luckytyphlosion: followupFuncDataIndex

> [9:55 PM] luckytyphlosion: @Shoomer in your mind, what were you thinking when you named the variable taskNum
[9:56 PM] Shoomer: Num was index in my mind. also, the index is always 14, so I didn't think the variable name mattered too much
[9:56 PM] Josh: Isn’t it the number of tasks?
[9:56 PM] luckytyphlosion: it's not
[9:56 PM] ExpoSeed: no
[9:56 PM] ExpoSeed: the name is misleading
[9:56 PM] ExpoSeed: which is lucky's point

> **[10:02 PM] shinny: the reason she picked num is cause the source mentions it as workno (no as in number)**
{ed note: 3 kek reacts}

luckytyphlosion:
Final comment:
> [10:48 PM] luckytyphlosion: anyway though, I think my solution is to quickly glance over shoomer's PRs, and if I notice anything weird I comment such but don't spend effort onto giving pointers as to what's wrong

This PR will stay open on the off-chance that Shoomer manages to improve this PR into an acceptable state.

PokeCodec:
Ok, I think this PR is in good shape

PokeCodec:
This ok?

PR #1309, opened 2021-01-19
Use constants for item data, misc other constants/fixes

GriffinRichards:
Miscellaneous fixes that had been "on the list" for a while.

- Use constants for item secondary IDs, "type", and battle usage
- Add some missing `ARRAY_COUNT` usage
- Use `GET_UNOWN_LETTER` generally instead of just in `battle_anim_mons.c`
- Add some other minor missing constant usage
- Fix a few bad script labels
- Note/fix a bug in AI script for Foresight

### DIFF
--- a/include/pokemon.h
+++ b/include/pokemon.h
@@ -240,11 +240,11 @@ struct Evolution
 
 #define NUM_UNOWN_FORMS 28
 
-#define GET_UNOWN_LETTER(personality) (( \
-      ((personality & 0x03000000) >> 18) \
-    | ((personality & 0x00030000) >> 12) \
-    | ((personality & 0x00000300) >> 6)  \
-    | ((personality & 0x00000003) >> 0)  \
+#define GET_UNOWN_LETTER(personality) ((   \
+      (((personality) & 0x03000000) >> 18) \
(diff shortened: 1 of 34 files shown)


include/pokemon.h
```diff
@@ -238,6 +238,15 @@ struct Evolution
     u16 targetSpecies;
 };
 
+#define NUM_UNOWN_FORMS 28
+
+#define GET_UNOWN_LETTER(personality) (( \
+      ((personality & 0x03000000) >> 18) \
```
jiangzhengwenjz, line 244: lacks parentheses for `personality`

src/data/items.h
```diff
@@ -21,10 +20,10 @@ const struct Item gItems[] =
         .price = 0,
         .description = sMasterBallDesc,
         .pocket = POCKET_POKE_BALLS,
-        .type = 0,
-        .battleUsage = 2,
+        .type = ITEM_MASTER_BALL - 1,
```
huderlem, line 23: This falls apart if someone moves the ball item constants to somewhere other than the beginning of the item constants.

GriffinRichards: That's already the case elsewhere. See `catchAttempts` for example

PR #1312, opened 2021-01-19
Fix misnamed ground effect flags & document reflections

sphericalice:
`GROUND_EFFECT_FLAG_ICE_REFLECTION` and `GROUND_EFFECT_FLAG_REFLECTION` have mixed up names, this fixes that.
## Description
In an attempt to make adding new types of reflections easier, I created enums for the two existing reflection types. In doing so I noticed the ground effect flags described above were named incorrectly. After fixing that, I also began documented the sprite data used in `field_effect_helpers.c`, and fixed a fakematch in `event_object_movement.c`'s `GetGroundEffectFlags_Reflection`:

```diff
-           objEvent->hasReflection = 0;
-           objEvent->hasReflection = 1;
+           objEvent->hasReflection++;
```

## **Discord contact info**
Spherical Ice#4683


include/event_object_movement.h
```diff
@@ -16,14 +16,22 @@ enum SpinnerRunnerFollowPatterns
     RUNFOLLOW_SOUTH_EAST_WEST
 };
 
+enum ReflectionTypes
+{
+    REFL_TYPE_NONE,
+    REFL_TYPE_ICE,
+    REFL_TYPE_WATER
+};
+#define NUM_REFLECTION_TYPES 2
```
GriffinRichards, line 25: This should be part of the enum, `reflectionFlags` can do `NUM_REFLECTION_TYPES - 1` (it's just excluding `REFL_TYPE_NONE`)

src/event_object_movement.c
```diff
@@ -7701,7 +7703,7 @@ static void GetGroundEffectFlags_JumpLanding(struct ObjectEvent *objEvent, u32 *
     }
 }
 
-static u8 ObjectEventCheckForReflectiveSurface(struct ObjectEvent *objEvent)
+static u8 GetObjEventReflectionType(struct ObjectEvent *objEvent)
```
GriffinRichards, line 7706: The `return 0;` and `result != 0` in this function should use `REFL_TYPE_NONE`. Also while we're here let's move the definition of `RETURN_REFLECTION_TYPE_AT` to above the function definition, it's confusing to define it in the middle of the code.

Could you also elaborate on the name change? For me `GetObjEventReflectionType` implies the reflection type is part of the object event, when what it seems to be doing is getting the reflection type of metatiles around the object event

src/event_object_movement.c
```diff
@@ -7514,21 +7514,23 @@ static void ObjectEventUpdateMetatileBehaviors(struct ObjectEvent *objEvent)
...
 
-    if (type)
+    if (reflType)
     {
         if (!objEvent->hasReflection)
         {
-            objEvent->hasReflection = 0;
-            objEvent->hasReflection = 1;
-            *flags |= reflectionFlags[type - 1];
+            objEvent->hasReflection |= TRUE;
```
GriffinRichards, line 7527: This seems like it's probably a fakematch too. It's a little cleaner now at least

GriffinRichards (commented):
The rest of it looks good

src/event_object_movement.c
sphericalice: Ah, yeah I wasn't sure whether to move `RETURN_REFLECTION_TYPE_AT`, it did seem like a weird place for it. Nice catch with the 0s too, I think I saw them and then forgot to change them.

As for the name change, my approach stemmed from being misled by `Check`-- to me, I expected it to return either TRUE or FALSE, not the actual reflection type. It didn't occur to me the extra semantics my name might have introduced though, so I see your issue. Perhaps a name like `ObjectEventGetNearbyReflectionType` captures its behaviour better? Or if you think the original name was fine as-is, I can revert that change.

src/event_object_movement.c
sphericalice: It was giving me a headache looking at that double assignment, so I'll take it even if it is a fakematch lol

src/event_object_movement.c
```diff
@@ -7518,13 +7518,13 @@ static void GetGroundEffectFlags_Reflection(struct ObjectEvent *objEvent, u32 *f
...
-    u8 reflType = GetObjEventReflectionType(objEvent);
+    u8 reflType = ObjectEventGetNearbyReflectionType(objEvent);
 
     if (reflType)
     {
-        if (!objEvent->hasReflection)
+        if (objEvent->hasReflection == 0)
         {
-            objEvent->hasReflection |= TRUE;
+            objEvent->hasReflection++;
```
sphericalice, line 7527: This was a leak-informed match.

src/event_object_movement.c
```diff
@@ -7514,21 +7514,23 @@ static void ObjectEventUpdateMetatileBehaviors(struct ObjectEvent *objEvent)
...
-    if (type)
+    if (reflType)
     {
-        if (!objEvent->hasReflection)
+        if (objEvent->hasReflection == 0)
         {
-            objEvent->hasReflection = 0;
-            objEvent->hasReflection = 1;
-            *flags |= reflectionFlags[type - 1];
+            objEvent->hasReflection++;
```
PokeCodec, line 7527:
```suggestion
            objEvent->hasReflection++; // Should be objEvent->hasReflection = TRUE;
```

PokeCodec (commented):
This could be useful.

PR #1313, opened 2021-01-22
Add VBlankIntrWait to BUGFIX

PokeCodec:


ipatix:
While going into low power state is nicer for battery life and debugging, I would not consider the normal way to be a bug. At least I think BUGFIX would be inappropriate.

LOuroboros:
> 
> 
> While going into low power state is nicer for battery life and debugging, I would not consider the normal way to be a bug. At least I think BUGFIX would be inappropriate.

I agree with this, personally.

The modification done to the function on Game Freak's behalf was clearly intentional, so labelling it as a bugfix feels wrong.

PokeCodec:
Doesn't mean it wasn't wrong.

Heck, the daycare bug is labeled as a bug fix, even though GF commented they intentionally changed the code due to some bug they found at the time.

ShinyDragonHunter:
> Doesn't mean it wasn't wrong.
> 
> Heck, the daycare bug is labeled as a bug fix, even though GF commented they intentionally changed the code due to some bug they found at the time.

The Daycare bug was present in RS, too

SBird1337:
A bug does not stop being a bug, because it was intentional (to fix something else) but because it is a bug. This is not a bug, it's just not really smart. (Which could be appended as a comment or something, but using Sleep does not fix any bug (Where there never was one)

SBird1337:
A bug does not stop being a bug, because it was intentional (to fix something else) but because it is a bug. This is not a bug, it's just not really smart. (Which could be appended as a comment or something, but using Sleep does not fix any bug (Where there never was one)

LOuroboros:
> 
> 
> Doesn't mean it wasn't wrong.
> 
> Heck, the daycare bug is labeled as a bug fix, even though GF commented they intentionally changed the code due to some bug they found at the time.

Whether the change in itself was wrong or a bad idea does not matter though. The bugfix constant, from what I'm seeing, is utilized to fix bugs in the code of Pokémon Emerald, not to change a function just because you don't like its behavior.

What you're proposing about the daycare bug sounds like entirely new information that nobody around knew until now. If you want to make a case for its removal then do go ahead.

To make deliberate changes in the game's code, which is something I personally thought was frowned upon around here, a new constant label that reflects the goal of... well, improving the code of the game like you're trying to do in this PR should be added.

GriffinRichards:
> I would not consider the normal way to be a bug. At least I think BUGFIX would be inappropriate.

Agreed. The entry in the wiki is sufficient.

PR #1320, opened 2021-02-01
Remove Fakematch and add documentation

SleepyDweams:
## Description

Removed fake match in battle_factory and added a description for how selection works


SaarthakMohan:
uh oh travis fails

ExpoSeed:
uh oh travis fails

reminder to use `make compare` to test matching rather than opening prs to test matching

PokeCodec:
It was a really stupid mistake of keeping the check in there. For some reason that was added back despite the make compare initially passing.

PokeCodec:
Ready to check

PR #1336, opened 2021-02-18
Document some ribbon usage

GriffinRichards:
Documents the Pokénav ribbon summary screen, the ribbon TV show, some misc ribbon usage, and some misc constants noticed while doing these


gflib/bg.c
```diff
@@ -259,7 +259,7 @@ static void SetBgAffineInternal(u8 bg, s32 srcCenterX, s32 srcCenterY, s16 dispC
             return;
         break;
     case 2:
-        if (bg < 2 || bg > 3)
+        if (bg < 2 || bg >= NUM_BACKGROUNDS)
```
PokeCodec, line 262:
```suggestion
        if (bg != 2 && bg != 3)
```
This makes more sense because not only does this match, but it indicates that case 2 ONLY works on bg numbers 2 and 3.

PokeCodec (commented):
Just a suggestion for case 2

PR #1349, opened 2021-02-27
BUGFIX implies NONMATCHING

SleepyDweams:

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -32,6 +32,9 @@
 #ifndef UBFIX
 #define UBFIX
 #endif
+#ifndef NONMATCHING
+#define NONMATCHING
+#endif
 #endif
 
 #endif // GUARD_CONFIG_H


huderlem:
It certainly didn't used to, when we knew we had functional differences in nonmatching code, but nowadays I'm not sure.

SleepyDweams:
Nowadays it does though

ShinyDragonHunter:
BUGFIX is exactly what it says it on the tin. They are fixes to the code locked behind ifdefs so it doesn’t have any effect on matching

ExpoSeed:
Are there any examples of bugs being fixed by `NONMATCHING`? If there are, then I think this PR is good

ShinyDragonHunter:
Also, please stop force pushing

PikalaxALT:

PR #1353, opened 2021-02-28
Document mon markings

GriffinRichards:


include/mon_markings.h
```diff
@@ -1,37 +1,36 @@
-#ifndef POKEEMERALD_MON_MARKINGS_H
-#define POKEEMERALD_MON_MARKINGS_H
+#ifndef MON_MARKINGS_H
+#define MON_MARKINGS_H
 
 #define NUM_MON_MARKINGS 4
 
 struct PokemonMarkMenu
```
huderlem, line 6: should the struct be renamed to `MonMarkingsMenu`?

PR #1355, opened 2021-03-01
Address UBFixes

SleepyDweams:
Fixed the UBFix functions so that the UB is resolved


PikalaxALT:

PR #1361, opened 2021-03-05
Fix libagbsyscall

SleepyDweams:
Ok, so the reason for this PR is because a while ago, I messed up the library because I didn't know what I was doing. The telltale sign of this is that one of the functions is labeled as an ARM function, instead of a thumb function.

Because the library is never actually incorporated into the ROM, just the calls to the library, it has not had an effect on the ROM itself. However, especially when dealing with MODERN, there are cases where the function is inlined into the rom itself, and this can cause issues.

I took a copy of Nintendo's own library for the GBA, and dumped using obj dump and ghidra, so each function could be carefully verified as correct. All but 2 functions were already perfect anyway, and thus needed no changes.

Also I ordered each function alphabetically.

### DIFF
--- a/berry_fix/payload/asm/libagbsyscall.s
+++ b/berry_fix/payload/asm/libagbsyscall.s
@@ -17,12 +17,12 @@ Div: @ 81E3B68
 	bx lr
 	thumb_func_end Div
 
-	thumb_func_start Mod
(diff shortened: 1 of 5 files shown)


PikalaxALT:

PR #1362, opened 2021-03-05
[LEAK_INFORMED] Fix fakematches

SleepyDweams:
There is a convoluted fake match of UpdateLvlInHealthbox and other functions, and basically I addressed that


PikalaxALT:

PR #1364, opened 2021-03-06
Turn flags into a bitfield

SleepyDweams:
Copied from pokefirered, this brings the improvements over so that instead of dealing with bitwise operations, we can access the variables directly.


PikalaxALT:

Issue #747, opened 2019-07-20
Hard-to-reproduce tilemap glitch


PikalaxALT:
When compiling Emerald with GCC-8.3.0 or GCC-9.1.0, a weird graphical glitch sometimes occurs after a BG redraw. This glitch is difficult to reproduce but has been observed on both emulator and console.

Example:
![Screenshot of messed-up bg tilemap when emerging from Meteor Falls (Route 115 side)](https://cdn.discordapp.com/attachments/442465020291317760/601830555599896576/unknown.png)

PikalaxALT:
Glitch occurs when REG_BG((x))CNT is not properly set for display.  In the screenshot above, REG_BG2CNT = 0000 when it should be 421C (priority 2, mosaic, screen base 28)

PikalaxALT:
Glitch is likely the result of the semaphore `sGpuRegBufferLocked` being the wrong type (`bool8` instead of `vu8`).

atasro2:
why hasn't this been closed yet?

PR #1370, opened 2021-03-09
Remove unneeded volatile qualifiers

SleepyDweams:
There are multiple variables with the volatile qualifier, for good reason. However in the past, this still wasn't enough. Which was why PR #1197 was made. This changed it so the copying process would be volatile. It turns out, this was all that needed to be done, since the variable itself is not memory mapped, only when copying via the DMA should the access be volatile.

The same were done with unneeded pointers. The CONTENTS of the registers need to be volatile, not a variable storing the register's memory address. For this reason, I changed the &<register name> to use the ADDR #define.

To ensure this worked and nothing actually broke, I built a modern version of the game using this repo, and 100% the game, all gyms, all battles, and no issues.

It was a thorough test to ensure nothing actually needed to be volatile.

This should be helpful for better MODERN builds.

### DIFF
--- a/gflib/gpu_regs.c
+++ b/gflib/gpu_regs.c
@@ -10,9 +10,9 @@
 
 static u8 sGpuRegBuffer[GPU_REG_BUF_SIZE];
 static u8 sGpuRegWaitingList[GPU_REG_BUF_SIZE];
-static volatile bool8 sGpuRegBufferLocked;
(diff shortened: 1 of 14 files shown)


GriffinRichards:
Reintroduces <https://github.com/pret/pokeemerald/issues/747>

PR #1371, opened 2021-03-10
Remove unneeded volatile qualifiers

SleepyDweams:
There are multiple variables with the volatile qualifier, for good reason. However in the past, this still wasn't enough. Which was why PR #1197 was made. This changed it so the copying process would be volatile. It turns out, this was all that needed to be done, since the variable itself is not memory mapped, only when copying via the DMA should the access be volatile.

The same were done with unneeded pointers. The CONTENTS of the registers need to be volatile, not a variable storing the register's memory address. For this reason, I changed the & to use the ADDR #define.

To ensure this worked and nothing actually broke, I built a modern version of the game using this repo, and 100% the game, all gyms, all battles, and no issues.

It was a thorough test to ensure nothing actually needed to be volatile.

This should be helpful for better MODERN builds.

### DIFF
--- a/gflib/gpu_regs.c
+++ b/gflib/gpu_regs.c
@@ -10,9 +10,9 @@
 
 static u8 sGpuRegBuffer[GPU_REG_BUF_SIZE];
 static u8 sGpuRegWaitingList[GPU_REG_BUF_SIZE];
-static volatile bool8 sGpuRegBufferLocked;
-static volatile bool8 sShouldSyncRegIE;
-static vu16 sRegIE;
+static bool8 sGpuRegBufferLocked;
+static bool8 sShouldSyncRegIE;
+static u16 sRegIE;
 
 static void CopyBufferedValueToGpuReg(u8 regOffset);
 static void SyncRegIE(void);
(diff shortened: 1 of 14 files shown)


ShinyDragonHunter:
I swear this got closed. What you’re wanting to merge brings back some issues that were addressed. If one of your PRs get closed, don’t open another one with the same changes.

luckytyphlosion:
you're actually braindead shoomer

luckytyphlosion:
shoomer ban justification

RadicalDweamer:
Hello,
I am here to ask if you even read the PR or looked at the main differences. There is literally another commit SPECIFICALLY made to address the issue with the first PR, but instead of reading it, you took the ultra shortcut of dismissing the PR entirely.

Can't say I am surprised, just disappointed in you.

luckytyphlosion:
Yes, I indeed took the ultra shortcut of dismissing the PR entirely.

luckytyphlosion:
In retrospect I should have looked at the PR before closing it, most of my judgement was based on comments in #pokeemerald plus your statement implying that doing a stress test implies correctness, but too late for that now. In any case, most if not all contributors are sick of the drama surrounding you and want you permanently blocked.

luckytyphlosion:
You're also unbanned from the Discord now, but you'll be given the muted role so you won't be able to talk in the contributor channels, just so you know.

PR #1383, opened 2021-03-19
Port Some Documentation from FR

Kurausukun:
## Description
Miscellaneous stuff I found while working on FR that can be ported to Emerald

## **Discord contact info**
Kurausukun#9923


src/main.c
```diff
@@ -132,7 +132,7 @@ void AgbMain()
             DoSoftReset();
         }
 
-        if (sub_8087634() == 1)
+        if (Overworld_SendKeysToLinkIsRunning() == 1)
```
GriffinRichards, line 135: `TRUE`, and below

PR #1386, opened 2021-03-23
Remove Travis badge

SatoMew:
## Description
This PR removes a leftover from the switch to GitHub Actions.

## **Discord contact info**
SatoMew#8484

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,9 @@ This is a decompilation of Pokémon Emerald.
 
 It builds the following ROM:
 
-* [**pokeemerald.gba**](https://datomatic.no-intro.org/index.php?page=show_record&s=23&n=1961) (SHA-1 hash: `f3ae088181bf583e55daf962a92bb46f4f1d07b7`)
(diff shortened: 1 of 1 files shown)


GriffinRichards:
Why the other changes?

SatoMew:
> Why the other changes?

The hash one is just for clarity while the support links were made more prominent to follow in the spirit of #1274, which mentions them a lot.

GriffinRichards:
They seem less prominent to me now, they had their own bold header before.

ExpoSeed:
an alternative would be to reorder the sections so contacts comes before see also

SatoMew:
Would it be OK to bolden that particular text instead? I removed the header because it seemed wasteful.

GriffinRichards:
It really seems fine to me where it is, I don't think it's hard to notice

huderlem:
I think the README is fine as it is.  Swapping the Travis badge for the GitHub Actions badge is all that should change.

SatoMew:
@GriffinRichards @huderlem Done!

PR #1394, opened 2021-04-01
Fix bg event macros

GriffinRichards:
The bg event macros incorrectly interpret the size of their arguments

### DIFF
--- a/asm/macros/map.inc
+++ b/asm/macros/map.inc
@@ -51,25 +51,25 @@
 	inc _num_traps
 	.endm
 
-	.macro bg_event x, y, elevation, kind, arg6, arg7, arg8
+	.macro bg_event x, y, elevation, kind, arg6, arg7
 	.2byte \x, \y
 	.byte \elevation, \kind
 	.2byte 0
-	.if \kind < 5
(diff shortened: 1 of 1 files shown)


hjk321:
Has this fix been pushed to pokefirered?

LOuroboros:
> 
> 
> Has this fix been pushed to pokefirered?

Nope.

https://github.com/pret/pokefirered/blob/master/asm/macros/map.inc#L46-L61

GriffinRichards:
FRLG's hidden items are not the same as emerald's. Emerald has 16 bits for the flag id, and prior to this PR it was artificially limited to 8. FRLG actually do only have 8 bits for the flag id, as the other 8 bits are used by the item quantity and a flag used by the itemfinder.
<https://github.com/pret/pokefirered/blob/master/include/global.fieldmap.h#L128-L131>

The macros could still be 'improved' (it currently calls `elevation` `byte`, isn't using the constant `BG_EVENT_HIDDEN_ITEM`, and the 2 bytes of padding after `kind` could be moved to `bg_event`), but they aren't incorrect like they were in emerald.

PR #1407, opened 2021-04-17
Fix modern target when libc functions are used.

ipatix:
## Description
I came accross this problem when I tried to compile pokeemerald with modern and with NDEBUG disabled. Because what happens is that suddenly libisagbprn is used and references vsprintf, which is a libc function.

The problem with that is newlib vsprintf internally references a bunch of other things resulting in a lot of linker errors. This includes missing functions and discarded sections.

The missing functions can be fixed by linking against libnosys, which provides stub implementation for various system calls. This is what microcontroller programs usually do since they usually don't need the functionality, even if it may be indirectly referenced.

The discarded sections are fixed by adjusting the modern linker script. The most interesting part here is that a couple of bytes are added to IWRAM which add up to 0x3C bytes in total. Since it's reasonably low, I didn't see a real problem with that.

0x34 of those bytes are used for newlibs malloc implementation. I'm not quite certain if it's actually able to function since no heap boundaries are actually defined anywhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,12 +67,14 @@ override CFLAGS += -mthumb-interwork -Wimplicit -Wparentheses -Werror -O2 -fhex-
 ROM := pokeemerald.gba
 OBJ_DIR := build/emerald
 LIBPATH := -L ../../tools/agbcc/lib
+LIB := $(LIBPATH) -lgcc -lc -L../../libagbsyscall -lagbsyscall
 else
 CC1              = $(shell $(CC) --print-prog-name=cc1) -quiet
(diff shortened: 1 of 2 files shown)


ipatix:
Okay, apparrently I'm an idiot, gimme a bit to fix this.

ipatix:
It should be functionally fine now. One thing one could change is the LIBPATH and LIB thing and just put it into a single line since there is no longer the need to keep them separate.

ipatix:
Before merging this, I'd like to check if certain link problems for normal agbcc need to be adressed when NDEBUG is not defined.

ipatix:
Okay, regular compilation with agbcc horribly breaks simply when disabling NDEBUG. However, I will not focus on this in this PR and I suggest that this should be solved separately.

ipatix:
Will still not compile with devkitARM's newlib. It seems to be missing functions in libc for some reason and I don't know why. At the moment I don't know how one would fix this, so I suggest that not to be part of this PR.

PR #1413, opened 2021-04-21
Use designated initializers for object_event_graphics_info.h

ExpoSeed:
## Description

Use designated initializers for object_event_graphics_info.h

Apologies in advance to hackers 

## **Discord contact info**

ExpoSeed#5613


GriffinRichards:
I agree that this is a reasonable change, but unfortunately at the moment this is incompatible with how Porymap parses this file

ExpoSeed:
Darn, wasn't aware of that :(

PR #1432, opened 2021-05-11
UBFix: uninitialized variables in m4a engine and siirtc.c

gAlfonso-bit:
This PR aims to correct “UB” where variables were read initialized in the m4a engine and siirtc


src/siirtc.c
```diff
@@ -378,7 +385,11 @@ bool8 SiiRtcSetAlarm(struct SiiRtcInfo *rtc)
     return TRUE;
 }
 
+#ifndef UBFIX
 static int WriteCommand(u8 value)
+#else
+static int WriteCommand(u8 value)
```
ipatix: Why is there an if else with the same function signature?

ShinyDragonHunter: > Why is there an if else with the same function signature?
I do think this got addressed in another one of their commits

src/siirtc.c
```diff
@@ -98,8 +103,12 @@ u8 SiiRtcProbe(void)
 
     errorCode = 0;
 
+#ifndef BUGFIX
```
huderlem, line 106: What is the bug here?

src/m4a.c
```diff
@@ -1525,6 +1525,10 @@ void ply_xwave(struct MusicPlayerInfo *mplayInfo, struct MusicPlayerTrack *track
 {
     u32 wav;
 
+#ifdef UBFIX
```
huderlem, line 1528: There actually isn't any undefined behavior with these of these zero assignments, due to the bit twiddling that occurs.

src/siirtc.c
gAlfonso-bit: If both power failure happened and the clock was 12 hour, the clock wouldn’t be reset

src/m4a.c
gAlfonso-bit: I don’t see any such thing. I see var being &=, which means the value is reassigned based on the previous value.

gAlfonso-bit:
Merge conflicts resolved and duplicate changes already upstream removed

src/m4a.c
huderlem: Yes, but you're ignoring what the result of the operations is.  All four groups of 8-bits in the 32-bit value are written individually such that all bits are accounted for, and therefore no undefined behavior.  This is true for `ply_xcmd_0C` because it operates on a 16-bit value there.

huderlem: That being said, there might still be an argument to explicitly set the 0, like you've done.  For example, on a 64-bit system, this would be undefined beahvior.

gAlfonso-bit: I think it’s best to have it here just in case, just so the compiler at least is aware, plus safer code isn’t a bad thing. Otherwise I will just remove them.

PR #1436, opened 2021-05-14
BUGFIX: battle scripts

gAlfonso-bit:
## **Discord contact info**

AlfonsoB#1500


src/battle_ai_script_commands.c
```diff
@@ -2025,7 +2035,11 @@ static void Cmd_if_holds_item(void)
     var2 = gAIScriptPtr[2];
     var1 = gAIScriptPtr[3];
 
+#ifndef UBFIX
```
huderlem, line 2038: This looks like a bug fix, not an undefined behavior fix.

ExpoSeed: adding onto this, it would be a good idea if you explain the bug itself and the effect of the bug in a comment

ExpoSeed (changes requested):

src/battle_script_commands.c
```diff
@@ -3439,7 +3439,11 @@ static void Cmd_getexp(void)
                     gBattleMons[0].spDefense = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_SPDEF);
                 }
                 // What is else if?
+#ifndef BUGFIX
```
huderlem, line 3442: I don't believe there is any bug here.

src/battle_script_commands.c
```diff
@@ -6168,7 +6172,12 @@ static void Cmd_recordlastability(void)
 {
     gActiveBattler = GetBattlerForBattleScript(gBattlescriptCurrInstr[1]);
     RecordAbilityBattle(gActiveBattler, gLastUsedAbility);
+
+#ifndef UBFIX
```
huderlem, line 6176: This seems like a bug, not undefined behavior.  Luckily, it's never used in any battle scripts.

gAlfonso-bit:
Merge conflicts and issues resolved

src/battle_script_commands.c
```diff
@@ -3433,23 +3433,28 @@ static void Cmd_getexp(void)
                     gBattleMons[0].defense = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_DEF);
                     // Why is this duplicated?
                     gBattleMons[0].speed = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_SPEED);
+#ifndef BUGFIX
                     gBattleMons[0].speed = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_SPEED);
+#endif
```
GriffinRichards, line 3438: This is not a bug.

src/battle_script_commands.c
```diff
@@ -3438,18 +3438,21 @@ static void Cmd_getexp(void)
...
+                // if statement below should be checking (gBattleTypeFlags & BATTLE_TYPE_DOUBLE) first, not last
                 if (gBattlerPartyIndexes[2] == gBattleStruct->expGetterMonId && gBattleMons[2].hp && (gBattleTypeFlags & BATTLE_TYPE_DOUBLE))
                 {
                     gBattleMons[2].level = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_LEVEL);
                     gBattleMons[2].hp = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_HP);
                     gBattleMons[2].maxHP = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_MAX_HP);
                     gBattleMons[2].attack = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_ATK);
                     gBattleMons[2].defense = GetMonData(&gPlayerParty[gBattleStruct->expGetterMonId], MON_DATA_DEF);
-                    // Duplicated again, but this time there's no Sp Defense
+                    // BUG: Speed is Duplicated again, but this time there's no Sp Defense being updated, even though it ought to be.
```
luckytyphlosion, line 3449: Duplicated should be lowercase

gAlfonso-bit:
For Cmd_if_holds_item, the bitwise OR operandi are switched. This doesn’t cause a bug in the ROM because the only time this function is used is when checking ITEM_PERSIM_BERRY, which is 140, and the following operand is sAI_Target, which is 0. This means the top 8 bits are 0 anyway. So the shift doesn’t actually matter. Is this BUGFIX even worth keeping or should bugfix remove the OR operation entirely?

gAlfonso-bit:
Perhaps the change should be a wiki entity instead of a bugfix? Any recommendations?

PR #1441, opened 2021-05-17
UBFIX sprite.c: change comparison with 0 to comparison with 1

gAlfonso-bit:
## **Discord contact info**

AlfonsoB#1500

### DIFF
--- a/gflib/sprite.c
+++ b/gflib/sprite.c
@@ -492,6 +492,12 @@ void SortSprites(void)
                 }
             }
         }
+#ifdef UBFIX //To sort the first two sprites
+        {
+            u8 temp = sSpriteOrder[1];
+            sSpriteOrder[1] = sSpriteOrder[0];
+            sSpriteOrder[0] = temp;
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


gflib/sprite.c
```diff
@@ -439,7 +439,11 @@ void SortSprites(void)
             }
         }
 
+#ifndef UBFIX
         while (j > 0
+#else
+        while (j > 1
```
huderlem, line 445: This introduces a bug where the first two elements won't be sorted properly because it will terminate the while loop too early.

gAlfonso-bit: What about now?

huderlem: It's now buggier and is also missing the closing directive for the new UBFIX section.  I'm going to close this PR because it's clear you aren't testing or validating any of the changes, which is especially important for an important piece of code like this.

gAlfonso-bit: Alright. I’m going to do an investigation and debug this section. Turns out this code is more complicated than I thought.

I apologize for my rash commits.

PR #1442, opened 2021-05-20
UBFIX: sprite.c

gAlfonso-bit:
Finally figured out a solution; break when j is 0, instead of checking at the beginning of the while loop

## **Discord contact info**

AlfonsoB#1500


ShinyDragonHunter:
Are you testing these fixes before suggesting them here?

gAlfonso-bit:
Everything is tested and working.

PR #1444, opened 2021-05-22
[NO SYNTAX CHANGES] Fix argument name mismatch warning and shadowing local variable

gAlfonso-bit:
No syntax other than variable names have been changed.

## **Discord contact info**

AlfsonsoB#1500

### DIFF
--- a/src/battle_interface.c
+++ b/src/battle_interface.c
@@ -2340,11 +2340,11 @@ static s32 CalcNewBarValue(s32 maxValue, s32 oldValue, s32 receivedValue, s32 *c
 
     if (maxValue < scale) // handle cases of max var having less pixels than the whole bar
     {
-        s32 newToAdd = Q_24_8(maxValue) / scale;
+        s32 delta = Q_24_8(maxValue) / scale;
 
         if (receivedValue < 0) // fill bar right
         {
-            *currValue += newToAdd;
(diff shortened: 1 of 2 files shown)


gAlfonso-bit:
If there’s a problem with the variable names, please do not hesitate to say so.

src/battle_interface.c
```diff
@@ -2340,11 +2340,11 @@ static s32 CalcNewBarValue(s32 maxValue, s32 oldValue, s32 receivedValue, s32 *c
 
     if (maxValue < scale) // handle cases of max var having less pixels than the whole bar
     {
-        s32 toAdd = Q_24_8(maxValue) / scale;
```
huderlem, line 2343: I think keeping the variable shadowing might be fine here.  Changing it to a Frankenstein name isn't worth it, in my opinion.  `delta` might be a better name than `toAdd` for the whole function..

gAlfonso-bit:
Does this work or should I just remove the change and only keep the Battle Factory function change

huderlem:
> Does this work or should I just remove the change and only keep the Battle Factory function change

No, that is not what I meant.  I don't think this PR adds any value as a whole--changing the entire function to use a "delta" instead of "toAdd" name isn't really an improvement either.

PR #1447, opened 2021-05-24
Fixed ChangeBgY_ScreenOff‘s signature.

gAlfonso-bit:
It’s the only function of the “ChangeBy” functions that uses a u32 for value when it should be s32 like the rest.

## **Discord contact info**

AlfonsoB#1500

### DIFF
--- a/gflib/bg.c
+++ b/gflib/bg.c
@@ -252,7 +252,6 @@ static void SetBgAffineInternal(u8 bg, s32 srcCenterX, s32 srcCenterY, s16 dispC
 
     switch (sGpuBgConfigs.bgVisibilityAndMode & 0x7)
     {
-
     default:
     case 0:
         return;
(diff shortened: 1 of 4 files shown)


gflib/bg.c
```diff
@@ -695,7 +696,7 @@ s32 ChangeBgY(u8 bg, s32 value, u8 op)
     return sGpuBgConfigs2[bg].bg_y;
 }
 
-s32 ChangeBgY_ScreenOff(u8 bg, u32 value, u8 op)
+s32 ChangeBgY_ScreenOff(u8 bg, s32 value, u8 op)
```
huderlem, line 699: Maybe they should all be `u32`.  How do you know which is right?

gflib/bg.c
```diff
@@ -252,17 +252,18 @@ static void SetBgAffineInternal(u8 bg, s32 srcCenterX, s32 srcCenterY, s16 dispC
 
     switch (sGpuBgConfigs.bgVisibilityAndMode & 0x7)
     {
+
```
huderlem, line 255: Why did you move the `default` and `0` cases to the top of the `switch`, in addition to adding an extra newline?

gflib/bg.c
```diff
@@ -280,10 +281,10 @@ static void SetBgAffineInternal(u8 bg, s32 srcCenterX, s32 srcCenterY, s16 dispC
     SetGpuReg(REG_OFFSET_BG2PC, dest.pc);
     SetGpuReg(REG_OFFSET_BG2PD, dest.pd);
     SetGpuReg(REG_OFFSET_BG2PA, dest.pa);
-    SetGpuReg(REG_OFFSET_BG2X_L, (s16)(dest.dx));
```
huderlem, line 283: Does this match without the explicit casts?

gflib/bg.c
gAlfonso-bit: Others are s32 and the values being operated on are signed. Plus it makes no sense why a delta parameter should be unsigned

PR #1448, opened 2021-05-25
[LEAK-INFORMED] Fix Fakematch in DrawWallpaper

Kurausukun:
## Description
Yes, you actually need a second variable to match it this way, it does not match if offset is modified within the function call no matter if you do * 256 or << 8. This compiler is bad

## **Discord contact info**
Kurausukun#9923

### DIFF
--- a/src/pokemon_storage_system.c
+++ b/src/pokemon_storage_system.c
@@ -5437,19 +5437,20 @@ static bool32 WaitForWallpaperGfxLoad(void)
 
 static void DrawWallpaper(const void *tilemap, s8 direction, u8 offset)
 {
-    s16 var = (offset * 2) + 3;
+    s16 var = offset * 256;
+    s16 var2 = (offset * 2) + 3;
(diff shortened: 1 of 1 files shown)


src/pokemon_storage_system.c
```diff
@@ -5437,19 +5437,20 @@ static bool32 WaitForWallpaperGfxLoad(void)
...
+    CopyRectToBgTilemapBufferRect(2, tilemap, 0, 0, 0x14, 0x12, x, 2, 0x14, 0x12, 0x11, var, var2);
 
+    if (direction)
+    {
+        if (direction > 0)
+            x += 0x14;
+        else
+            x -= 4;
     FillBgTilemapBufferRect(2, 0, x, 2, 4, 0x12, 0x11);
+    }
```
GriffinRichards, line 5453: The original early return would still match here, you only need to remove `x *= 1`. Regardless it has incorrect indentation

Kurausukun: oops, sec

PR #1451, opened 2021-05-29
Use standard strcpy code

gAlfonso-bit:
The strcpy code is public domain and it also works here. Best to use it because the compiler recognizes this as strcpy.


PikalaxALT:
This PR is spam.

PR #1452, opened 2021-05-30
Consolidate sprite tags

huderlem:
This is an attempt to consolidate all sprite tags into a single file, `constants/sprite_tags.h`.  There is a common mistake made by decomp users, which is accidentally duplicating sprite tag values that are already loaded.  This causes graphics or palettes to load incorrectly and can be very difficult to debug.  Since sprite tags are currently defined separately in their respective files, it's easy to accidentally duplicate tags--especially within the battle systems.  By laying them all out explicitly in a single file, it should be simple to determine which tags are already in-use.


GriffinRichards:
Any objection to me taking `TAG_NONE` from this so it can go upstream? I don't imagine this will be finished any time soon.

GriffinRichards:
Wild guess, this isn't getting picked up again. Feel free to reopen if that changes

PR #1454, opened 2021-05-31
[Leak informed] Cleaned up battle_transition and other files too

gAlfonso-bit:
## Description

Had a list of cleanups I wanted to try out.

Will probably whittle this down to only the useful things.

Had to cheat a little for the fake match and macro

## **Discord contact info**

AlfonsoB#1500


huderlem:
This is largely the same PR you sent yesterday, and you're even trying to include things that were rejected 2 PRs.  Given the sideways nature of your PRs, the fact that you don't have a good understanding of the changes you make, and the fact that I don't want to play this game anymore, I'm blocking your ability to create PRs.

PR #1460, opened 2021-06-03
Buildsystem no longer builds temporary files for C compilation by default, and other makefile optimizations.

luckytyphlosion:
## Description
- Buildsystem no longer generates *.i and *.s files when compiling C sources. The old behaviour can be requested by defining `KEEP_TEMPS=1`. This required modifying `preproc` to accept input from stdin. The method that yama used to read in an entire file is actually non-standard and technically undefined behaviour ([external explanation](https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek%28%29+and+ftell%28%29+to+compute+the+size+of+a+regular+file))
- Optimizations in the makefile:  no more unnecessary recursive makes (i.e. `compare` and `modern`, these would also scan dependencies multiple times although without this change it still would have been fixed), skip over certain makefile rules for rules that don't build the ROM, use minimal makefile to build tools.
- Fix issue with NODEP=1 not compiling.

## **Discord contact info**
luckytyphlosion#1166


luckytyphlosion:
Forgot to mention, these changes also adds explicit dependency rules for C files in src, C files in gflib, asm files in src, and asm files in asm/, so that if you include a file in any of the aforementioned file types that doesn't exist, you will get at least a semi-relevant error instead of something completely cryptic. This isn't ideal, but it at least gives enough information to debug the problem.

tools/preproc/c_file.cpp
```diff
@@ -23,39 +23,66 @@
...
-    if (std::fread(m_buffer, m_size, 1, fp) != 1)
-        FATAL_ERROR("Failed to read \"%s\".\n", filename.c_str());
+            numAllocatedBytes += CHUNK_SIZE;
+            bufferOffset += CHUNK_SIZE;
+            m_buffer = (char *)realloc(m_buffer, numAllocatedBytes);
+            if (m_buffer == NULL) {
+                FATAL_ERROR("Failed to allocate memory to process file \"%s\"!", filename.c_str());
+            }
+        } else {
+            FATAL_ERROR("Failed to read \"%s\". (error: %s)", filename.c_str(), std::strerror(errno));
```
GriffinRichards, line 75: This fails when I test it locally.

```
c_file.cpp:75:95: error: use of undeclared identifier 'errno'
            FATAL_ERROR("Failed to read \"%s\". (error: %s)", filename.c_str(), std::strerror(errno));
                                                                                              ^
1 error generated.
```

GriffinRichards:
Nice. As stated in Discord the optimizations only improve clean build speed by about 7% (for me at least), but improving the missing dependency error is big. Comparing the errors:

Old (after a clean, can also just silently fail with an apparently successful build):

```
Assembler messages:
Error: can't open src/file_with_missing_dependency.s for reading: No such file or directory
```

New:

```
gmake: *** No rule to make target 'src/missing_dependency.h', needed by 'build/emerald/src/file_with_missing_depdency.o'.  Stop.
```

PR #1466, opened 2021-06-17
Fix sATypeMove_Table definition

Sewef:
## Description

`NUMBER_OF_MON_TYPES - 1` should be the max string size.

## **Discord contact info**

Citron#3400

### DIFF
--- a/src/battle_message.c
+++ b/src/battle_message.c
@@ -1328,7 +1328,7 @@ static const u8 sText_SpaceIs[] = _(" is");
 static const u8 sText_ApostropheS[] = _("'s");
 
 // For displaying names of invalid moves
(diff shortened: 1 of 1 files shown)


src/battle_message.c
```diff
@@ -1328,7 +1328,7 @@ static const u8 sText_SpaceIs[] = _(" is");
 static const u8 sText_ApostropheS[] = _("'s");
 
 // For displaying names of invalid moves
-static const u8 sATypeMove_Table[][NUMBER_OF_MON_TYPES - 1] =
+static const u8 sATypeMove_Table[NUMBER_OF_MON_TYPES][17 =
```
LOuroboros, line 1331:
```suggestion
static const u8 sATypeMove_Table[NUMBER_OF_MON_TYPES][17] =
```

PR #1476, opened 2021-07-02
Fix modern builds when devkitARM is not in PATH.

luckytyphlosion:
## Description
Done by just exporting PATH for the `$(shell $(MODERNCC))` calls

## **Discord contact info**
luckytyphlosion#1166


Makefile
```diff
@@ -28,6 +28,7 @@ LD := $(PREFIX)ld
 # note: the makefile must be set up so MODERNCC is never called
 # if MODERN=0
 MODERNCC := $(PREFIX)gcc
+PATH_MODERNCC := PATH=$(TOOLCHAIN)/bin:PATH $(MODERNCC)
```
GriffinRichards, line 31: Why not just `MODERNCC := PATH=$(TOOLCHAIN)/bin:PATH $(PREFIX)gcc`?

luckytyphlosion: That could work as well but maybe some future hypothetical `MODERNCC` might want the variable alone or something (I don't really have a good reason here).

PR #1484, opened 2021-07-20
Remove address comments

GriffinRichards:
The address comments have been made redundant with the creation of the symbols branch. More often than not they just confuse newcomers now. 

Comments were removed with the regex ` @ [A-F\d]{6,8}\n` to be replaced with `\n`


ShinyDragonHunter:
Can the voicegroup inc files have the voice number instead of the address offset? An example of what I mean would be `voice_keysplit_all voicegroup001  @ 0` instead of `voice_keysplit_all voicegroup001  @ 8675D04`

GriffinRichards:
I don't see how the comment you've shown as an example is particularly helpful, but in any case that's a proposal for its own PR. This is just removing the address comments.

PR #1492, opened 2021-08-09
Fix CgbSound Fakematches (Thanks to Revo)

Kurausukun:
## Description
Due to Revo decompiling an earlier version of m4a, these fakematchings have come to light; I am porting them here for consistency in m4a across projects.

## **Discord contact info**
Kurausukun#9923

### DIFF
--- a/src/m4a.c
+++ b/src/m4a.c
@@ -1198,7 +1198,7 @@ void CgbSound(void)
             }
             else
             {
-                unsigned int envMask = 0xF;
+                u32 envMask = 0xF;
                 *nrx2ptr = (envelopeStepTimeAndDir & envMask) + (channels->envelopeVolume << 4);
                 *nrx4ptr = channels->n4 | 0x80;
                 if (ch == 1 && !(*nrx0ptr & 0x08))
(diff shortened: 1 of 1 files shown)


src/m4a.c
```diff
@@ -1198,8 +1198,8 @@ void CgbSound(void)
             }
             else
             {
-                envelopeStepTimeAndDir &= 0xf;
-                *nrx2ptr = (channels->envelopeVolume << 4) + envelopeStepTimeAndDir;
+                unsigned int envMask = 0xF;
```
GriffinRichards, line 1201: `u32` (or `u8`, it doesn't really matter here). In any case I doubt `envMask` is needed, why not just use `0xF` directly

ProjectRevoTPP: It's required on older m4a versions and with this var it matches both ways, so I think it was like this originally and direct use is actually a fakematch.

Kurausukun: It's not needed here, but it is needed in previous revisions, according to Revo, so I kept it like that for consistency. I asked the same thing. I can change it to u32 though, that's fair

PR #1508, opened 2021-09-29
Use compact weather macros

Pokestia:
Title says all.

Feel free to close/ignore if not helpful.


GriffinRichards:
I agree with the change, I'm just not sure the existing names make sense in the context of assignment. Perhaps it's time to rename them (and prefix them `B_` to distinguish battle weather from overworld weather) e.g. `WEATHER_SANDSTORM_ANY` -> `B_WEATHER_SANDSTORM`

Pokestia:
I can do that.

The AI_WEATHER constants refer to the battle weathers as well, should I rename them as well? e.g. `AI_WEATHER_SUN` -> `AI_B_WEATHER_SUN`

Also, `WEATHER_HAIL` and `WEATHER_HAIL_ANY` have the same value so I will unify them as `B_WEATHER_HAIL.`

GriffinRichards:
No, there's no need to rename the AI constants. Those should be clear enough already, as the AI is only a battle feature.

The extra Hail constant was created I believe to help accommodate people adding things like Snow Warning. They could be renamed like this:
`WEATHER_HAIL_ANY` -> `B_WEATHER_HAIL`
`WEATHER_HAIL` -> `B_WEATHER_HAIL_TEMPORARY`

PR #1509, opened 2021-10-03
Document remaining symbols in pokemon.c

GriffinRichards:


src/data/trainers.h
```diff
@@ -11258,7 +11258,7 @@ const struct Trainer gTrainers[] = {
     [TRAINER_STEVEN] =
     {
         .partyFlags = F_TRAINER_PARTY_HELD_ITEM | F_TRAINER_PARTY_CUSTOM_MOVESET,
-        .trainerClass = TRAINER_CLASS_PKMN_TRAINER_3,
+        .trainerClass = TRAINER_CLASS_RIVAL,
```
LOuroboros, line 11261: `TRAINER_STEVEN` is hardly considered a rival though. Or to put it in perspective, he's as much of a rival as Lance in Pokémon GSC.
Isn't there a better label for this trainer class?

GriffinRichards: It's pretty clearly the rival trainer class, the only other trainers that use it are Brendan May and Wally, and it's used to play the Rival battle music. Steven also isn't part of the Elite Four in Emerald, he encounters the player throughout the game and is the final battle akin to Red in GSC

LOuroboros: I mean, yes, I understand this trainer class is used by the actual rivals, which are the opposite gendered main character and Wally, but this is Steven. Regardless of whether he's the champion or not, he's not really a rival. Neither is Red in GSC, since you're bringing him up. I have a hard time thinking anyone would say Red was a rival when he was just the endgame challenge.

If there isn't a better label, so be it. I'm just saying it feels wrong to have a trainer class that is supposedly meant to encompass the rivals, but then is used to mark a trainer that is not considered a rival.

GriffinRichards: Take it up with Game Freak, this label helps convey what the trainer class does. It gets treated specially to play the rival music and handle Wally's name. The battle with Steven in Emerald plays the rival music, and the class is otherwise used for the rivals.

PR #1510, opened 2021-10-04
Clean up roamer.c

GriffinRichards:


src/roamer.c
```diff
@@ -5,51 +5,74 @@
...
-    MAP_NUM = 1, // map number
+    MAP_GRP, // map group
+    MAP_NUM, // map number
 };
 
+#define ROAMER (&gSaveBlock1Ptr->roamer)
 EWRAM_DATA static u8 sLocationHistory[3][2] = {0};
 EWRAM_DATA static u8 sRoamerLocation[2] = {0};
 
+#define __ MAP_NUM(UNDEFINED) // For empty spots in the location table
```
PikalaxALT, line 22: `__` conflicts with a preproc directive representing a charcoded string with no terminator.

PikalaxALT (changes requested):

src/roamer.c
GriffinRichards: Preproc handles this fine. It expects a left paren afterwards, otherwise it continues on. Did you mean this IDE support hack?
<https://github.com/pret/pokeemerald/blob/master/include/global.h#L25>

PR #1533, opened 2021-10-29
Renamed "SIZE" text controller

LOuroboros:
## Description
Both, the label and the comment for the text controller `SIZE` are incorrect.
I don't think it changes the size specifically.
After fiddling with it, I feel like it changes the font for the current text string.
There isn't one single valid value to use with it either *(`SMALL`)*, there's 5 for the English font, and 3 for the Japanese font, I'd say.

Maybe some constants should be added for these values.

## **Discord contact info**
Lunos#4026


GriffinRichards:
Could you rename the corresponding C define as well? The one here:
https://github.com/pret/pokeemerald/blob/6dc3c9864c80557a81c080e0a5893d9d8765c635/gflib/text.h#L214

LOuroboros:
> 
> 
> Could you rename the corresponding C define as well? The one here:
> 
> https://github.com/pret/pokeemerald/blob/6dc3c9864c80557a81c080e0a5893d9d8765c635/gflib/text.h#L214

Done :eyes:

charmap.txt
```diff
@@ -414,7 +414,7 @@ HIGHLIGHT = FC 02 @ same as fc 01
 SHADOW = FC 03 @ same as fc 01
 COLOR_HIGHLIGHT_SHADOW = FC 04 @ takes 3 bytes
 PALETTE = FC 05 @ used in credits
-SIZE = FC 06 @ note that anything other than "SMALL" is invalid
+FONT = FC 06 @ valid values are 0, 1, 2, 7 and 8
```
GriffinRichards, line 417: Oops, sorry to not lump the comments together, but looks like 6 is a valid font ID as well (for Braille)

LOuroboros: Should I modify the comment like "0, 1, 2, 6, 7 and 8" or "0, 1, 2, 6 (braille), 7 and 8"? :eyes:

GriffinRichards: Sure. If we end up adding font id constants later we'll probably replace those numbers with the constant names, or a note on where to find them.

LOuroboros: > Whoops by sure I meant the latter lol

Done :eyes:

PR #1538, opened 2021-11-02
Rename some player movement speed functions

cbt6:
## Description
- The `WalkFastest` functions are incorrectly named because they use the speed `MOVE_SPEED_FASTER` and not `MOVE_SPEED_FASTEST`
- Since the speed numbers used in the `GoSpeedX` function names are wrong, use the movement speeds' named constants instead

## **Discord contact info**
cbt#8922


src/data/object_events/movement_action_func_tables.h
```diff
@@ -602,12 +602,20 @@ u8 (*const gMovementActionFuncs_FaceRight[])(struct ObjectEvent *, struct Sprite
     MovementAction_PauseSpriteAnim,
 };
 
+enum {
+    MOVE_SPEED_NORMAL, // walking
+    MOVE_SPEED_FAST_1, // running / surfing / sliding (ice tile)
+    MOVE_SPEED_FAST_2, // water current / acro bike
+    MOVE_SPEED_FASTER, // mach bike's max speed
+    MOVE_SPEED_FASTEST,
+};
```
GriffinRichards, line 611: Move these either to the top of `event_object_movement.c` or to one of its header files so they're not buried.

PR #1541, opened 2021-11-03
Use constants for movement speeds

cbt6:
## Description
Missed out on a few functions in #1538.

## **Discord contact info**
cbt#8922

### DIFF
--- a/src/event_object_movement.c
+++ b/src/event_object_movement.c
@@ -8349,15 +8349,19 @@ static const SpriteStepFunc sStep8Funcs[] = {
 };
 
 static const SpriteStepFunc *const sNpcStepFuncTables[] = {
-    sStep1Funcs,
-    sStep2Funcs,
-    sStep3Funcs,
-    sStep4Funcs,
-    sStep8Funcs,
+    [MOVE_SPEED_NORMAL] = sStep1Funcs,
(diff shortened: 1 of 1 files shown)


src/event_object_movement.c
```diff
@@ -5184,7 +5184,7 @@ static void InitMovementNormal(struct ObjectEvent *objectEvent, struct Sprite *s
 
 static void StartRunningAnim(struct ObjectEvent *objectEvent, struct Sprite *sprite, u8 direction)
 {
-    InitNpcForMovement(objectEvent, sprite, direction, 1);
+    InitNpcForMovement(objectEvent, sprite, direction, MOVE_SPEED_FAST_1);
```
GriffinRichards, line 5187: Despite referring to variables called speed, aside from this one none of these new uses are related to your previous uses of these constants. I'm not sure it makes sense to connect them.

The previous uses relate to normal movement (`sDirectionAnimFuncsBySpeed` from `InitMovementNormal`, and ultimately in `sStepTimes` and `sNpcStepFuncTables`, which should also be using these constants then). 

The new ones in jump movements are indexes into `sJumpInitDisplacements`, `sJumpDisplacements`, and later to the arrays in `DoJumpSpriteMovement` and `DoJumpSpecialSpriteMovement`. These arrays don't map to this set of constants.

PR #1547, opened 2021-11-07
Clean up sound, add CRY_MODE constants

GriffinRichards:
- Adds `CRY_MODE_*` and other constants for the `PlayCry` functions, and gives them non-generic names (i.e. `PlayCry1` --> `PlayCry_Normal`)
- `cry2` and `gCryTable2` are now `cry_reverse` and `gCryTable_Reverse`
- Numbers the phonemes in `direct_sound_samples ` and moves them to their own sub-folder, as in pokeruby
- Gives generic names to the remaining address-named `direct_sound_samples` and programmable wave data


GriffinRichards:
Including a fix to an issue with calcrom revealed by this PR. Undocumented symbols includes false positives with legitimate uses of "unknown" (e.g. `gPokedexAreaScreenAreaUnknown_Gfx`) that are not found by the "partially documented" command, and because calcrom assumes "undocumented" is a subset of "partially documented" and subtracts the two this can result in a negative number of partially documented symbols.

PR #1556, opened 2021-11-17
Use movement action constant values over enum

cbt6:
## Description
Also added movement actions for Acro Bike for completion's sake.

## **Discord contact info**
cbt#8922


asm/macros/movement.inc
```diff
@@ -1,131 +1,166 @@
-	.macro create_movement_action name
-	enum _\name
+	.macro create_movement_action name, value
 	.macro \name
```
huderlem, line 2: These should probably just be constants instead of macros, I think?  The macros simply output a single byte.

GriffinRichards: The macros are simplified names for the constants (plus the byte directive). Movement scripts would be unwieldy with the C defines

asm/macros/movement.inc
```diff
@@ -1,131 +1,166 @@
-	.macro create_movement_action name
-	enum _\name
+	.macro create_movement_action name, value
```
GriffinRichards, line 1: Looks like the old macro was created before we started adding argument requirements.
Should be `.macro create_movement_action name:req, value:req`

asm/macros/movement.inc
huderlem: True, I forgot about that.

PR #1559, opened 2021-11-21
Fix incorrect flower animation palettes

GriffinRichards:
The graphics for the Mauville tileset anim `flower_1` and `flower_2` are identical.


GriffinRichards:
It seems they're separate because the metatiles they're used to animate have different palettes. Equivalently one set of graphics could be re-exported to have the correct palette instead of them being identical.

PR #1575, opened 2021-12-17
[LEAK INFORMED] Documented the unk19 item field

LOuroboros:
## Description
This field is called "convenience button" in the original source code of Pokémon Emerald.
My personal guess is that the field was meant to signal that the Player should be allowed to register the items that have this field set to 1/`TRUE`.

As all the items from the no. 360 *(FRLG's Bicycle)* to the no. 366 *(Teachy TV)* have the flag set to "1",  I suspect that the items inserted afterward, which a Player has no reason to register for usage with the Select button with the exception of the Powder Jar, simply have it set to 1/`TRUE` by mistake, either because they were automatically generated or because whoever was in charge of writing the data of those items simply followed the lead of the items that come before.

Fun story:
In the original code, there's actually 2 fields that share the same name.
There's this field in the Items' data that Pret currently calls `unk19`, and then there's the field that Pret calls "`registeredItem`" inside of the `SaveBlock1` struct.
Both of them have the exact same name in the original source code.
That coincidence is the sole reason why I was able to come up with this theory of mine.

## **Discord contact info**
Lunos#4026


include/item.h
```diff
@@ -14,7 +14,7 @@ struct Item
     u8 holdEffectParam;
     const u8 *description;
     u8 importance;
-    u8 unk19;
+    u8 registrability;
```
GriffinRichards, line 17: There should probably be a comment here indicating that this field has no effect so that the name doesn't mislead people. Also across all repos it's only ever 0/1, so it should be a `bool8` and use `TRUE`.

PR #1579, opened 2021-12-19
Fix map *.inc script formatting inconsistencies

pineapplemachine:
## Description

Fix inconsistencies such as mismatched indentation, or missing commas to separate arguments.

I'm working on a tool for programmatically processing the decompiled source code, and fixing these formatting inconsistencies makes it easier for me to write automated tests verifying the behavior of my assembly parser. (This works by processing source -> tokens -> AST -> back to source, then compare input and output sources with comments removed and newlines normalized.) I can imagine that these changes will also make life a little easier for anyone else who is trying to analyze the contents of this repository with automated tools,

## **Discord contact info**

legume


data/maps/BattleFrontier_BattleDomeBattleRoom/scripts.inc
```diff
@@ -5,17 +5,17 @@
 .set LOCALID_PLAYER, 13
 .set LOCALID_OPPONENT, 15
 
+.set NO_DRAW, 0
+.set DRAW_TRAINER, 1
+.set DRAW_TUCKER, 2
```
pineapplemachine, line 10: Probably doesn't make sense to define assembly-time constants inside a labeled block? I moved them out and placed them with the other assembly-time constants.

Also, normalizing whitespace to always be a single space after the comma is very helpful for automated tools and arguably helps human readability and maintainability in most cases as well.

data/maps/BattleFrontier_BattlePalaceLobby/scripts.inc
```diff
@@ -472,8 +472,8 @@ BattleFrontier_BattlePalaceLobby_Text_FeatWillBeRecorded:
 
 @ Unused
 BattleFrontier_BattlePalaceLobby_Text_BattlePointsFor7WinStreak:
-    .string "For the feat of your 7-win streak,\n"
-    .string "we present you with Battle Point(s).$"
+	.string "For the feat of your 7-win streak,\n"
+	.string "we present you with Battle Point(s).$"
```
pineapplemachine, line 476: Inconsistent indentation

data/maps/BattleFrontier_Lounge3/scripts.inc
```diff
@@ -134,18 +134,18 @@ BattleFrontier_Lounge3_EventScript_LostChallenge::
 	end
 
 BattleFrontier_Lounge3_EventScript_RewardBet5::
-	buffernumberstring STR_VAR_1, (BET_AMOUNT_5 * 2)
-	setvar VAR_0x8004, (BET_AMOUNT_5 * 2)
+	buffernumberstring STR_VAR_1, BET_AMOUNT_5 * 2
+	setvar VAR_0x8004, BET_AMOUNT_5 * 2
```
pineapplemachine, line 138: Parenthesis here are extraneous and when used unnecessarily like this can make automated analysis a bit more finicky

data/maps/BattleFrontier_BattlePikeLobby/scripts.inc
```diff
@@ -31,7 +31,7 @@ BattleFrontier_BattlePikeLobby_EventScript_QuitWithoutSaving::
 	lockall
 	msgbox BattleFrontier_BattlePikeLobby_Text_FailedToSaveBeforeQuitting, MSGBOX_DEFAULT
 	closemessage
-	pike_set PIKE_DATA_WIN_STREAK 0
+	pike_set PIKE_DATA_WIN_STREAK, 0
```
pineapplemachine, line 34: Missing commas are arguably errors, not just formatting inconsistencies! If the `PIKE_DATA_WIN_STREAK` #define was ever changed to something other than a plain numeric constant, for this example, it's conceivable that the change could trigger a syntax error here. Not that this is likely to happen...but I think it's better to be safe with these things.

data/maps/BattleFrontier_Lounge3/scripts.inc
GriffinRichards: Certain compilers will interpret this incorrectly as multiple arguments without the parentheses. Please revert this.

data/maps/Route113_GlassWorkshop/scripts.inc
```diff
@@ -2,14 +2,14 @@ Route113_GlassWorkshop_MapScripts::
...
-	.set WHITE_FLUTE_PRICE,  1000
-	.set BLACK_FLUTE_PRICE,  1000
+	.set RED_FLUTE_PRICE, 500
+	.set WHITE_FLUTE_PRICE, 1000
+	.set BLACK_FLUTE_PRICE, 1000
 	.set PRETTY_CHAIR_PRICE, 6000
-	.set PRETTY_DESK_PRICE,  8000
-	.set LOWEST_ASH_PRICE,   BLUE_FLUTE_PRICE
+	.set PRETTY_DESK_PRICE, 8000
+	.set LOWEST_ASH_PRICE, BLUE_FLUTE_PRICE
```
GriffinRichards, line 12: These whitespace changes are unnecessary noise, they amount to preference. If someone is making a tool to process scripts it should be up to them to correctly parse whitespace.

pineapplemachine: Thanks for your feedback.

---

> These whitespace changes are unnecessary noise, they amount to preference. If someone is making a tool to process scripts it should be up to them to correctly parse whitespace.

As I mentioned in the PR description, the way this affects me is that it reduces my options for automated testing by comparing the input sources against output sources reconstructed from the AST. No auto-formatter will be smart enough to reproduce this stylized indentation in the same places (and only the same places), for example. It's not necessarily related to whitespace parsing. I could completely ignore whitespace after commas when comparing the sources, but then testing can tell me less about how correctly the parser/serializer was able to reconstruct the input.

Even putting aside automated tools, this stylized indentation is a practice that I personally avoid when writing code because it does come with some drawbacks even for human readability and maintenance.

You can read about some arguments for and against this kind of vertical column alignment in an /r/programming thread linked here. I fall in the camp that stylistic indentation is a maintenance burden and that the excessive whitespace harms readability in most cases rather than helping it: https://old.reddit.com/r/programming/comments/2nrs2y/why_i_vertically_align_my_code_and_you_should_too/

That being said, it is absolutely a subjective issue. The case I'd make here is that one formatting style cannot be said to be objectively better for humans than the other, but one certainly can be said to be objectively better for automated tools. This is because always having a consistent single space after commas to separate arguments means fewer, simpler rules that a tool needs to know about.

Yes, of course a parser must be able to handle multiple consecutive whitespace characters after a comma, but inconsistent formatting can still complicate analysis or verification in other ways besides parsing. Especially if formatting rules are complex, subjective, or not machine-reproducible. The arbitrary decision of in what cases to use this stylized indentation checks every one of those three qualifiers, I believe.

Look at this PR, for example! Thanks to an automated tool I wrote to process this code, I was able to rapidly spot numerous inconsistencies (some arguably errors, such as missing commas) that should not only be relevant to automated tools. Is this kind of automated code analysis not something you'd want to support and make easier for yourself and others?

---

> Certain compilers will interpret this incorrectly as multiple arguments without the parentheses. Please revert this.

If that's true, then that would be very concerning. This is the only place in any file within pokeemerald with an *.inc extension I found where a multiplication operation was placed into parentheses like this. `asm/macros/pokemon_data.inc` and `constants/m4a_constants.inc` both contain similar operations that are not enclosed within parentheses.

Compilers that would be broken by this change should have already been broken by those files, no? Or by the large number of binary arithmetic operations other than multiplication are not enclosed within parentheses? You can find examples of these in at least `asm/macros/battle_script.inc`, `berry_fix/payload/constants/gba_constants.inc`, and `pokeemerald/constants/gba_constants.inc`.

GriffinRichards: Re: whitespace, yes as we both said for humans (which I'm more concerned with) it's essentially preference. I'm not sure the ability to exactly machine-process and reproduce the scripts is a common use for this repo, but as the change doesn't really matter otherwise it's fine.

The parentheses do matter because they can and have caused errors in the past. I don't know the specifics of why it's ok in some instances, except to say that most of the macros you've referred to are not in use. In any case they should have parentheses.

PR #1596, opened 2022-01-09
Add missing array constants in field_player_avatar

GriffinRichards:


src/field_player_avatar.c
```diff
@@ -579,17 +593,17 @@ static u8 CheckMovementInputNotOnBike(u8 direction)
     if (direction == DIR_NONE)
     {
         gPlayerAvatar.runningState = NOT_MOVING;
-        return 0;
+        return PLAYER_STILL;
```
cbt6, line 596: Why not reuse `NOT_MOVING`?

GriffinRichards: Guess we could. Didn't occur to me because it's not actually using the value of `gPlayerAvatar.runningState`.

PR #1605, opened 2022-01-18
Declare gHeap in malloc.c, so it doesn't have to be explicitly mentioned in the linker script

GameCubeGBA:
As is, gHeap is not explicitly defined outside of being a nebulous extern that requires the linkerscript to know what to do with it. 
There ought to be a place where it if defined, so let's declare it in malloc.c
## Description

Because gHeap is first on the list, malloc EWRAM data must be linked first in the ROM. I changed the link order so that gflib .o files would be linked first so that the memory location of gflib code (since that is where gDecompressBuffer and gHeap are modified) is not too far away from the code accessing said arrays/buffers. I will be happy to make any changes if so needed.

## **Discord contact info**
Pizza Delivery Irida (Rose)#7522

### DIFF
--- a/gflib/malloc.c
+++ b/gflib/malloc.c
@@ -1,4 +1,7 @@
 #include "global.h"
+#include "malloc.h"
+
+EWRAM_DATA ALIGNED(4) u8 gHeap [HEAP_SIZE] = {0};
 
 static void *sHeapStart;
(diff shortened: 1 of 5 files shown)


GameCubeGBA:
Rebased due to merge conflicts in ld_script_modern.txt

PR #1606, opened 2022-01-18
Make sSpriteTileRanges a 2d array and use bitwise functions for operations on bitmaps

GameCubeGBA:
Cleanup AddSubspritesToOamBuffer too.

## Description

In short, these changes aren't simply stylistic. These changes better capture the essence of what is going on.

The program is checking for a specific bit to be set to know if the tile is allocated or not, so bitwise operations absolutely make sense here.

The 2d array gets rid of a lot of the manual index operations required to access and set data if the array were just one.

Finally, AddSubspritesToOamBuffer has been cleaned up without multiple variables hiding away or obscuring what is truly happening.

If these changes aren't ideal, you are free to make a comment and give feedback!

Thank you guys so much!

## **Discord contact info**

Pizza Delivery Irida (Rose)#7522

### DIFF
--- a/gflib/sprite.c
+++ b/gflib/sprite.c
@@ -12,22 +12,21 @@
 
 #define SET_SPRITE_TILE_RANGE(index, start, count) \
 {                                                  \
-    sSpriteTileRanges[index * 2] = start;          \
-    (sSpriteTileRanges + 1)[index * 2] = count;    \
+    sSpriteTileRanges[index][0] = start;           \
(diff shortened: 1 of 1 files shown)


GameCubeGBA:
Update: Fixed one UB case (assigning u8 to an s8) without changing behavior or affecting matching.

GameCubeGBA:
Any updates? (kindly)

PR #1608, opened 2022-01-19
Fix some incorrect fieldmap names

GriffinRichards:
Fixes a bunch of ambiguous or incorrect language referring to metatiles / map data.

- The `METATILE_` constants in `global.fieldmap.h` refer to two different data formats, ironically neither of which are actual metatiles (one refers to map 'blocks', the other to metatile attributes).
- The mask for elevation is coincidentally the same as the mask for metatile layer type, and the former is incorrectly used for both.
- `GetBehaviorByMetatileId` returns all of a metatile's attributes (i.e. the behavior and layer type), not just the behavior.
- `MapGridGetTileAt` / `MapGridBorderTileAt` get map blocks, not tiles
- The names of the `gBGTilemapBuffers` are confusing because their numbers aren't the same as their associated background, e.g. `gBGTilemapBuffers2` is assigned to bg 1.
- Adds metatile layer type constants
- Height/z coord are used interchangeably with elevation. Replaces them with the latter.
- Define the flag array in `metatile_behavior` with an explicit size so that it doesn't need to list all behaviors. Additionally drop the flag macro in favor of using the existing flag names, so you can see what properties each behavior has without referring back to the macro's definition.


src/fieldmap.c
```diff
@@ -344,44 +344,44 @@ static void FillEastConnection(struct MapHeader const *mapHeader, struct MapHead
 
 u8 MapGridGetZCoordAt(int x, int y)
```
huderlem, line 345: Change to `GetElevationAt()`?

PR #1613, opened 2022-01-23
Move M4A Track Definitions

Kurausukun:
## Description
It makes a lot more sense to define them here instead of awkwardly shoving them at the bottom of m4a_1.s, which is prone to people deleting it without noticing (which is how this PR came to be).  Also made some defines to make them easier to change. Thanks to pidgey for helping.

## **Discord contact info**
Kurausukun#9923

### DIFF
--- a/ld_script_modern.txt
+++ b/ld_script_modern.txt
@@ -27,7 +27,7 @@ SECTIONS {
         /* .bss starts at 0x3000000 */
         src/*.o(.bss);
         gflib/*.o(.bss);
-		data/*.o(.bss);
+        data/*.o(.bss);
         *libc.a:*.o(.bss*);
         *libnosys.a:*.o(.bss*);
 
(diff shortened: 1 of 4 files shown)


GriffinRichards:
You need to update `ld_script_modern.txt`, it doesn't include the `.bss` section for `data`

Kurausukun:
> You need to update `ld_script_modern.txt`, it doesn't include the `.bss` section for `data`

Fixed

ld_script_modern.txt
```diff
@@ -27,6 +27,7 @@ SECTIONS {
         /* .bss starts at 0x3000000 */
         src/*.o(.bss);
         gflib/*.o(.bss);
+		data/*.o(.bss);
         *libc.a:*.o(.bss*);
```
GriffinRichards, line 31: Tabs vs. spaces

Kurausukun: fixed

PR #1619, opened 2022-01-26
Use .mon so the right pointer type is passed

GameCubeGBA:
This happens to work because .mon is the first field in the list. However, a compiler I was using was complaining about different pointer types, so this makes the pointer type correct.

### DIFF
--- a/src/daycare.c
+++ b/src/daycare.c
@@ -341,7 +341,7 @@ static void Debug_AddDaycareSteps(u16 numSteps)
 
 u8 GetNumLevelsGainedFromDaycare(void)
 {
-    if (GetBoxMonData(&gSaveBlock1Ptr->daycare.mons[gSpecialVar_0x8004], MON_DATA_SPECIES) != 0)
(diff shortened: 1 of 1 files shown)


src/daycare.c
```diff
@@ -341,7 +341,7 @@ static void Debug_AddDaycareSteps(u16 numSteps)
 
 u8 GetNumLevelsGainedFromDaycare(void)
 {
-    if (GetBoxMonData(&gSaveBlock1Ptr->daycare.mons[gSpecialVar_0x8004], MON_DATA_SPECIES) != 0)
+    if (GetBoxMonData(&gSaveBlock1Ptr->daycare.mons[gSpecialVar_0x8004].mon, MON_DATA_SPECIES) != 0)
```
sphericalice, line 344:
```suggestion
    if (GetBoxMonData(&gSaveBlock1Ptr->daycare.mons[gSpecialVar_0x8004].mon, MON_DATA_SPECIES) != SPECIES_NONE)
```

GameCubeGBA:
Committed ! ^_^

PR #1625, opened 2022-02-07
BUGFIX: Fix roamer status field

GameCubeGBA:
Since GetMonData and SetMonData both deal with u32s when it comes to status, the status for the roamer ought to be u32. This seems to be an oversight where the devs assumed the status was a u8. This does more than fix the previous bugfix, but also the bugfix missed in pret where the status field is set to the output of GetMonData whenever UpdateRoamerHPStatus is called.

### DIFF
--- a/include/global.h
+++ b/include/global.h
@@ -569,14 +569,27 @@ struct Roamer
     /*0x08*/ u16 species;
     /*0x0A*/ u16 hp;
     /*0x0C*/ u8 level;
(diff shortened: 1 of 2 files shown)


GriffinRichards:
This is a misunderstanding of how status works. The field in the save is intentionally 8 bits, as only primary status needs to be saved. The only mistake is already fixed.

Issue #1554, opened 2021-11-16
Tracking Porymap breaking changes


GriffinRichards:
This issue is for tracking a few potential improvements that haven't been made because they would require changes to Porymap. This way if it's decided one breaking change should be made the rest can be easily referred to and piled on.

- `object_event_graphics_info.h` could be made more readable and less prone to errors from missing a field by using designated initializers (see #1413). Porymap is currently unable to parse this.
- As of #1550 files in `graphics/pokenav` are organized into subfolders, except for those read/written by Porymap whose paths would need to be updated. See  #1636 
- #1558 adds among other things constants for special warp ids. They can't be used in the `dest_warp_id` field because Porymap treats this as an int and won't properly handle macro names. See #1755
- Additional constants in `include/constants/heal_locations.h` aren't preserved by Porymap, which prevents adding `HEAL_LOCATION_NONE`. See #1671 
- Not sure how worth changing it is, but the only reason `data/tilesets.s` is still in assembly is because Porymap expects it to be. Related: #1736 
- `MAP_NONE` is more accurately `MAP_DYNAMIC`, and `MAP_UNDEFINED` would be `MAP_NONE`. This could be updated, but Porymap currently hard codes "`MAP_NONE`" as "None" and expects any other `MAP_` constant in a warp destination field to be a valid map, or it rejects the event. See #1755

GriffinRichards:
The connection direction constants are unnecessarily duplicated (in [map.inc](https://github.com/pret/pokeemerald/blob/677b4fc394516deab5b5c86c94a2a1443cb52151/asm/macros/map.inc#L91-L96) and [global.fieldmap.h](https://github.com/pret/pokeemerald/blob/7940f121f66bb0ebe0932bc642c3d9b4015f79a7/include/global.fieldmap.h#L23-L28)).
Porymap hard-codes the names of the connection constants in `map.inc`, which complicates using a single set of constants everywhere

GriffinRichards:
Reopening to track breaking changes in preparation for Porymap 6

- [ ] #2023 will update `MAP_NUM`/`MAP_GROUP` in a way that Porymap's Heal Location parsing doesn't currently handle.
- [ ] Related to the above, perhaps Heal Location data should be converted to JSON
- [ ] Porymap can stop writing some generated files (https://github.com/huderlem/porymap/issues/556) after #1949 is merged.
[...]

PR #1643, opened 2022-03-12
Format object event graphics info

cbt6:
## **Discord contact info**
cbt#8922


GriffinRichards:
Please see #1554

mrgriffin:
Note to self: I assume Porymap can't handle macros, but it would be nice if `size`, `width`, `height`, and potentially `subspriteTables` were derived from `oam` (although note that the truck has 48x48 subsprite table but a 32x32 oam). It'd also be handy if the `-mwidth` and `-mheight` of the graphics could be derived from the width and height.

GriffinRichards:
@mrgriffin If you are interested in doing that for your repo, here's a full list of dimension disagreements (had to find these anyway):


| Object  | `width`/`height` | `subspriteTables`  | `oam` | 
| ------------- | ------------- | ------------- | ------------- |
| SS Tidal | 96x40 | 96x40 | 8x8 |
| Submarine Shadow | 88x32 | 88x32 | 8x8 |
| Truck | 48x48 | 48x48 | 32x32 |
| Berry Tree | 16x16 | `NULL` | 16x16 |
| Town Map (FRLG) | 32x16 | 16x16 | 32x16 |
| SS Anne (FRLG) | 128x64 | 128x64 | 8x8 |

Additionally, all the "Normal" player avatar objects are 16x32 but have a `size` of 512, and `BerryTree` and `BerryTreeEarlyStages` are 16x16 but have a `size` of 256.

Issue #1592, opened 2022-01-06
PNG.h missing on Apple Silicon M1 (Installed everything under Rosetta)


Manurocker95:
Well, I installed devkitpro and stuff under rosetta and AGBCC didn't show any issue when compiling but when compiling pokeemerald, it says png.h is missing and I guess other tools. There was other user with the same issue but never got an answer, any idea?

![image](https://user-images.githubusercontent.com/17181899/148460713-df9cd97c-b2f4-4691-a0cd-d0f427792683.png)

GriffinRichards:
Have you installed `libpng`, per [pokeemerald's installation instructions](https://github.com/pret/pokeemerald/blob/master/INSTALL.md#macos)?

Manurocker95:
> Have you installed `libpng`, per [pokeemerald's installation instructions](https://github.com/pret/pokeemerald/blob/master/INSTALL.md#macos)?

Yep

GriffinRichards:
Then evidently it's not installed correctly. `png.h` is a header file belonging to `libpng`. I haven't tried the installation process on M1 as I don't have access to one, so I can't tell you if there are any additional steps you'd need to take to properly install it in that environment.

Manurocker95:
> Then evidently it's not installed correctly. `png.h` is a header file belonging to `libpng`. I haven't tried the installation process on M1 as I don't have access to one, so I can't tell you if there are any additional steps you'd need to take to properly install it in that environment.

I tested installing libpng under rosetta, which you can clearly see it's installed:

![image](https://user-images.githubusercontent.com/17181899/148517513-4cc18f57-39b7-4807-98ec-b8aed81b1173.png)

And natively and it's kinda weird everybody with an M1 have the same exact issue (And that's why I open this issue)

Manurocker95:
Btw, the file is is there 
![image](https://user-images.githubusercontent.com/17181899/148519247-61c04d8c-6f44-4155-b5d7-b9560736b648.png)

Manurocker95:
OMFG, there was a discord thread. Reinstalling pnglib under rosetta explicitely over the usr/local instead of regular path worked!!!

![image](https://user-images.githubusercontent.com/17181899/148522148-4d7c9292-b264-47b1-8ef7-38c2592367b7.png)

Manurocker95:
Until M1 native support ever arrives, the installing solution for mac should be added to the installing readme. We can close this already!

connor-jensen:
[...]

PR #1645, opened 2022-03-16
Fix finding libpng in gbagfx and rsfont on macOS M1

acidghost:
On macOS 12 M1, Homebrew installs libpng in a custom location.

## Description

Use `pkg-config` to find libpng library and headers.
This should generally work on any *nix system, but I only tested this on macOS 12 on M1.
This is related to #1592.

Note that I'm able to compile everything just fine natively on arm64/aarch64 with pret/agbcc#52.

## **Discord contact info**

acidghost#8150


coltongit:
Shouldn't this PR also be sent to pokefirered?

PR #1651, opened 2022-04-05
use json for region map sections

garakmon:
- update nlohmann/json and pantor/inja in tools/
  - create new inja template


coltongit:
pokeruby companion PR at https://github.com/pret/pokeruby/pull/842

PR #1652, opened 2022-04-13
Clarify when to use make compare

ExpoSeed:
After discord discussions, I clarified that compare should only be used for vanilla emerald and not for modifications like hacks. Also made some language later a bit more explicit. This should cause less confusion about make compare

Some notes: I bolded the notice to put emphasis on it, and I put it before the command so if someone stops reading at the command they will know it will fail after modifications. I also moved the note for windows later to put more emphasis on the instructions

I'm not opposed to doing the other option of removing compare and mentioning it elsewhere, but I believe make compare is a useful sanity check to verify that not only did the ROM build, it built exactly what is expected and our installation is correct. I also believe it's a valuable enough sanity check worth knowing about for any hacker (it has caught errors for me in the past), not necessarily only for contributors

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -418,20 +418,24 @@ If you aren't in the pokeemerald directory already, then **change directory** to
 ```bash
 cd pokeemerald
 ```
-To build **pokeemerald.gba** for the first time and confirm it matches the official ROM image (Note: to speed up builds, see [Parallel builds](#parallel-builds)):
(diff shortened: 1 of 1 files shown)


GriffinRichards:
As I and others suggested on Discord, let's instead move `make compare` under "Building Guidance", alongside the explanations of `make -jN`, `make DINFO=1` and `make modern`. It's not necessary for installation.

ExpoSeed:
my reasoning for having it during installation is already in the OP, though I'm not opposed to putting it in the section either

GriffinRichards:
> though I'm not opposed to putting it in the section either

:thumbsup: Let's do that

coltongit:
This LGTM

PR #1654, opened 2022-04-20
Don't Build Modern with -g by Default

Kurausukun:
## Description
The only reason -g is enabled for the regular target is for that weird nonmatching case; there's no good reason to enable debug symbols by default for the modern target, plus this makes DINFO actually useful (it does nothing currently). Also adds a note in INSTALL.md about this only being necessary for modern.

## **Discord contact info**
Kurausukun#9923

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -451,14 +451,6 @@ Replace `<output of nproc>` with the number that the `nproc` command returned.
 
 `nproc` is not available on macOS. The alternative is `sysctl -n hw.ncpu` ([relevant Stack Overflow thread](https://stackoverflow.com/questions/1715580)).
 
-## Debug info
-
-To build **pokeemerald.elf** with enhanced debug info:
-```bash
-make DINFO=1
-```
-Note that this is only necessary for the `modern` target; the regular target has debug info enabled by default.
-
 ## devkitARM's C compiler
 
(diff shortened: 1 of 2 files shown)


GriffinRichards:
I wouldn't think there's any downside to including `-g` in the modern build, right? Why have it in the default build but require `DINFO` for a modern build?

GriffinRichards:
iirc we landed on also making sure the `README` doesn't suggest `make DINFO=1`, as this would still do nothing.

Ideally setting `DINFO` would always be correctly reflected, i.e. `make DINFO=0` should be valid (and would presumably produce a non-matching build). That doesn't have to happen here though.

Kurausukun:
> iirc we landed on also making sure the `README` doesn't suggest `make DINFO=1`, as this would still do nothing.
> 
> Ideally setting `DINFO` would always be correctly reflected, i.e. `make DINFO=0` should be valid (and would presumably produce a non-matching build). That doesn't have to happen here though.

Is that not what [this change](https://github.com/pret/pokeemerald/pull/1654/files#diff-09b140a43ebfdd8dbec31ce72cafffd15164d2860fd390692a030bcb932b54a0R460) is for, or are you talking about something else

GriffinRichards:
What I mean is that the line above that tells users to do `make DINFO=1`, which does nothing.

Kurausukun:
You're right, sorry, it's been a while since I looked at this. I made some structural changes to INSTALL.md to address this

PR #1657, opened 2022-04-28
Sync battle anim args

PikalaxALT:
## Description

Syncs expansion of battle anim command args with FireRed.
References https://github.com/pret/pokefirered/pull/501

## **Discord contact info**
PikalaxALT#5823


data/battle_anim_scripts.s
```diff
@@ -2597,8 +2597,8 @@ Move_LOW_KICK:
 	end
 
 Move_EARTHQUAKE:
-	createvisualtask AnimTask_HorizontalShake, 5, (MAX_BATTLERS_COUNT + 1), 10, 50
-	createvisualtask AnimTask_HorizontalShake, 5, MAX_BATTLERS_COUNT, 10, 50
+	createvisualtask AnimTask_HorizontalShake, 5, ANIM_PLAYER_RIGHT, 10, 50
+	createvisualtask AnimTask_HorizontalShake, 5, ANIM_PLAYER_LEFT, 10, 50
```
GriffinRichards, line 2601: Where do these constant names come from? They don't correspond with what the function is doing at all.

PikalaxALT: You're right, I was looking at AnimTask_ShakeMon2 in FireRed and blindly assumed that all functions using an ANIM_X argument used the same pool of values.

PR #1668, opened 2022-05-25
Static assertion for size of save structs

mrgriffin:
The names were chosen to get this kind of error message:
```
src/save.c:76: size of array `SaveBlock2FreeSpace' is negative
```


mrgriffin:
```
src/save.c:81: sizeof applied to an incomplete type
src/save.c:81: `HALL_OF_FAME_MAX_TEAMS' undeclared here (not in a function)
```
Oops. 1 sec.

PR #1677, opened 2022-06-09
Use more TEXT_COLOR Constants

Hexoca:
## Description

Replace magic numbers found in Font/TextPrinters with the pre existing TEXT_COLOR constants

## **Discord contact info**

Hexoca#0640


GriffinRichards:
I'm generally hesitant about these text color constants, because all these values really do is indicate which color position to use and the actual colors change depending on the palette (with exception to `TEXT_COLOR_TRANSPARENT`). I imagine all the uses of the gray text colors are correct here. Are you sure about the rest?

Hexoca:
> I'm generally hesitant about these text color constants, because all these values really do is indicate which color position to use and the actual colors change depending on the palette (with exception to `TEXT_COLOR_TRANSPARENT`). I imagine all the uses of the gray text colors are correct here. Are you sure about the rest?

I agree that I also do not like the current names in the first place, as they themselves are not fully clear on what they are. I did not check if every text palette loaded matches up with the current display names, I would say its about 90-95% use the standard format of 2(DarkGray), 1(White), 3(Light Gray).  But I do know some break this standard (Like the options menu).

Would a change like

`#define TEXT_PALETTE_COLOR_1 0x1 // Usually White` 
`#define TEXT_PALETTE_COLOR_2 0x2 // Usually Dark Gray`
`#define TEXT_PALETTE_COLOR_3 0x3 // Usually Light Gray`
`Feel free to suggest a better name`

do better? sort of how the current 'TEXT_DYNAMIC_COLOR' is setup? I think something like that to replace the ambiguous numbers in the various text print calls could help make it clear what the arguments are for at a glance. While also being more descriptive on what the values actually relate to instead of the color names.

coltongit:
@Hexoca please rebase this PR to fix merge conflicts

PR #1689, opened 2022-06-18
Remove UBFIX related to task destruction

GriffinRichards:
These UBFIXs are based on misunderstandings of how tasks work. The following are not UB:
- Destroying a task that has already been destroyed. Destroyed tasks are marked inactive, and inactive tasks are ignored by `DestroyTask`.
- Reading/writing task data of a destroyed task. The memory for this data is not freed, and it will still contain the same data until a new task is created in that slot.


coltongit:
Would there be any real graphical issue because the memory's never freed and is still used by the task that just got destroyed?

PR #1696, opened 2022-06-23
Fix tools, fix all warnings, and make tools cross-compatible

coltongit:
## Description

This PR syncs tools with pokefirered to allow copies of most tools from either repo to build matching ROMs in both pokefirered and pokeemerald. This ports the FRLG code from pokefirered to mapjson and vice versa, allowing both games to use each others' tools and still compile and match perfectly.
References it's companion PR, https://github.com/pret/pokefirered/pull/509
## **Discord contact info**

chloerawr(female symbol)#5011


GriffinRichards:
Perhaps I'd have more confidence in this if you could explain or justify some of these changes beyond "fix tools." Your changes to `generate_map_events_text` at least are a misunderstanding of differences between the repos.

coltongit:
> Perhaps I'd have more confidence in this if you could explain or justify some of these changes beyond "fix tools." Your changes to `generate_map_events_text` at least are a misunderstanding of differences between the repos.

Literally the only thing that's different between the two repos' compile paths are that sign events are output a little differently between pokeemerald and pokefirered. This was tested to work across both repos by using each others' `tools` directories upon compile

GriffinRichards:
> sign events are output a little differently between pokeemerald and pokefirered

You've misunderstood the significance of that difference, and the change you've introduced for it is nonsense. Consider that you're introducing a version check in a function that isn't called by the version you're checking for.

> the two repos' compile paths

I don't know what you mean by this.

> This was tested to work across both repos by using each others' tools directories upon compile

I understand that it compiles, the workflow can already verify this. I'm looking for justification of the changes. Not all of this is syncing between the repos, there are novel changes being introduced.

coltongit:
> > sign events are output a little differently between pokeemerald and pokefirered
> 
> You've misunderstood the significance of that difference, and the change you've introduced for it is nonsense. Consider that you're introducing a version check in a function that isn't called by the version you're checking for.
> 
> > the two repos' compile paths
> 
> I don't know what you mean by this.
> 
> > This was tested to work across both repos by using each others' tools directories upon compile
> 
> I understand that it compiles, the workflow can already verify this. I'm looking for justification of the changes. Not all of this is syncing between the repos, there are novel changes being introduced.

This also makes the tools compatible with pokeruby, however pokeruby itself may need to be brought up to speed before it can take on the changes introduced by this and pokefirered's companion PRs.

GriffinRichards:
> pokeruby would hit that code line and check, because it currently uses bg_event instead of bg_sign_event like pokeemerald does, the change was necessary for pokeruby being able to work with these tools as well

The correct solution is to sync the changes to the sign macro, not to add conditionals to the other repos handling the difference in name. If anything this clutters the tools and makes editing them more confusing, which seems far more likely than anyone needing to copy the tools directory from one repo to another.

coltongit:
> > pokeruby would hit that code line and check, because it currently uses bg_event instead of bg_sign_event like pokeemerald does, the change was necessary for pokeruby being able to work with these tools as well
> 
> The correct solution is to sync the changes to the sign macro, not to add conditionals to the other repos handling the difference in name. If anything this clutters the tools and makes editing them more confusing, which seems far more likely than anyone needing to copy the tools directory from one repo to another.

Yes, that would be better, however this acts as a stopgap measure until that specific action is taken on pokeruby, which would be perhaps something to discuss in pokeruby (getting it up to speed with pokeemerald and pokefirered)

GriffinRichards:
> however this acts as a stopgap measure

It's a cluttering stopgap for a nonexistent problem. If you'd like to sync the macro names please just do that, rather than change the tools to handle different names from other repos.

If you'd like to propose these other changes to the tools, please also do so with some justification.

PR #1706, opened 2022-07-11
Add a style guide so future users know how to format their code for contribution

### LINKED ISSUE #598: Contributing.md
We should have a contributing.md file with code style guidelines.

rawr51919:
## Description

Adds a style guide so it's easy to show what is expected from future contributors and their work.
Port from https://github.com/pret/pokefirered/pull/63
Fixes #598
## **Discord contact info**

chloerawr(female symbol)#5011


STYLEGUIDE.md
```diff
@@ -0,0 +1,288 @@
...
+{
+    MyFunction();
+    return 0;
+}
+```
+
+### Control Structures
+
+When comparing whether or not a value equals `0`, don't be explicit unless the
+situation calls for it.
```
mrgriffin, line 194: Probably worth mentioning that we do `== TRUE`/`== FALSE` _all over the place_. It's bizarre to draw a distinction between `0` and `FALSE` in this way; it's exactly backward of how you'd expect drawing a distinction to shake out, e.g. we're saying to only write `==` in situations where languages like Java with a `boolean` type would not require it, and omit `==` where it would require it.

```
$ git grep '== \(TRUE\|FALSE\)' pret/master | wc -l
1038
```

mrgriffin (changes requested):
I think this style guide should mention other conventions such as:

`CB1_`/`CB2_` as a prefix for things intended for use as a MainCallback (depending on whether they're `callback1` or `callback2`).
`VBlank_` as a prefix for things intended for use as a VBlank interrupt handler.
`Task_` as a prefix for tasks, `u8 taskId` as its argument, `s16 *data = &gTasks[taskId].data;` as a local alias for data. The use of `#define tName data[X]` to label task data.
`FieldCB_` as a prefix for functions assigned to `gFieldCallback`.

I'm sure there must be others that I've forgotten.

PR #1712, opened 2022-07-23
Added a constant for NPC trainer name length

LOuroboros:
## Description
I'm submitting this on behalf of Egg.

> Egg Today at 6:20 AM
> @Lunos Feel like making a PR to pokeemerald? 😄 .trainerName field is missing a string length define and it actually should be 11 and not 12 
> https://github.com/pret/pokeemerald/blob/70577e844ddcb8092ab9e55dd0dd7993f68f4122/include/data.h#L74 
as evidenced by here and also something I found in the French version https://github.com/pret/pokeemerald/blob/e0fae879da1b773bf90fca145e047ccdb7613938/include/global.tv.h#L84

## **Discord contact info**
Lunos#4026


DizzyEggg:
@LOuroboros I guess you could also update the .unk2 field to this constant in the sGFRomHeader. As Spherical pointed out, it most likely was used for this.

LOuroboros:
> @LOuroboros I guess you could also update the .unk2 field to this constant in the sGFRomHeader. As Spherical pointed out, it most likely was used for this.

I wasn't too sure if I should, because it's also missing the usage of `TYPE_NAME_LENGTH` and `ABILITY_NAME_LENGTH` :eyes: ...

The file does have a comment pointing out where they're most likely used in the file though.

https://github.com/pret/pokeemerald/blob/master/src/rom_header_gf.c#L117-L119

What should I do, use the constant directly, or add a comment indicating that the `unk2` field is most likely the new constant added here?

GriffinRichards:
They're not used because there are more than two members with a value of 12, and we don't know which they would be. There's only one 10. It's a reasonable guess I suppose. The other name lengths in there already are known from the colosseum multiboot decomp.

src/rom_header_gf.c
```diff
@@ -111,7 +111,7 @@ static const struct GFRomHeader sGFRomHeader = {
     .mysteryEventFlag = FLAG_SYS_MYSTERY_EVENT_ENABLE,
     .pokedexCount = NATIONAL_DEX_COUNT,
     .playerNameLength = PLAYER_NAME_LENGTH,
-    .unk2 = 10,
+    .unk2 = TRAINER_NAME_LENGTH,
```
GriffinRichards, line 114: If we're assuming `unk2` is the trainer name length it should be renamed

include/global.tv.h
```diff
@@ -81,7 +81,7 @@ typedef union // size = 0x24
         /*0x01*/ bool8 active;
         /*0x02*/ u16 species;
         /*0x04*/ u8 pokemonName[POKEMON_NAME_LENGTH + 1];
-        /*0x0F*/ u8 trainerName[11];
+        /*0x0F*/ u8 trainerName[TRAINER_NAME_LENGTH + 1];
```
GriffinRichards, line 84: Despite the name, this holds a player name not an NPC trainer name.
<https://github.com/pret/pokeemerald/blob/57222323c1cadf35cf8d408c84bf6a50b7347dcf/src/tv.c#L1555>

GriffinRichards:
You can also replace `HILL_TRAINER_NAME_LENGTH` with `TRAINER_NAME_LENGTH + 1`

include/global.tv.h
LOuroboros: This variable needs to hold a size that is at least the same of `TRAINER_NAME_LENGTH` though.
Should I do `PLAYER_NAME_LENGTH + 4` or something, or should I leave it as 11? 🤔 ...

GriffinRichards: You could leave it as `+ 4` with a comment that it's larger than it needs to be, or you could split the 3 unused bytes into their own array like this
```c
        /*0x0F*/ u8 trainerName[PLAYER_NAME_LENGTH + 1];
        /*0x17*/ u8 unused[3];
```

LOuroboros: Alrighty, I went for the latter since it made more sense. Didn't really remember it was an option.

Should I rebase and force push to pack things nicely into a single commit, or should I just leave things as is? :eyes:

GriffinRichards: Up to you. GitHub also has the option to squash it into a single commit on merge if you'd like me to do that.

LOuroboros: I don't mind either way myself, so I guess I'll leave it up to you; merging as is or squashing on merge :eyes:

PR #1713, opened 2022-07-23
Change GAME_LANGUAGE to LANGUAGE_ENGLISH

DizzyEggg:
I checked this function in all of the other languages and it actually always returns LANGUAGE_ENGLISH. 
My best assumption is this function was used more of a distinction between japanese and others(possibly to get which font to use, latin or japanese)

### DIFF
--- a/src/international_string_util.c
+++ b/src/international_string_util.c
@@ -206,7 +206,7 @@ void TVShowConvertInternationalString(u8 *dest, const u8 *src, int language)
     ConvertInternationalString(dest, language);
 }
 
-// It's impossible to distinguish between Latin languages just from a string alone, so the function defaults to LANGUAGE_ENGLISH. This is the case in all of the versions  the game.
+// It's impossible to distinguish between Latin languages just from a string alone, so the function defaults to LANGUAGE_ENGLISH. This is the case in all of the versions of the game.
 int GetNicknameLanguage(u8 *str)
 {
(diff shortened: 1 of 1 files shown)


GriffinRichards:
This would benefit from an explanatory comment, or perhaps the function should be renamed. It might otherwise be confusing why `GetNicknameLanguage` can only ever return 2 of the 6/7 possible languages.

atsen-dev:
Hi, there's any way to add a language like french on this game ? Thanks

GriffinRichards:
Please keep comments on PRs related to the PR itself. There's a Discord server linked in the readme for general questions and discussion

DizzyEggg:
Added a comment @GriffinRichards

src/international_string_util.c
```diff
@@ -206,6 +206,7 @@ void TVShowConvertInternationalString(u8 *dest, const u8 *src, int language)
     ConvertInternationalString(dest, language);
 }
 
+// It's impossible to distinguish between Latin languages just from a string alone, so the function defaults to LANGUAGE_ENGLISH. This is the case in all of the versions  the game.
```
LOuroboros, line 209:
```suggestion
// It's impossible to distinguish between Latin languages just from a string alone, so the function defaults to LANGUAGE_ENGLISH. This is the case in all of the versions of the game.
```

PR #1714, opened 2022-07-23
Add constant for species length

Pokecreatorfr:
I added a constant for the length of the species names
Pokecreatorfr#2657


GriffinRichards:
`POKEMON_NAME_LENGTH` is the constant for the length of a species name. The array you've added this constant for is presumably just larger because of padding (and regardless there's no need to have a constant used only in a single array size)

Pokecreatorfr:
Is it better to do POKEMON_NAME_LENGTH +2?

GriffinRichards:
No, now that I look at the field your interpretation is also incorrect. This array doesn't store species names, it stores Pokedex category names. They are unrelated.

PR #1718, opened 2022-07-27
Cleaned by comparing evolution_scene.c with pokefirered

AsparagusEduardo:
## **Discord contact info**
AsparagusEduardo#6051


GriffinRichards:
Lets cut these case comments down. The majority of these aren't adding any new information, they're just restating the name of the case constant or describing something obvious (in one case it's even restating an existing comment a few lines below it).

AsparagusEduardo:
Ready for re-review.

PR #1722, opened 2022-08-02
Added BUGFIX for pokenav option glow on return from condition graph

aarant:
## Description
This is a bugfix for the graphical glitch that occurs with the option menu glow when returning from the PokéNav's Condition graph. This seems to occur on some emulators and possibly real hardware?

Vanilla:
Vanilla PokeNav menu

Fixed:
Fixed menu

### DIFF
--- a/src/pokenav_menu_handler_gfx.c
+++ b/src/pokenav_menu_handler_gfx.c
@@ -397,6 +397,12 @@ bool32 OpenPokenavMenuNotInitial(void)
         return FALSE;
 
     gfx->pokenavAlreadyOpen = TRUE;
+    #ifdef BUGFIX
+    // BUG: When returning from the condition graph to the menu, on some platforms the option glow may cut off
+    // midway through the screen. The condition graph changes REG_WIN0V, so we reset the window height when returning
+    // to fix this.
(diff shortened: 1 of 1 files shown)


AsparagusEduardo:
> This seems to occur on some emulators and possibly real hardware

Does it happen in mGBA? Currently is the most accurate GBA emulator, so if it happens there it's very likely it also happens in hardware.

aarant:
Yes, it occurs in mGBA, there was some discussion about this bug in PRET a while back. I believe it occurs in VBA as well?

bleakassassin:
I think what was determined was that it _does_ happen in mGBA and VBA, but it _doesn't_ happen on real hardware.

If anything, this would be a bugfix for emulators. It might not show up on real hardware due to dumb luck, but I think that behavior should be properly emulated first before adding what amounts to a workaround for emulators.

LOuroboros:
People in the Discord server stated that the issue happens in VBA, mGBA with a GBA BIOS, and in MyBoy! too.

It doesn't happen in real hardware though, as stated by RichterSnipes and Jaizu who checked on their own GBAs.

Considering all of the above, this may be worth opening a ticket for in mGBA's repository.

*And perhaps VBA-M's? MyBoy! is not open source as far as I know,*

PR #1726, opened 2022-08-05
Minor color cleanup

AsparagusEduardo:
## Description
All other uses of `COLOR_HIGHLIGHT_SHADOW` have its values on the same set of curly braces.

## **Discord contact info**
AsparagusEduardo#6051


src/daycare.c
```diff
@@ -1195,8 +1195,8 @@ static void DaycareAddTextPrinter(u8 windowId, const u8 *text, u32 x, u32 y)
     gTextFlags.useAlternateDownArrow = 0;
     printer.letterSpacing = 0;
     printer.lineSpacing = 1;
-    printer.fgColor = 2;
-    printer.bgColor = 1;
+    printer.fgColor = TEXT_COLOR_DARK_GRAY;
+    printer.bgColor = TEXT_COLOR_WHITE;
     printer.shadowColor = 3;
```
Iniquitatis, line 1200:
```suggestion
    printer.shadowColor = TEXT_COLOR_LIGHT_GRAY;
```

src/contest.c
```diff
@@ -5418,9 +5418,9 @@ static void Contest_PrintTextToBg0WindowStd(u32 windowId, const u8 *b)
     printerTemplate.letterSpacing = 0;
     printerTemplate.lineSpacing = 0;
     printerTemplate.unk = 0;
-    printerTemplate.fgColor = 15;
-    printerTemplate.bgColor = 0;
-    printerTemplate.shadowColor = 8;
+    printerTemplate.fgColor = TEXT_DYNAMIC_COLOR_6;
+    printerTemplate.bgColor = TEXT_COLOR_TRANSPARENT;
+    printerTemplate.shadowColor = TEXT_COLOR_BLUE;
```
GriffinRichards, line 5423: Please only use these constants where you're sure they correctly apply (and this is generally why I think these are bad constants, because it's not clear they only apply to a limited set of 'standard' palettes. On top of which "dynamic color" adds no information, because "this color index depends on what's in the palette" is true of every color index except transparent).

In this specific case it's wrong at least, color 8 in this palette is a light gray.

PR #1728, opened 2022-08-08
Convert MON_DATA constants into enum to ease editing and addition of fields.

AsparagusEduardo:
## Description
Seeing as it's quite common for users to modify the `BoxPokemon` struct, I figured I would convert these constants into an enum to help them remove, edit or add fields to it.

## **Discord contact info**
AsparagusEduardo#6051


mrgriffin:
Not 100% sure but I think `include/constants` only uses `#define`s so that it can be imported in `.s` files, so this new `enum` should probably go into `include/pokemon.h` instead?

Actually, I double-checked and there's one example, `include/constants/pokedex.h`:
```
pokeemerald$ git grep enum -- include/constants/ 
include/constants/pokedex.h:enum {
include/constants/pokedex.h:enum {
```
So idk :shrug:

AsparagusEduardo:
I had to separate that enum into its own header precisely because preproc couldn't handle it, as .s files use other constants in include/pokemon.h

include/pokemon.h
```diff
@@ -3,6 +3,99 @@
 
 #include "sprite.h"
 
+// Flags for Get(Box)MonData / Set(Box)MonData
```
GriffinRichards, line 6: While we're here we should probably call these values or something, given they're not really flags.

AsparagusEduardo:
Corrected comment.

PR #1732, opened 2022-08-10
Reference TMs and HMs by move instead of number

mrgriffin:
Multiple related changes aimed at making TMs and HMs easier to modify. In particular the aim is to get an error if you remove a move from the TM/HM list, and to allow you to reorder the TM/HM list without getting errors.

1. Remove the numbers from TM/HM names, e.g. `ITEM_TM_FOCUS_PUNCH` instead `ITEM_TM01_FOCUS_PUNCH`.
2. Use the named items instead of the numbered items throughout the code and scripts.
3. Automatically generate `sTMHM_Moves`.
4. Automatically generate a TM/HM bitset for the learnsets.

The implementation of `include/constants/tms_hms.h` uses [X-macros](https://en.wikipedia.org/wiki/X_Macro#Example_2_with_X_macro_as_argument).

The `TMHMLearnset` struct also automatically grows to contain more TMs, but unfortunately I couldn't convince agbcc to unroll a loop in `CanMonLearnTMHM` or `CanSpeciesLearnTMHM` that would have automatically supported larger TM sets.

In a few places I turned `ITEM_HM01_CUT` into `ITEM_HM01` and `ITEM_TM01_FOCUS_PUNCH` into `ITEM_TM01`; these instances were (imo) checking to see if a thing was a HM or TM, and didn't care specifically which move they taught.

As is customary, here is a `sed` script for generating the changes to a vanilla-formatted `src/data/pokemon/tmhm_learnsets.h`:
```
sed -i -r -e '/SPECIES_/s/ *= TMHM_LEARNSET\(TMHM\(.M.._([A-Z_]+)\)/ = { .learnset = {\n        .\1 = TRUE,/' -e '/SPECIES_/s/ *= TMHM_LEARNSET\(0\)/ = { .learnset = {\n    } }/' -e 's/^.*\| TMHM\(.M.._([A-Z_]+)\)/        .\1 = TRUE,/' -e 's/TRUE,\),/TRUE,\n    } },/' src/data/pokemon/tmhm_learnsets.h
```


mrgriffin:
This obviously breaks extended TM/HM learnsets. _But_ it's really easy to make them work again. Something like this should work:
```diff
+u32 CanMonLearnTMHM(struct Pokemon *mon, u8 tm)
+{
+   return CanSpeciesLearnTMHM(GetMonData(mon, MON_DATA_SPECIES2, 0), tm);
+}

u32 CanSpeciesLearnTMHM(u16 species, u8 tm)
{
   +int i;
    if (species == SPECIES_EGG)
    {
        return 0;
    }
-   else if (tm < 32)
-   {
-       u32 mask = 1 << tm;
-       return gTMHMLearnsets[species].as_u32s[0] & mask;
-   }
-   else
-   {
-       u32 mask = 1 << (tm - 32);
-       return gTMHMLearnsets[species].as_u32s[1] & mask;
-   }
+   for (i = 0; i < ARRAY_COUNT(gTMHMLearnsets[species].as_u32s); i++)
+   {
+       if (tm < 32)
+           return gTMHMLearnsets[species].as_u32s[i] & (1 << tm);
+       tm -= 32;
+   }
+   return 0;
}
```

GriffinRichards:
We can always squash additional commits on merge if you'd prefer it remain a single commit, it's hard to look at a PR this large when it keeps refreshing from force pushes

src/data/items.h
```diff
@@ -3513,560 +3513,560 @@ const struct Item gItems[] =
 
 // TMs/HMs
 
-    [ITEM_TM01_FOCUS_PUNCH] =
+    [ITEM_TM01] =
```
GriffinRichards, line 3516: The price and description are still linked to the specific TM/HM, so if you're aiming for fewest things to update these should probably stay move-specific. That would just leave the item name.

src/data/party_menu.h
```diff
@@ -1117,64 +1117,9 @@ static const u8 *const sUnused_StatStrings[] =
...
-    [ITEM_TM50 - ITEM_TM01] = MOVE_OVERHEAT,
-    [ITEM_HM01 - ITEM_TM01] = MOVE_CUT,
-    [ITEM_HM02 - ITEM_TM01] = MOVE_FLY,
-    [ITEM_HM03 - ITEM_TM01] = MOVE_SURF,
-    [ITEM_HM04 - ITEM_TM01] = MOVE_STRENGTH,
-    [ITEM_HM05 - ITEM_TM01] = MOVE_FLASH,
-    [ITEM_HM06 - ITEM_TM01] = MOVE_ROCK_SMASH,
-    [ITEM_HM07 - ITEM_TM01] = MOVE_WATERFALL,
-    [ITEM_HM08 - ITEM_TM01] = MOVE_DIVE,
+    FOREACH_TMHM(TMHM_MOVE)
```
GriffinRichards, line 1124: I'm a little hung-up on the loss of simplicity for this table (but I'm not sure I could give you a good example of why it should stay the way it is; this is probably fine).

GriffinRichards (commented):
1. There's a handful of number-specific TM/HM names left (e.g. `FLAG_RECEIVED_TM44`), unless you're interested I can update those.

2. I'm a little worried about dropping a blackbox of nested macros on something as likely to be edited by novices as the TM/HM list, but I think the likely roadblock at the moment (learnset size) already is one to a degree. It might be worth describing how to expand the learnset size in `tmhm_learnsets.h`, or even conditionally including your solution if the size exceeds 64.

mrgriffin:
Oh! I didn't expect this PR to see any activity! I'll try to get the changes you've suggested done over the weekend, good idea with trying to conditionally include the solution to expand learnsets—I'll see what I can do for that. Or feel free to tackle things yourself if you'd like :)

data/event_scripts.s
```diff
@@ -56,6 +56,7 @@
 	.include "asm/macros.inc"
 	.include "asm/macros/event.inc"
 	.include "constants/constants.inc"
+	.include "constants/tms_hms.inc"
```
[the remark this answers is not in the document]
mrgriffin: Good idea! I had to include `asm/macros.inc` in `data/battle_ai_scripts.s` to make this work, because otherwise `enum` isn't defined there.

src/data/items.h
mrgriffin: I'm still trying to work out a good approach to doing this. Ideally I think in this file the constants would still be `ITEM_TM01_FOCUS_PUNCH` etc etc, but I don't have a nice way to make `FOREACH_TM`/`FOREACH_HM` tag each element with a number that can be `##`ed onto the names. Perhaps I might be able to do something with `__COUNTER__`, although then I'll still want to look for a way to pad 1-9 with a leading `0`.

src/pokemon.c
```diff
@@ -6199,15 +6200,26 @@ u32 CanMonLearnTMHM(struct Pokemon *mon, u8 tm)
     {
         return 0;
     }
-    else if (tm < 32)
+
+    // Fewer than 64 moves, use GF's method (for matching).
+    if (sizeof(struct TMHMLearnset) <= 8)
```
GriffinRichards, line 6205: Clever! `CanSpeciesLearnTMHM` needs this too

data/maps/MauvilleCity_GameCorner/scripts.inc
```diff
@@ -342,10 +342,10 @@ MauvilleCity_GameCorner_EventScript_BuyTM24::
 MauvilleCity_GameCorner_EventScript_BuyTM13::
 	checkcoins VAR_TEMP_2
 	goto_if_lt VAR_TEMP_2, TM13_COINS, MauvilleCity_GameCorner_EventScript_NotEnoughCoinsForTM
```
GriffinRichards, line 344: This file still has some `TM##` names

src/data/items.h
GriffinRichards: Ok, well I think this is fine otherwise. I'm not worried about it being fully automated like that, so if it's a choice between e.g. `ITEM_TM01` and `ITEM_TM_FOCUS_PUNCH` here I'd go with the latter.

mrgriffin: Pushed this change! :)

mrgriffin:
Hopefully ready to be squashed! :heart:

PR #1735, opened 2022-08-11
Add some missing battle arena constants

GriffinRichards:
Add constants for the values in `BattleScript_ArenaDoJudgment` and the judgement icon frames. Also added an explicit size to `sMindRatings` so that every move doesn't need to be listed.


GriffinRichards:
Also, doesn't have to happen now (just to avoid some merge conflicts) but we use the spellings `Judgment` and `Judgement` interchangeably, we should pick one.

PR #1746, opened 2022-08-19
Standardize the UI code

mrgriffin:
Hopefully these changes make it easier to understand what the UI code is doing at a glance. It's largely in preparation for the idea of compile-time checking for the base blocks and palette numbers.
When referencing inside a a palette I think of it as being an ID plus an offset, e.g. `0xF0 + 3` which is represented as `BG_PLLT_ID(0xF) + 3`.

1. Write palette IDs in hex.
2. Introduce a `PLTT_SIZE_4BPP` constant loads of multiples of 16 colors, and `PLTT_SIZEOF(n)` for other sizes of loads.
3. Introduce `BG_PLLT_ID` to turn, e.g. `0xF` into `0x0F0`, and `OBJ_PLTT_ID` to turn, e.g. `0xF` into `0x1F0`. There is also `OBJ_PLTT_ID2` for matching reasons.

TODO:
1. Write oam `paletteNum`s in hex. And identify which ones should be omitted because the sprite uses a palette tag.
2. Give names to the various palette IDs, ala `BLG_WINDOW_PALETTE_NUM`.
3. Create a constant for `16`/`0x10` when steping through palettes (e.g. in `src/palette.c`), and for passing to things like `BlendPalette`.
4. Write all base blocks in hex.
5. Write explicit sizes on the various palettes in `include/graphics.h` so that we can use `sizeof` where possible.

### DIFF
--- a/gflib/sprite.c
+++ b/gflib/sprite.c
@@ -1615,7 +1615,7 @@ void LoadSpritePalettes(const struct SpritePalette *palettes)
 
 void DoLoadSpritePalette(const u16 *src, u16 paletteOffset)
 {
-    LoadPalette(src, paletteOffset + 0x100, 32);
+    LoadPalette(src, OBJ_PLTT_OFFSET + paletteOffset, PLTT_SIZE_4BPP);
 }
 
 u8 AllocSpritePalette(u16 tag)
--- a/include/gba/defines.h
+++ b/include/gba/defines.h
@@ -67,6 +67,12 @@
 #define DISPLAY_WIDTH  240
 #define DISPLAY_HEIGHT 160
 
+#define PLTT_SIZEOF(n) ((n) * sizeof(u16))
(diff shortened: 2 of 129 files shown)


mrgriffin:
I'm not opposed to them being decimal, I just picked the one I think in (which happens to be hex because so much of the code uses it).

mrgriffin:
What's your opinion on hex base blocks? I think in some instances they can be quite clean (e.g. if you're deliberately skipping over a screen), but perhaps those should also be in decimal for consistency?

~~I mean it might not matter because ideally we'd want to generate both the palette and base block constants, but...~~

GriffinRichards:
I tend to think of base blocks as hex values. If I were writing the actual palette offset I would probably also write that in hex (as most of them are written currently), it's just the palette number I think of as decimal (like numbering each block of colors as slot 0, slot 1, etc)

PR #1753, opened 2022-08-25
Added constants for GetLeadMonFriendshipScore

LOuroboros:
## Description
I had this idea while I was working on a PR for a separate project and I figured I may as well bring it here since it could help with overworld script readability :eyes:

Each constant represents a value returned by the `GetLeadMonFriendshipScore` function.
The `GE` stands for "Greater or Equal than", which is an abbreviature already seen in the overworld scripting macros used by this project, [like for example `goto_if_ge`](https://github.com/pret/pokeemerald/blob/master/asm/macros/event.inc#L1781).

## **Discord contact info**
Lunos#4026


include/constants/pokemon.h
```diff
@@ -181,6 +181,15 @@
...
 #define FRIENDSHIP_EVENT_FAINT_LARGE      8 // If opponent was >= 30 levels higher. See AdjustFriendshipOnBattleFaint
 
+// Constants for GetLeadMonFriendshipScore
+#define FRIENDSHIP_NONE   0
+#define FRIENDSHIP_GE_1   1
+#define FRIENDSHIP_GE_50  2
+#define FRIENDSHIP_GE_100 3
+#define FRIENDSHIP_GE_150 4
+#define FRIENDSHIP_GE_200 5
+#define FRIENDSHIP_MAX    6
```
AsparagusEduardo, line 191: I don't think it's clear enough what `GE` is. Here's my recommendation:
```suggestion
#define FRIENDSHIP_NONE   0
#define FRIENDSHIP_SCORE_1   1
#define FRIENDSHIP_SCORE_50  2
#define FRIENDSHIP_SCORE_100 3
#define FRIENDSHIP_SCORE_150 4
#define FRIENDSHIP_SCORE_200 5
#define FRIENDSHIP_SCORE_MAX 6
```

mrgriffin: Personally I don't mind about `GE`, and I don't particularly like using `SCORE`, because when I read it that doesn't imply the "and up"... But now that you mention it, perhaps `FRIENDSHIP_1_TO_49`, `FRIENDSHIP_50_TO_99`, etc might be more clear?
Although, idk if that is actually an improvement, e.g. here:
```diff
-goto_if_ge VAR_RESULT, FRIENDSHIP_GE_150, PacifidlogTown_House2_EventScript_GiveReturn
+goto_if_ge VAR_RESULT, FRIENDSHIP_150_TO_199, PacifidlogTown_House2_EventScript_GiveReturn
```
I suppose it's somewhat of a wash.

AsparagusEduardo: I think I like this solution the best

LOuroboros: I don't like referring to friendship by such strict values myself, but if more people prefer that, I'll do it :eyes:

LOuroboros: Alright, I renamed them :eyes:

src/field_specials.c
```diff
@@ -65,6 +65,7 @@
 #include "constants/weather.h"
 #include "constants/metatile_labels.h"
 #include "palette.h"
+#include "constants/pokemon.h"
```
GriffinRichards, line 68: This'll already be included, it's part of `global.h`

LOuroboros: Corrected :eyes:

PR #1756, opened 2022-08-26
Declarations for 2- and 3-argument GetMonData

mrgriffin:
We _definitely_ shouldn't name the functions `GetMonData` and `GetMonData2`, so this can't be merged without somebody coming up with a better suggestion.

Also, Rose on Discord suggested having something that wraps `GetMonData` for constant `MON_DATA_*` arguments, which might be nice (although the ones that are used both with `GetMonData` and `GetMonData2` would end up being pretty ugly, so perhaps that's a nicer dream than reality).

I _think_ this should match, but just to be sure I've opened a PR so that the CI system can tell me if it doesn't.


tustin2121:
Might I suggest `GetMonData` and `GetMonData_Param`?
Or perhaps `GetMonData_CopyOut` or `_Out`, or something like that that specifies that the pointer added to the signature is an out parameter?

GriffinRichards:
Is this really an improvement? It seems more likely that someone would search for the definition of e.g. `GetMonData2` and get confused when they only find a declaration (assuming they're not sure what the `alias` attribute is doing). The current declarations are bad practice but I'd think that causes people less grief than this would.

mrgriffin:
It's an improvement in the sense that it catches errors at compile time. Currently you can `GetBoxMonData()` and get some very interesting results.

But to be clear, I mostly did this to see if it was possible, it's up to others to push for its inclusion or not.

mrgriffin:
Reviving this in response to [rh-hideout#3170](https://github.com/rh-hideout/pokeemerald-expansion/pull/3170).

Much less invasive than before, all calls (except one erroneous one in `src/daycare.c`) have stayed the same, and instead we use a macro to choose between `GetMonData2` and `GetMonData3` based on the number of arguments.

I think this change does achieve type-safety, e.g. the call in `src/daycare.c` was "wrong". It worked, because `DaycareMon` starts with a `BoxPokemon` member, and so C guarantees that the previous code worked. But I think it's nicer to pass the `BoxPokemon` pointer directly?

GriffinRichards:
Could you put some similar comment in `pokemon.c`? I imagine some people will be confused when for example they can't find a definition of `GetMonData` in there, only `GetMonData2` and `GetMonData3`.

mrgriffin:
I've added some wording that I hope is useful :)

Feel free to squash this if you merge it.

PR #1758, opened 2022-08-28
Soft-Boiled label style

AsparagusEduardo:
## Description
Like Solar Beam, Soft-Boiled changed its spelling in Gen 6, so I updated the labels to reflect this.

## **Discord contact info**
AsparagusEduardo#6051


GriffinRichards:
This leaves some inconsistencies, e.g. if the camel case is `SoftBoiled` then the uppercase should be `SOFT_BOILED`. Personally I prefer using the Gen 3 name given this is a Gen 3 repo (as in keeping `SOFTBOILED` one word), but I think I'm in the minority there.

AsparagusEduardo:
That's fair. In the context of this repo it should be consistent.

PR #1768, opened 2022-09-06
[debug] add support for mgba printf

SBird1337:
* adds support for mgba printf debugging as well as adding support for switching between debugging configuration
* adds `mini_printf` as an alternative to libc printf as well as switches to choose a pretty printing handler
* adds a pretty printing format to `mini_printf` to print preproc encoded strings

## Description
The codebase already included support to optionally enable various debugging features (NoCashPrint, AGBPrint) - this PR will add a handler for mgba logs, as well as unify the implementations by adding compile time switches to choose one of them.

Additionally it adds a pretty printing implementation (<https://github.com/mludvig/mini-printf>) - This is mainly due to the fact that some versions of libc don't like to link against the standard pretty printing implementations (dkp).

It also allows for configurable custom pretty printing formats. This PR includes a format (`%S`) to print preproc/pokehex formatted strings to the various logs.

## **Discord contact info**
Karathan#1337

### DIFF
--- a/include/gba/isagbprint.h
+++ b/include/gba/isagbprint.h
@@ -17,8 +17,8 @@ void MgbaClose(void);
 void MgbaPrintf(const char *pBuf, ...);
 void MgbaAssert(const char *pFile, s32 nLine, const char *pExpression, bool32 nStopProgram);
 #define DebugPrintf(pBuf, ...) MgbaPrintf(pBuf, __VA_ARGS__)
-#define AGB_ASSERT(exp) (exp) ? ((void*)0) : MgbaAssert(__FILE__, __LINE__, #exp, 1)
-#define AGB_WARNING(exp) (exp) ? ((void*)0) : MgbaAssert(__FILE__, __LINE__, #exp, 0)
+#define AGB_ASSERT(exp) (exp) ? ((void*)0) : MgbaAssert(__FILE__, __LINE__, #exp, TRUE)
+#define AGB_WARNING(exp) (exp) ? ((void*)0) : MgbaAssert(__FILE__, __LINE__, #exp, FALSE)
 
 // Not used in this configuration
 #define AGBPrintfInit()
@@ -26,8 +26,8 @@ void MgbaAssert(const char *pFile, s32 nLine, const char *pExpression, bool32 nS
 void NoCashGBAPrintf(const char *pBuf, ...)
 void NoCashGBAAssert(const char *pFile, s32 nLine, const char *pExpression, bool32 nStopProgram)
 #define DebugPrintf(pBuf, ...) NoCashGBAPrintf(pBuf, __VA_ARGS__)
-#define AGB_ASSERT(exp) (exp) ? ((void*)0) : NoCashGBAAssert(__FILE__, __LINE__, #exp, 1);
-#define AGB_WARNING(exp) (exp) ? ((void*)0) : NoCashGBAAssert(__FILE__, __LINE__, #exp, 0)
+#define AGB_ASSERT(exp) (exp) ? ((void*)0) : NoCashGBAAssert(__FILE__, __LINE__, #exp, TRUE);
+#define AGB_WARNING(exp) (exp) ? ((void*)0) : NoCashGBAAssert(__FILE__, __LINE__, #exp, FALSE)
 
 // Not used in this configuration
 #define MgbaOpen()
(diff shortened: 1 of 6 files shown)


include/gba/isagbprint.h
```diff
@@ -1,50 +1,58 @@
...
-void AGBPutc(const char cChr);
-void AGBPrint(const char *pBuf);
+#if (LOG_HANDLER == LOG_HANDLER_MGBA_PRINT)
+bool32 MgbaOpen(void);
+void MgbaClose(void);
+void MgbaPrintf(const char *pBuf, ...);
+void MgbaAssert(const char *pFile, s32 nLine, const char *pExpression, bool32 nStopProgram);
+#define DebugPrintf(pBuf, ...) MgbaPrintf(pBuf, __VA_ARGS__)
+#define AGB_ASSERT(exp) (exp) ? ((void*)0) : MgbaAssert(__FILE__, __LINE__, #exp, 1)
+#define AGB_WARNING(exp) (exp) ? ((void*)0) : MgbaAssert(__FILE__, __LINE__, #exp, 0)
```
AsparagusEduardo, line 21: I think these should use TRUE or FALSE for `nStopProgram` uses.

include/gba/isagbprint.h
```diff
@@ -1,50 +1,58 @@
...
+#define MgbaOpen()
+#define MgbaClose()
 #endif
-
-#undef AGB_WARNING
-#ifdef NDEBUG
-#define AGB_WARNING(exp)
-#else
-#define AGB_WARNING(exp) (exp) ? ((void *)0) : AGBAssert(__FILE__, __LINE__, #exp, 0);
 #endif
```
GriffinRichards, line 48: This whole conditional block can be made a lot shorter/easier to read. Declarations for functions not used by each configuration can remain the same outside the conditional. The only thing that needs to change depending on configuration is the definition of `DebugPrintf`, `AGB_ASSERT`, and `AGB_WARNING`.

include/gba/isagbprint.h
```diff
@@ -1,50 +1,58 @@
...
+#else // Default to AGBPrint
 void AGBPrintf(const char *pBuf, ...);
-void AGBPrintFlush1Block(void);
-void AGBPrintFlush(void);
 void AGBAssert(const char *pFile, int nLine, const char *pExpression, int nStopProgram);
-#endif
+void AGBPrintInit(void);
+#define DebugPrintf(pBuf, ...) AGBPrintf(const char *pBuf, ...)
+#define AGB_ASSERT(exp) (exp) ? ((void*)0) : AGBAssert(__FILE__, __LINE__, #exp, 1)
+#define AGB_WARNING(exp) (exp) ? ((void*)0) : NoCashGBAAssert(__FILE__, __LINE__, #exp, 0)
```
GriffinRichards, line 42: `AGBAssert` for `AGB_WARNING`

include/gba/isagbprint.h
```diff
@@ -1,50 +1,58 @@
...
 #define    AGB_ASSERT_EX(exp, file, line)
 #else
-#define    AGB_ASSERT_EX(exp, file, line) (exp) ? ((void *)0) : AGBAssert(file, line, #exp, 1);
-#endif
-
-#ifdef NDEBUG
-#define    AGB_WARNING_EX(exp, file, line)
-#else
-#define    AGB_WARNING_EX(exp, file, line) (exp) ? ((void *)0) : AGBAssert(file, line, #exp, 0);
+#define    AGB_ASSERT_EX(exp, file, line) AGB_ASSERT(exp);
```
GriffinRichards, line 55: Surely this depends on the fact that `AGB_ASSERT_EX` isn't used by pokeemerald? If it were needed for matching (as it is in pokefirered) this wouldn't work because it's discarding the hardcoded file name and line number.

SBird1337: yeah, you're right, I can either remove it alltogether since I don't see why it would be needed or add it back in with the hardcoded parameters.

GriffinRichards: Up to you, yeah it would only be necessary if we had some Emerald debug build we were trying to match (in which case we could just add the macro back from pokefirered)

include/gba/isagbprint.h
SBird1337: this is supposed to be the AGBPrintf version of the function anyway, must be a copy & paste error.

include/gba/isagbprint.h
SBird1337: 12670e8e3ee0ca49dd88d36e83ec698bfc7b227f

include/gba/isagbprint.h
SBird1337: 44a5a8d31e5920c7f2eab588f772dd13ff7ff5f1

include/gba/isagbprint.h
SBird1337: 22329d3318cde3f07ed5401226c0df2b5d951957

PR #1787, opened 2022-09-20
Moved MAX_MONEY to constants/global.h

LOuroboros:
## Description
I noticed that there was a handy constant label for the max. amount of money the Player can have, but that it was inside a .c file for some reason, so I decided to move it to a nicer location where it's easier to see and that allows to make use of it in other files easily too :eyes:

## **Discord contact info**
Lunos#4026


GriffinRichards:
Constants are regularly defined in C files when they are only used in that file. I don't think we should be moving constants to `global.h` just in case people want to use it elsewhere.

LOuroboros:
Alright then, I suppose no one else cares either since there haven't been any comments, so I'll just close it and keep it to myself :eyes:

PR #1790, opened 2022-09-25
Gender icon in naming screen based on Pokémon name length

AsparagusEduardo:
## Description
For comparison: 10, 11, 12 and 14
image image image image

## **Discord contact info**
AsparagusEduardo#6051


GriffinRichards:
For whatever it's worth it should really be `sNamingScreen->template->maxChars`, but that wouldn't match (and this is accurate for vanilla)

PR #1795, opened 2022-10-04
Remove include/tilesets.h

cbt6:
## **Discord contact info**
cbt#8922


GriffinRichards:
Ironic given the current conversation in Discord, but this will just end up getting reintroduced when/if `tilesets.s` is converted to C

PR #1804, opened 2022-10-16
Fix weather color map documentation

huderlem:
"Gamma shifts" were a completely incorrect way of describing the palette effects used in the weather system.  Instead, they are 
 better described as "brightness & contrast colors maps".


src/overworld.c
```diff
@@ -813,7 +813,7 @@ void LoadMapFromCameraTransition(u8 mapGroup, u8 mapNum)
     LoadSecondaryTilesetPalette(gMapHeader.mapLayout);
 
     for (paletteIndex = 6; paletteIndex < 13; paletteIndex++)
```
GriffinRichards, line 815: While you're in the neighborhood...
```suggestion
    for (paletteIndex = NUM_PALS_IN_PRIMARY; paletteIndex < NUM_PALS_TOTAL; paletteIndex++)
```

PR #1806, opened 2022-10-18
Renamed battle script end commands into distinct names

AsparagusEduardo:
## Description
```
end -> endmove
end2 -> endscript
end3 -> endreturn
```

## **Discord contact info**
AsparagusEduardo#6051


GriffinRichards:
I'm not sure about `endreturn`, though I don't have an alternate proposal. It's effectively just an extension of `return`, it doesn't really have anything more in common with the `end` commands

PR #1809, opened 2022-10-22
Include VSCode C/C++ extension settings to minimize disk space used for indexing

AsparagusEduardo:
## Description
These are settings that I've been using for the VSCode C/C++ extension that has helped me for a while now.
Without these, an index was created in the .vscode folder that a lot of the time was Gigabytes in size.

## **Discord contact info**
AsparagusEduardo#6051


GriffinRichards:
What is the motivation for which files are listed in `files.associations`? It seems like a random selection of headers (and includes files which presumably exist in your local project but not in this repo)

SBird1337:
This seems to be a bit random in general. For once the codebase is written in c89 standard, not c99.

Configuration files for IDEs are also a little bit user-centric by design. You'll want to specify your compiler for vscode to not lose its complete mind and you may also want to setup debugging interfaces that you can't just include in a public repository here because everybody does things a little differently.

AsparagusEduardo:
> the codebase is written in c89 standard, not c99.

Will fix

> Configuration files for IDEs are also a little bit user-centric by design.

I wanted to introduce a default configuration that will help new users not to waste unnecessary disk space and CPU to index stuff. Advanced users can modify it past that if they wish.

AsparagusEduardo:
Changes applied. Ready for re-review.

PR #1810, opened 2022-10-24
Renamed gBaseStats to gSpeciesInfo

AsparagusEduardo:
## Description
"Base Stats" is a misnomer for the data that's being stored in that struct, as the only fields which would actually be considered Base Stats are HP, Attack, Defense, Special Attack, Special Defense and Speed.

## **Discord contact info**
AsparagusEduardo#6051


src/pokemon.c
```diff
@@ -2806,7 +2806,7 @@ static u16 CalculateBoxMonChecksum(struct BoxPokemon *boxMon)
 
 #define CALC_STAT(base, iv, ev, statIndex, field)               \
 {                                                               \
-    u8 baseStat = gBaseStats[species].base;                     \
+    u8 baseStat = gSpeciesInfo[species].base;                     \
```
GriffinRichards, line 2809: Alignment

GriffinRichards (commented):
Personally I didn't think `BaseStats` was misleading enough to warrant changing, but I didn't see anyone really defending the name.

`base_stats.h` and `OLD_UNOWN_BASE_STATS` would also need to be renamed.

AsparagusEduardo:
Applied review changes. In any case, I'm open to other name suggestions.

PR #1811, opened 2022-10-25
Cleaned up Bard Sounds

AsparagusEduardo:
## Description
Noticed that most data for Bard Sounds could be simplified, so I did :P

## **Discord contact info**
AsparagusEduardo#6051


src/data/bard_music/bard_sounds.h
```diff
@@ -1,7 +1,10 @@
 #ifndef GUARD_BARD_SOUNDS_TABLE_H
 #define GUARD_BARD_SOUNDS_TABLE_H
 
-#define NULL_BARD_SOUND { 0xff }
+#define BARD_SOUND_FULL(var00, var01, var02) { var00, var01, var02, 0x0, 0x0 }
+#define BARD_SOUND(var00)   BARD_SOUND_FULL(var00, 0x0, 0x0)
```
GriffinRichards, line 5: We don't need a bunch of macros for this kind of thing (especially if they don't really clarify what's happening). 0's will be filled automatically, you can if you want just remove the 0's, e.g. this
```c
{ 1, 2, 3, 0, 0},
{ 4, 0, 0, 0, 0},
```
will match as this
```c
{ 1, 2, 3 },
{ 4 },
```

AsparagusEduardo: ...you're right.
Would you prefer if I use that format to clean up the bard sounds? (except in hexadecimal)

GriffinRichards: I'd prefer it to the macros yeah, up to you if you prefer that over the current format. I guess ideally it'd look something like this
```c
{ .var00 = 1, .var01 = 2, .var02 = 3 }
{ .var00 = 1 }
```
but with meaningful names (except `var02` I suppose, which is never used)

AsparagusEduardo:
Ready for re-review

PR #1824, opened 2022-12-05
Species animation tables into macros

AsparagusEduardo:
## Description
Converted species animation tables into macros.

(Open to suggestions, as I'm not sure how much I can continue to streamline this)

## **Discord contact info**
AsparagusEduardo#6051


src/data/pokemon_graphics/front_pic_anims.h
```diff
@@ -1,7396 +1,1928 @@
...
+ANIM_BACK_AND_FORTH_4_TIMES(Kingler_1, 5, 15, 10, 15, 25, 10, 10, 10, 10);
+ANIM_SWITCH_BACK_AND_FORTH_2_TIMES(Voltorb_1, 4, 25);
+ANIM_SWITCH_BACK_AND_FORTH_2_TIMES(Electrode_1, 8, 30);
+ANIM_CONSTANT_BACK_AND_FORTH_4_TIMES(Exeggcute_1, 10);
+ANIM_2NDFRAME_1STFRAME_1_TIME(Exeggutor_1, 35, 10);
+ANIM_2NDFRAME_1STFRAME_3_TIMES_CONSTANT(Cubone_1, 14);
+ANIM_2NDFRAME_1STFRAME_3_TIMES_CONSTANT(Marowak_1, 14);
+ANIM_2NDFRAME_1STFRAME_1_TIME(Hitmonlee_1, 15, 20);
+ANIM_2NDFRAME_1STFRAME_4_TIMES(Hitmonchan_1, 5, 5, 5, 5, 5, 15, 25, 10);
+ANIM_BACK_AND_FORTH_1_TIME(Lickitung_1, 28, 32, 15);
```
GriffinRichards, line 575: I appreciate that this might be something you prefer for your own workflow, but I think having 40+ macros for what are effectively arbitrary animations makes this file unexpectedly difficult to read and understand. Yes the original has more lines, but it at least looks like every other animation set up in the repo.

AsparagusEduardo: What if I only submit the `gMonFrontAnimsPtrTable`, `SINGLE_ANIMATION` and `DOUBLE_ANIMATION` macros? would that be better?

AsparagusEduardo: Well, honestly I did that due to the lack of consistency with other similar tables like `gMonShinyPaletteTable` where it has these macros:
```
    SPECIES_SHINY_PAL(BULBASAUR, gMonShinyPalette_Bulbasaur),
```

AsparagusEduardo: Would it be better if it's made into `ANIM_CMD(BULBASAUR, sAnims_Bulbasaur),` ?

GriffinRichards: The macro isn't doing anything useful then. If you want to list the full symbol name, just write it out normally without a macro:
```c
[SPECIES_BULBASAUR] = sAnims_Bulbasaur,
```

PR #1827, opened 2022-12-06
[fix] DebugPrintf with no format arguments

SBird1337:
Fixes `DebugPrintf` with no format arguments.

## Description
Using `DebugPrintf("Hello");` like that would error due to a malformation of `__VA_ARGS__` - this uses `##` escape which works on `cpp` and apparently also `clang` to remove the Comma `,` character if `__VA_ARGS__` is not set.

## **Discord contact info**
Karathan#1337

### DIFF
--- a/include/gba/isagbprint.h
+++ b/include/gba/isagbprint.h
@@ -29,17 +29,17 @@ void AGBPrintInit(void);
 
 #if (LOG_HANDLER == LOG_HANDLER_MGBA_PRINT)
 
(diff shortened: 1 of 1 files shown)


SBird1337:
After thinking about it for a bit I also added a `DebugPrintfLevel` macro which lets you choose the log level you desire to print to. If the platform does not support log levels, this defaults to `DebugPrintf`, on `mGBA` it supports the log levels defined in `isagbprint.h`.

PR #1828, opened 2022-12-13
Added MOVE_EFFECT_NONE

LOuroboros:
## Description
While reading Tri-Attack's move effect and looking at the relevant parts of the code, I brought up a doubt that I was having and that GriffinR helped me with. He taught me that a status flag is chosen for the effect straight the `sMoveEffectBS_Ptrs` array.

The thing is that at a glance, `MOVE_EFFECT_SLEEP` is at the top of the list, which is a constant label with a value of 1.
That can be confusing if you're expecting it to hold a value of 0 as the first member of the array *(to be honest, I guess in itself that's kind of silly. Just because it's at the top of an array it doesn't mean its value is going to be 0.)*

Thus, I feel it would be a good idea to add a `MOVE_EFFECT_NONE` to make the codee clearer at a glance.
By doing this consistency is kept with how other constant labels are laid out too though. There's `SPECIES_NONE`, `ITEM_NONE` and `ABILITY_NONE` for example.

## **Discord contact info**
Lunos#4026

### DIFF
--- a/include/constants/battle.h
+++ b/include/constants/battle.h
@@ -231,6 +231,7 @@
 #define B_WEATHER_ANY                 (B_WEATHER_RAIN | B_WEATHER_SANDSTORM | B_WEATHER_SUN | B_WEATHER_HAIL)
 
 // Move Effects
+#define MOVE_EFFECT_NONE                0
 #define MOVE_EFFECT_SLEEP               1
 #define MOVE_EFFECT_POISON              2
 #define MOVE_EFFECT_BURN                3
--- a/src/battle_script_commands.c
+++ b/src/battle_script_commands.c
@@ -2196,17 +2196,17 @@ u8 GetBattlerTurnOrderNum(u8 battlerId)
     return i;
 }
 
-#define INCREMENT_RESET_RETURN                  \
-{                                               \
-    gBattlescriptCurrInstr++;                   \
-    gBattleCommunication[MOVE_EFFECT_BYTE] = 0; \
-    return;                                     \
+#define INCREMENT_RESET_RETURN                                 \
+{                                                              \
+    gBattlescriptCurrInstr++;                                  \
+    gBattleCommunication[MOVE_EFFECT_BYTE] = MOVE_EFFECT_NONE; \
+    return;                                                    \
 }
 
-#define RESET_RETURN                            \
-{                                               \
-    gBattleCommunication[MOVE_EFFECT_BYTE] = 0; \
-    return;                                     \
+#define RESET_RETURN                                           \
+{                                                              \
+    gBattleCommunication[MOVE_EFFECT_BYTE] = MOVE_EFFECT_NONE; \
+    return;                                                    \
 }
 
 void SetMoveEffect(bool8 primary, u8 certain)
@@ -2487,7 +2487,7 @@ void SetMoveEffect(bool8 primary, u8 certain)
         }
         else if (statusChanged == FALSE)
         {
-            gBattleCommunication[MOVE_EFFECT_BYTE] = 0;
+            gBattleCommunication[MOVE_EFFECT_BYTE] = MOVE_EFFECT_NONE;
             gBattlescriptCurrInstr++;
             return;
         }
@@ -2876,7 +2876,7 @@ void SetMoveEffect(bool8 primary, u8 certain)
         }
     }
 
-    gBattleCommunication[MOVE_EFFECT_BYTE] = 0;
+    gBattleCommunication[MOVE_EFFECT_BYTE] = MOVE_EFFECT_NONE;
 }
 
 static void Cmd_seteffectwithchance(void)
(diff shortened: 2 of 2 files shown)


GriffinRichards:
I think it should still be left out of `sStatusFlagsForMoveEffects`; it's a sparse array, it's excluding many other move effects already. It should be expected that the indexes are not at their numbered position.

Fair enough to replace `[0] =` though. If we have a constant for this, it should be used at least in the rest of this file where the move effect is set to 0.

LOuroboros:
> I think it should still be left out of `sStatusFlagsForMoveEffects`; it's a sparse array, it's excluding many other move effects already. It should be expected that the indexes are not at their numbered position.

I understand your point, but I think making crystal clear for the user that `MOVE_EFFECT_SLEEP` is not the first move effect code-wise is more important. Besides, `sMoveEffectBS_Ptrs` is also excluding many move effects. Neither of the 2 arrays take every single `MOVE_EFFECT` into account.

I pushed a commit making use of the new constant wherever it seemed correct to use it though :eyes:

GriffinRichards:
The 0th element of `sMoveEffectBS_Ptrs` does not have a value of 0, though. It can't be excluded.

For `sStatusFlagsForMoveEffects`, to put it another way: why is it important that you list `MOVE_EFFECT_NONE` and not also say `MOVE_EFFECT_TRI_ATTACK`? The same argument could be made that it's "important" to know the values of `MOVE_EFFECT_FLINCH` and `MOVE_EFFECT_UPROAR` are not adjacent (and I would argue it's not important, and it's easier to read by not listing all of the irrelevant elements that aren't used)

LOuroboros:
> why is it important that you list `MOVE_EFFECT_NONE` and not also say `MOVE_EFFECT_TRI_ATTACK`?

Because a user can easily be misled into thinking that the very first move effect in that table, without looking at any of the effects below it, is the very first entry; the 0th `MOVE_EFFECT`.
Clearing up the fact that it's not helps understand the code that makes use of the first few `MOVE_EFFECTS` more easily by making it so the user won't have to go out of their way to look at the constants defined inside `include/constants/battle.h` when they can just look at the arrays in which they're used right here, in `src/battle_script_commands.c`.

It's less hassle for the user imo.

> The same argument could be made that it's "important" to know the values of MOVE_EFFECT_FLINCH and MOVE_EFFECT_UPROAR are not adjacent

I don't think so? Why would it be?

GriffinRichards:
In what situation with this table is the position of first element relevant and not the rest?

LOuroboros:
Reading the effect of Tri-Attack, which you yourself helped me with a few days back.

One of the things I was misunderstanding while reading how its effect is handled is how it makes use of `Random()` to look for a `MOVE_EFFECT`, because one of the mistakes I was making while reading it is thinking that `MOVE_EFFECT_SLEEP` was the 0th move effect, thus things didn't add up to me.

When I went to check the values of the `MOVE_EFFECT` labels, I understand that `MOVE_EFFECT_SLEEP` was the `MOVE_EFFECT` Number 1, not the `MOVE_EFFECT` Number 0.

I was thinking that `MOVE_EFFECT_SLEEP` (0) +3 would make `sStatusFlagsForMoveEffects` return `STATUS1_FREEZE`, `MOVE_EFFECT_POISON` (1) + 3 would make it return `STATUS1_PARALYSIS` and `MOVE_EFFECT_BURN` would make it return `STATUS1_TOXIC_POISON`.
That misunderstanding came from the fact that as soon as I saw `MOVE_EFFECT_SLEEP` at the top of each array, my brain thought "Oh, this is the first `MOVE_EFFECT`, so it must be the 0th, right?" and so on.
*Until again, I checked and realized later on that it's actually the `MOVE_EFFECT` 1 and in order to understand Tri-Attack's effect I had to think an invisible 0 into account.*

LOuroboros:
I accept this may be a subjective thing and that I was the one who misread the arrays and was not aware of the right numbering of the `MOVE_EFFECT` constants, and thus that I'm the one who was at fault.

My point is that it would have been simpler and easier to get things right from the get go if it was made clear that the very first `MOVE_EFFECT` of the array is actually not the 0th but the 1st.
*Which is confusing when a majority of other arrays always start at 0 imo.*

GriffinRichards:
It sounds like your problem was "I saw this was the first listed index, therefore I assumed its value was 0".

For one, it's not the first element in the table right below that (`sMoveEffectBS_Ptrs`), and for two, why is that argument only relevant for 0? e.g. "I saw `MOVE_EFFECT_FLINCH` was the 8th element in `sStatusFlagsForMoveEffects`, so I assumed its value was 7"?

LOuroboros:
> It sounds like your problem was "I saw this was the first listed index, therefore I assumed its value was 0".

It was. I said as much myself.

> For one, it's not the first element in the table right below that (sMoveEffectBS_Ptrs)

Fair enough.

> and for two, why is that argument only relevant for 0? e.g. "I saw MOVE_EFFECT_FLINCH was the 8th element in sStatusFlagsForMoveEffects, so I assumed its value was 7"?

I mean, it's not that it's only relevant for 0 per se, I'm just speaking about my personal usage experience here.

I don't have any other arguments to provide. I think it would help readability, but it seems we're not in the same wavelength here, so I'm just going to take a step back and close this PR. Thanks and sorry for bothering.

PR #1835, opened 2022-12-29
Fix ANIM_SPRITE on regular builds with BUGFIX.

Jaizu:
Dicussion about this being a BUGFIX define instead limiting it only to modern. Credit SBird.

## **Discord contact info**
Jaizu#0001


GriffinRichards:
This reintroduces the bug to modern builds, as `BUGFIX` is (intentionally) not required by modern.

PR #1837, opened 2023-01-06
event-object: Rewrite SetObjectEventDirection

AreaZeroArven:
agbcc now compiles this file matching
without the strange assignment to the s8

As for why it did not before, I cannot answer

Signed-off-by: Arven

### DIFF
--- a/src/event_object_movement.c
+++ b/src/event_object_movement.c
@@ -2341,12 +2341,10 @@ u8 CreateCopySpriteAt(struct Sprite *sprite, s16 x, s16 y, u8 subpriority)
 
 void SetObjectEventDirection(struct ObjectEvent *objectEvent, u8 direction)
 {
-    s8 d2;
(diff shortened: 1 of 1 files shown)


src/event_object_movement.c
```diff
@@ -2341,12 +2341,10 @@ u8 CreateCopySpriteAt(struct Sprite *sprite, s16 x, s16 y, u8 subpriority)
 
 void SetObjectEventDirection(struct ObjectEvent *objectEvent, u8 direction)
 {
-    s8 d2;
     objectEvent->previousMovementDirection = objectEvent->facingDirection;
     if (!objectEvent->facingDirectionLocked)
     {
```
LOuroboros, line 2346: The curly brackets here can be removed. The project doesn't lay out a scope for `if` statements that execute 1 single operation.

PR #1838, opened 2023-01-06
event-object: simplify SetObjectSubpriorityByElevation

AreaZeroArven:
As is, the current function is hard to understand.

This change matches while also making the intention clear.

Signed-off-by:Arven


AZero13:
You can close this without merge. It is stale.

PR #1850, opened 2023-01-23
Sync trade

GriffinRichards:


src/trade.c
```diff
@@ -2004,62 +2076,57 @@ static void DrawTradeMenuPartyMonInfo(u8 whichParty, u8 monIdx, u8 x, u8 y, u8 w
...
     }
 }
 
-static void PrintNicknamesForTradeMenu(void)
+static void PrintTradePartnerPartyNicknames(void)
 {
     rbox_fill_rectangle(1);
-  //PrintPartyNicknamesForTradeMenu(TRADE_PLAYER); ?
-    PrintPartyNicknamesForTradeMenu(TRADE_PARTNER);
+  //PrintPartyNicknames(TRADE_PLAYER); ?
```
sphericalice, line 2116: What does this comment mean?

GriffinRichards: That's pre-existing. But presumably wondering if the commented line should be there

AsparagusEduardo: Kinda looks like a bugfix suggestion

PR #1866, opened 2023-02-25
Whitelist instead of Blacklist for TOOLDIRS in Makefiles

AsparagusEduardo:
## Description
Originally by @mrgriffin.

When switching from a branch with additional tools back to vanilla pokeemerald, the those tools' folders lingered, meaning that it caused issues to build.
This PR fixes that by explicitely defining the tool folders, while also make it unnecesary to define an exception for poryscript.

## **Discord contact info**
AsparagusEduardo#6051

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,8 @@ JSONPROC := tools/jsonproc/jsonproc$(EXE)
 
 PERL := perl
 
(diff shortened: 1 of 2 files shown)


AsparagusEduardo:
Re-wrote the description, as the first on I did it in a rush.

PR #1870, opened 2023-03-12
Voice group development

Bassoonian:
Reorganises the voice groups (which are unhelpfully called voicegroup000.inc to voicegroup190.inc), giving them more meaningful names and organising them into folders. Also minorly changes mid2agb to allow for strings as voice group names rather than only integers.

## Description
Still a rough early draft, so feel free to disregard this for the time being

## **Discord contact info**
Jasper#5206


Kurausukun:
Some criticisms:

- I do not agree with the "instruments" folder for multiple reasons. We do not need even more subfolders in my opinion, and being that this is a change for readability, I don't think it does its job. Given that all voicegroups are definitions of instruments, adding a folder for certain voicegroups and labelling it as "instruments" is just confusing.
- Instead, the voicegroups currently in the "instruments" subfolder should be moved back in with the rest of them and given descriptive names such as `voicegroup_trumpet`, `voicegroup_strings`, etc. For the two main drumsets, I propose `voicegroup_drumset_rs` and `voicegroup_drumset_frlg`, as that is how they are used; for any incidental drumsets used for songs, we can use a similar naming scheme of `voicegroup_drumset_songname`, replacing `songname` with the appropriate song.

PR #1876, opened 2023-03-24
Add macros for map script terminators

GriffinRichards:
This would be a reasonable time to rename `map_script_2` to something descriptive like `map_script_table`, but it might not be worth the conflicts.

This PR won't disrupt poryscript or any new map scripts downstream.


martingaldeca (approved):
I'm new checking code in this repo (so not sure if I'm able to approve anything), but it makes sense to me

PR #1880, opened 2023-04-04
Renamed AllocZeroed to Calloc

AsparagusEduardo:
## Description
This came up in a pokeplatinum review. Since `Alloc` exists as a GF equivalent to `alloc`, it makes sense that `AllocZeroed` should be called `Calloc` instead.

## **Discord contact info**
AsparagusEduardo#6051

### DIFF
--- a/gflib/malloc.c
+++ b/gflib/malloc.c
@@ -128,7 +128,7 @@ void FreeInternal(void *heapStart, void *pointer)
     }
 }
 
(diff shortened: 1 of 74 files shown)


mrgriffin:
I don't have a particularly strong opinion either way, but it's worth mentioning that `AllocZeroed` doesn't have the API of `calloc`—[`calloc` takes two parameters and multiplies them together to compute the size to allocate](https://en.cppreference.com/w/c/memory/calloc).

PR #1885, opened 2023-05-06
Remove outdated info for clarity

redsquidz:
DevkitArm link no longer has `devkitpro-pacman.amd64.deb` listed; just the pkg file. This was confusing so it should be removed. I've only edited the WS1 entry, others may need to be updated too but I'm on mobile so the interface is difficult lol

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -482,7 +482,7 @@ Otherwise, follow the instructions below to install devkitARM.
     >   If the above command does not work, try the above command but replacing `apt` with `apt-get`.
     </details>
 
-2. Once `gdebi-core` is done installing, download the devkitPro pacman package [here](https://github.com/devkitPro/pacman/releases). The file to download is `devkitpro-pacman.amd64.deb`.
+2. Once `gdebi-core` is done installing, download the devkitPro pacman package [here](https://github.com/devkitPro/pacman/releases).
 3. Change directory to where the package was downloaded. For example, if the package file was saved to **C:\Users\\_\<user>_\Downloads** (the Downloads location for most users), enter this command, where *\<user> is your **Windows** username:
 
     ```bash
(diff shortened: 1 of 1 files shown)


GriffinRichards:
Ok, but is this step no longer required or does an alternative step need to be described?

redsquidz:
I had to install armkit for CrystalDust if that's what you mean? I was getting along just fine with my own hack without it

GriffinRichards:
No what you had to install for a different repo isn't relevant, just pokeemerald. I'm not on Windows so I don't know what is or isn't currently required.

redsquidz:
I would say we could lose the whole step then since i was able to get through without devkitArm... or at the very least include a disclaimer that it's for older builds like CD.

However, if devkitArm is required for other systems, would it also be needed for building repos made on those systems? Not relevant to a personal emerald windows build, but with how many forks are out there mentioning something would go along way to avoid poking the bee's nest, if that cross-platform requirement is true

redsquidz:
Oh you know that's a good point. I guess there's our answer then!

GriffinRichards:
I'm still not clear on what's required for WSL1. The updated instructions are still directing people to download pacman from devkitPro, it's just less specific about what because a package was removed from their GitHub. I'll let some other Windows person chime in.

redsquidz:
That's true, it'd be good to have someone test it from scratch before the instruction is dropped entirely.

If we want to close this we could just clear up the pkg/pacman confusion. It was the original reason for the PR anyway and would stop any wild goose chases in their tracks, which was all the PR was made for

sophxm:
The existing guidance is outdated, but this proposed change contains unnecessary steps and isn't fully clear (and the Debian/Ubuntu installation remains outdated).

Steps 1-3 for WSL should probably just be replaced with a link to the dkP instructions [on their wiki](https://devkitpro.org/wiki/devkitPro_pacman#Debian_and_derivatives) as is done for the Arch instructions.

Steps 1-2 for Debian/Ubuntu can be replaced in the same way (with `sudo gdebi devkitpro-pacman.amd64.deb` from step 3 also being removed, as was done to WSL step 4 in this commit).

PR #1887, opened 2023-05-15
Update GetRandomScaledFrontierTrainerId and GetRandomScaledFrontierTrainerIdRange to use FRONTIER_STAGES_PER_CHALLENGE instead of 7

Petuuuhhh:
Update GetRandomScaledFrontierTrainerId and GetRandomScaledFrontierTrainerIdRange to use FRONTIER_STAGES_PER_CHALLENGE instead of 7.
## Description
Same as above. Will help people who want to find where the Battle Tower handles how many trainers are in a row in trainer gauntlets.
## **Discord contact info**
Yak Attack#9182

### DIFF
--- a/src/battle_tower.c
+++ b/src/battle_tower.c
@@ -1105,7 +1105,7 @@ u16 GetRandomScaledFrontierTrainerId(u8 challengeNum, u8 battleNum)
 {
     u16 trainerId;
 
-    if (challengeNum <= 7)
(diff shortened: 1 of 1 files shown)


Petuuuhhh:
In that case, I feel a new constant is needed for the "gauntlet number", or whatever it's called, as discussed on Discord.

PR #1888, opened 2023-05-15
Add FRONTIER_REPEAT_CHALLENGES_START_HARDEST_RANGE constant

Petuuuhhh:
Add FRONTIER_REPEAT_CHALLENGES_START_HARDEST_RANGE constant
## Description
Every time you complete 7 battles, the Battle Frontier challenge repeats. After 7 repeats, the Battle Frontier uses trainer IDs that always come from the last, hardest range, which is the same for both trainer ID tables. This constant addresses that.
## **Discord contact info**
Yak Attack#9182


Petuuuhhh:
Oof, I thought I could change the branch name and the PR would stay open.

PR #1891, opened 2023-05-28
Fix compilation for modern under certain circumstances

ipatix:
When prepending something to PATH, we should use PATH=/foo/bar:$PATH and not PATH=/foo/bar:PATH. Because your shell won't certainly find something under the literal name PATH.

If one is not using devkitarm, TOOLCHAIN will resolve to an empty string. If TOOLCHAIN is an empty string the only search path for the modern gcc will be /bin which may work on some installations but is probably not what was originally intended with this line. So let's just search the standard search path if nothing can be found under $(TOOLCHAIN).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ LD := $(PREFIX)ld
 # note: the makefile must be set up so MODERNCC is never called
 # if MODERN=0
 MODERNCC := $(PREFIX)gcc
-PATH_MODERNCC := PATH=$(TOOLCHAIN)/bin:PATH $(MODERNCC)
+PATH_MODERNCC := PATH="$(PATH)" $(MODERNCC)
 
 ifeq ($(OS),Windows_NT)
 EXE := .exe


GriffinRichards:
From Soph#1219 on Discord:
> I think `$(PATH)` would be better than `$$PATH` unless there's something i've forgotten.
`$$PATH` works for you because it escapes the first `$` and passes `$PATH` to the `$(shell)` functions below, but if it was ever used in a non-shell context then you would just get the literal text `$PATH`. These makefiles also use `.SECONDEXPANSION` to expand variables twice, if that was ever moved higher up the file or if any of the rules ever used `PATH_MODERNCC` directly, it would go from `$$PATH` -> `$PATH` (aka `$(P)ATH`) -> `ATH` (as `$(P)` isn't defined) and we'd have the same problem you found. `$(PATH)` would avoid those

ipatix:
I have two reasons against this:

1. The contents of `PATH_MODERNCC` is explicitly shell syntax anyway
2. This breaks if you have a path variable with special characters which make doesn't support (i.e. space or paranthesis) which you commonly have for WSL setups. And yes, it does break on my machine.

If that's not a sufficient, I'd suggest getting rid of PATH setting for MODERNCC invocation if no TOOLCHAIN is specified. To me that now seems like the cleanest way of doing it.

sophxm:
Sorry, I could have explained my discord message more thoroughly.

> 1. The contents of `PATH_MODERNCC` is explicitly shell syntax anyway

Indeed, I was thinking of e.g. logging this variable, where the literal text `$PATH` would be written rather than its contents.

> 2. This breaks if you have a path variable with special characters which make doesn't support (i.e. space or paranthesis) which you commonly have for WSL setups. And yes, it does break on my machine.

Good point, this can be rectified with no downsides using quotes though.

One thing I didn't notice before is that `PATH` is already amended at the start of the file if `$TOOLCHAIN` exists. So we can simply use:
```make
PATH_MODERNCC := PATH="$(PATH)" $(MODERNCC)
```

This ensures that we use whatever value of `PATH` we configured in the makefile, avoids theoretical future issues with `.SECONDEXPANSION`, and should fix the problem with spaces in `PATH` that you noticed.

The reason we need to assign to `PATH` here in the first place is because variables set in makefiles aren't exported to `$(shell ...)` calls (until [Make 4.4](https://lists.gnu.org/archive/html/info-gnu/2022-10/msg00008.html), which only came out a few months ago and isn't in the Debian/Ubuntu package repo yet).

Thanks for spotting the error in my original suggestion!

N.B. you could also stop Windows' `PATH` from being included in WSL's: https://stackoverflow.com/a/51345880 (but of course it's still good to fix this for anyone who might not wish to do so).

ipatix:
@sophxm So what do you think about this solution? [51a3156](https://github.com/pret/pokeemerald/pull/1891/commits/51a3156698c741a7a526cc63fc32d74e8f78a8fa)

sophxm:
> @sophxm So what do you think about this solution? [51a3156](https://github.com/pret/pokeemerald/pull/1891/commits/51a3156698c741a7a526cc63fc32d74e8f78a8fa). Strictly speaking this would avoid prepending an erronous `/bin/` to the path if TOOLCHAIN is not set.

@ipatix Because `PATH` has already been modified earlier in the makefile (see [this](https://github.com/pret/pokeemerald/blob/51a3156698c741a7a526cc63fc32d74e8f78a8fa/Makefile#L15)), with that proposal we'll get: `$(TOOLCHAIN)/bin:$(TOOLCHAIN)/bin:...PATH...` (if `TOOLCHAIN` is set).

The following code should avoid that and also fix the quote issue:
```make
PATH_MODERNCC := PATH="$(PATH)" $(MODERNCC)
```
`PATH="$(PATH)"` is redundant if `TOOLCHAIN` wasn't set, but it doesn't make any difference to the command.

Admittedly it's not immediately obvious why we're setting `PATH` (which is because `PATH` in the shell could be different to `$(PATH)` in Make), so maybe a comment would also be good.

ipatix:
Sorry, I don't understand why you'd get `$(TOOLCHAIN)/bin:$(TOOLCHAIN)/bin:...PATH...`
PATH is never set if TOOLCHAIN is not set.

sophxm:
> Sorry, I don't understand why you'd get `$(TOOLCHAIN)/bin:$(TOOLCHAIN)/bin:...PATH...` PATH is never set if TOOLCHAIN is not set.

Yeah, if `TOOLCHAIN` isn't set then `PATH` won't be modified. But if `TOOLCHAIN` _is_ set then we first add `$(TOOLCHAIN)/bin` to it [here ](https://github.com/pret/pokeemerald/pull/1891/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R17), then your change also adds it to `PATH_MODERNCC`, so we end up with:

```make
PATH_MODERNCC := PATH=$(TOOLCHAIN)/bin:$(TOOLCHAIN)/bin:$(..system_path_here..) arm-none-eabi-gcc
```

I think changing this commit so all it does is `PATH_MODERNCC := PATH="$(PATH)" $(MODERNCC)` should avoid this duplication and fix the problem with spaces in `PATH` that you found.

Feel free to ping me on discord if that doesn't make sense or if I missed something, might be faster.

sophxm:
LGTM

PR #1898, opened 2023-07-10
Move sprite.c and string_util out of GFLib

AreaZeroArven:
These two files do not contain hardware-interfacing code, and thus make more sense in src than GFLib.

Plus, they were never part of the actual GFLib if the first place.


luckytyphlosion:
If you're planning on moving stuff out of `gflib` then you may as well move everything out of `gflib` (but you will cause lots of anguish with forks trying to update)

mrgriffin:
`gflib` no longer exists :)

PR #1901, opened 2023-07-24
Detect potential misalignment in modern

mrgriffin:
Wraps `CPU_FILL`, `CPU_COPY`, `DmaSet`, `DMA_FILL`, `DMA_CLEAR`, `DMA_COPY`, `CpuSet`, and `CpuFastSet` with a static assert which checks the source and destination arguments are appropriately aligned to either 16 or 32 bits, depending (in the sense of `alignof`, rather than the looser sense of "has an address with the bottom X bits clear").

I would have liked to be able to switch between the strict `alignof` behavior and the looser one, but I haven't yet been able to work out a way to make GCC give me that information, assuming such a way even exists.

Also adds `ALIGNMENT(x)` attributes to all variables/members which were flagged as potentially misaligned, or in a few places changed the types of `INCBIN`s. Note none of these variables are actually misaligned currently, but changes to the surrounding code can cause them to become misaligned, e.g. deleting unused members of `struct`s.

### DIFF
--- a/include/decompress.h
+++ b/include/decompress.h
@@ -3,7 +3,7 @@
 
 #include "sprite.h"
 
-extern u8 gDecompressionBuffer[0x4000];
+extern u8 ALIGNED(4) gDecompressionBuffer[0x4000];
 
 void LZDecompressWram(const u32 *src, void *dest);
 void LZDecompressVram(const u32 *src, void *dest);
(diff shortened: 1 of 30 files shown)


GriffinRichards:
They really did a number on `gIntro3Bg_Pal`, huh. Annoying that can't just be `u16`.

PR #1914, opened 2023-08-19
Added newer decomp projects to readme

AnonymousRandomPerson:
## Description
Added Gen 4 decomp projects to readme.

## **Discord contact info**
Some Body (anonymousrandom)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,14 @@ Other disassembly and/or decompilation projects:
 * [**Pokémon Pinball: Ruby & Sapphire**](https://github.com/pret/pokepinballrs)
 * [**Pokémon FireRed and LeafGreen**](https://github.com/pret/pokefirered)
 * [**Pokémon Mystery Dungeon: Red Rescue Team**](https://github.com/pret/pmd-red)
-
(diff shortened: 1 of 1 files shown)


GriffinRichards:
I don't personally have a problem removing the IRC link but there are definitely people who do want it there, so that should stay (whether it gets used or not I have no idea, I assume it does a little).

As an aside, this system of updating links is silly. These should get centralized somewhere under [pret](https://github.com/pret).

AnonymousRandomPerson:
I asked about IRC in the pret Discord server and haven't gotten any response so far. I noticed it wasn't included in any of the newer Gen 4 repos, so I interpreted that as the IRC channel being obsolete.

I agree that the system could be better. If there's a central place that we can agree on, I'd be happy to update all the repos to link to that.

GriffinRichards:
Fwiw, you opened the first PR removing the IRC link less than 15 minutes after asking. I don't really care about whether it's linked, but I do know the people who care about this kind of thing care a lot.

AnonymousRandomPerson:
Well, got a response faster when I made a PR. I added the IRC link back.

PR #1921, opened 2023-09-10
[LEAK-INFORMED] Add `blockBoxRS` field to `struct BoxPokemon`

citrusbolt:
## Description
Documents an unused data field in the Pokémon data structure.  None of the GBA games read or write to it, and to my knowledge, no event distributions utilized it.  I have not observed any behavior related to it in Colosseum, XD: Gale of Darkness, Diamond, Pearl, Platinum, HeartGold, or SoulSilver, but Box: Ruby & Sapphire will display "There is a Pokémon that can't be transferred" when trying to deposit a Pokémon that has the flag set to `TRUE`.

## **Discord contact info**
`@citrusbolt`

### DIFF
--- a/include/pokemon.h
+++ b/include/pokemon.h
@@ -202,7 +202,8 @@ struct BoxPokemon
     u8 isBadEgg:1;
     u8 hasSpecies:1;
     u8 isEgg:1;
-    u8 unused:5;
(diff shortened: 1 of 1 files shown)


GriffinRichards:
👍
Did you also confirm that any of the other 4 `unused` bits don't share this behavior?

citrusbolt:
Yep, the adjacent 4 bits don't seem to be referred to for this.

PR #1924, opened 2023-09-17
added installation troubleshooting info for macOS users

ecopsychologer:
## Description

I added two notes, one about using a VPN to run 'sudo dkp-pacman -Sy' because their servers wouldn't let me connect otherwise. The second was information on running terminal using rosetta, because my make failed unless I opened terminal in rosetta.
## **Discord contact info**

ecopsychologer#5583

done.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -294,7 +294,11 @@ If this works, then proceed to [Installation](#installation). Otherwise, ask for
     ```
 
     The command with gba-dev will ask for the selection of packages to install. Just press Enter to install all of them, followed by entering Y to proceed with the installation.
+<details>
+    <summary><i>Note for users with trouble running 'sudo dkp-pacman -Sy'...</i></summary>
 
+>   Using a VPN may help accesing devkitpro servers.
+</details>
 4. After the tools are installed, devkitARM must now be made accessible from anywhere by the system. To do so, run the following commands:
 
     ```bash
@@ -387,6 +391,13 @@ If this works, then proceed to [Installation](#installation). Otherwise, ask for
 >   Consider adding an exception for the `pokeemerald` and/or `decomps` folder in Windows Security using
 >   [these instructions](https://support.microsoft.com/help/4028485). This prevents Microsoft Defender from
 >   scanning them which might improve performance while building.
+</details>
+
+<details>
+    <summary><i>Note for macOS users...</i></summary>
+
+>   If make fails, try following
+>   [these instructions](https://support.apple.com/en-us/HT211861) to open Terminal using Rosetta.
 </details>
 
 1. If pokeemerald is not already downloaded (some users may prefer to download pokeemerald via a git client like GitHub Desktop), run:


GriffinRichards:
Using Rosetta is required for Apple Silicon Macs. The [pokefirered INSTALL.md](https://github.com/pret/pokefirered/blob/master/INSTALL.md?plain=1#L272-L276) already mentions this, just copy the comment from there instead.

ecopsychologer:
Thanks for the tip, I added all the instructions mentioning the m1 chips to this install.md

grunt-lucas:
That's odd. I am on an M1 mac, and I am not using a special Rosetta terminal profile. Nor do I have an amd64 Homebrew or libpng. I am using the regular arm64 homebrew installed in `/opt`, and my libpng is an arm64 dylib. Obviously my devkitARM binaries run through rosetta, but that happens in the background automatically. I don't M1 mac users need a special homebrew install in `/usr/lib`.

ecopsychologer:
Yeah tbh I only added that because I was asked to copy it from the pokefirered, I installed homebrew a while back and just brew install things

GriffinRichards:
> That's odd

You'd have to ask the M1 users who added it. I don't have access to one, so I can't verify any of this. I'm going off of what other users have told me. Feel free to update pokefirered if you disagree with its instructions, they should be the same across repos regardless.

grunt-lucas:
Ah I see. It's no big deal. I just always get curious around things related to Rosetta and installation processes. Because oftentimes I am bewildered about why something worked for me and not someone else, or vice-versa.

ecopsychologer:
Is there anything then I should change to get this added? I think having the extra documentation will help users setting this up

ecopsychologer:
Maybe make them both a 'try this then that' situation?

GriffinRichards:
If someone else with an M1 can verify this I don't have a problem with it. I don't have access to one myself.

INSTALL.md
```diff
@@ -270,13 +270,19 @@ If this works, then proceed to [Installation](#installation). Otherwise, ask for
...
+</details>
 
 1. Open the Terminal.
 2. If Homebrew is not installed, then install [Homebrew](https://brew.sh/) by following the instructions on the website.
 3. Run the following command to install libpng.
 
     ```bash
-    brew install libpng
+    brew install libpng # Intel Macs
+    /usr/local/bin/brew install libpng # Apple Silicon (M1) Macs
```
grunt-lucas, line 285: Imo these changes are not necessary

samirg1:
Hi there, I'm on an M1 and recently came across the issues mentioned with libpng, came across this and both of these changes helped me fix the issue. Thank you!

jarechiga11:
> Ah I see. It's no big deal. I just always get curious around things related to Rosetta and installation processes. Because oftentimes I am bewildered about why something worked for me and not someone else, or vice-versa.


> I think the advice is just outdated. I've built pokeemerald on multiple M1 macs by installing the regular Apple Silicon Homebrew, and `brew install`ing the Apple Silicon libpng. The devkitpro toolchain is still amd64, so that will run under Rosetta automatically. But I am pretty sure libpng is only needed to build the tools, and those are going to be built by your Apple Silicon C/C++ compiler (clang by default on MacOS).

Hey, I'm the original contributor from pokefirered that added the notes about rosetta being required for Apple Silicon Macs (M1 at the time). From my personal experience, it was required not only to install libpng but also for the `make` command to work. 

When i tried installing libpng through the typical brew command, it would appear to install fine but then the `make` command would always fail with this error
```
ld: warning: ignoring file '/usr/local/homebrew/Cellar/libpng/1.6.37/lib/libpng16.16.dylib': found architecture 'x86_64', required architecture 'arm64'
...
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I tried reinstalling libpng and even brew a few times but always got that error. It wasn't until I tried using the rosetta terminal for installing libpng and running the `make` command that I would get a successful build.

I guess it doesnt seem to affect every Apple Silicon user... So maybe there's a more appropriate fix, as I actually consider rosetta to still be a workaround too.

INSTALL.md
GriffinRichards: I still can't speak to M1, but now on an M3 I can add that in addition to `brew install libpng` I needed to install pkg-config with `brew install pkg-config` in order for libpng to be found.

PR #1926, opened 2023-09-20
Solved a bunch of -Wall errors on modern

AsparagusEduardo:
## Description
Most of these were picked from https://github.com/rh-hideout/pokeemerald-expansion/pull/3267.

## **Discord contact info**
AsparagusEduardo


src/fieldmap.c
```diff
@@ -48,6 +48,7 @@ static const struct MapConnection *GetIncomingConnection(u8 direction, int x, in
 static bool8 IsPosInIncomingConnectingMap(u8 direction, int x, int y, const struct MapConnection *connection);
 static bool8 IsCoordInIncomingConnectingMap(int coord, int srcMax, int destMax, int offset);
 
+
 #define GetBorderBlockAt(x, y)({                                                                   \
     u16 block;                                                                                     \
```
GriffinRichards, line 53: ?

src/librfu_rfu.c
```diff
@@ -1400,7 +1400,11 @@ static u16 rfu_STC_setSendData_org(u8 ni_or_uni, u8 bmSendSlot, u8 subFrameSize,
 {
     u8 bm_slot_id, sendSlotFlag;
     u8 frameSize;
+#ifdef UBFIX
+    u8 *llFrameSize_p = NULL;
+#else
     u8 *llFrameSize_p;
+#endif
```
GriffinRichards, line 1407: This doesn't actually fix anything (aside from silencing an uninitialized warning). In the case where this would be used it'd still be undefined behavior, because you're dereferencing a null pointer.

src/m4a_tables.c
```diff
@@ -261,10 +261,9 @@ const struct PokemonCrySong gPokemonCrySongTemplate =
     0, // block count
     255, // priority
     0, // reverb
-    (struct ToneData *)&voicegroup000,
-    NULL,
-    NULL,
-    0,
+    (struct ToneData *)&voicegroup000, // tone
+    {NULL, NULL}, // part
+    0, // gap
```
GriffinRichards, line 266: You can specify the struct member names with `.tone =` etc. if you want to do that, rather than comment their names. Otherwise just the fix is fine here.

src/battle_anim_effects_1.c
```diff
@@ -4734,7 +4734,7 @@ static void AnimFalseSwipeSlice(struct Sprite *sprite)
 
 static void AnimFalseSwipePositionedSlice(struct Sprite *sprite)
 {
-    sprite->x = sprite->x = GetBattlerSpriteCoord(gBattleAnimTarget, BATTLER_COORD_X_2) + 0xFFD0 + gBattleAnimArgs[0];
+    sprite->x = GetBattlerSpriteCoord(gBattleAnimTarget, BATTLER_COORD_X_2) + 0xFFD0 + gBattleAnimArgs[0];
```
GriffinRichards, line 4737: While you're here, that should be `-48` not `0xFFD0`

GriffinRichards (commented):
Just to play devil's advocate, do we definitely want to silence unused warnings upstream? I'd think if someone enables `-Wunused` they might reasonably want to know about these from the compiler, as opposed to searching the repo.

AsparagusEduardo:
> Just to play devil's advocate, do we definitely want to silence unused warnings upstream? I'd think if someone enables `-Wunused` they might reasonably want to know about these from the compiler, as opposed to searching the repo.

I was thinking that having them in the repo would be useful for users that want to do cleanup, by searching all instances of `UNUSED` directly instead of searching through the console log with `-Wunused` (which would get cutoff) or trying to compile 20+ times using `-Werror=unused` until it succedes.

src/m4a_tables.c
AsparagusEduardo: You're right, that's better. Will do.

src/librfu_rfu.c
AsparagusEduardo: I knew I forgot to copy something from the PR :(
I'll apply the null pointer dereferencing fixes as well.

src/fieldmap.c
AsparagusEduardo: My mistake, I meant to copy a fix here, but went back on it. Will correct.

src/battle_anim_effects_1.c
AsparagusEduardo: Done, though changing the one in `AnimFalseSwipeSlice` doesn't match.

AsparagusEduardo:
Changes applied.

src/battle_anim_effects_1.c
GriffinRichards: In `AnimFalseSwipeSlice` it needs to explicitly be `-48`, so `GetBattlerSpriteCoord(gBattleAnimTarget, BATTLER_COORD_X_2) + -48` or `-48 + GetBattlerSpriteCoord(gBattleAnimTarget, BATTLER_COORD_X_2)` match. Don't feel like you need to go back and do that though, given it's unrelated to the PR.

src/librfu_rfu.c
```diff
@@ -1426,7 +1430,11 @@ static u16 rfu_STC_setSendData_org(u8 ni_or_uni, u8 bmSendSlot, u8 subFrameSize,
     else if (gRfuLinkStatus->parentChild == MODE_CHILD)
         llFrameSize_p = &gRfuLinkStatus->remainLLFrameSizeChild[bm_slot_id];
     frameSize = llsf_struct[gRfuLinkStatus->parentChild].frameSize;
+#ifdef UBFIX
+    if (llFrameSize_p && (subFrameSize > *llFrameSize_p || subFrameSize <= frameSize))
+#else
     if (subFrameSize > *llFrameSize_p || subFrameSize <= frameSize)
+#endif
```
GriffinRichards, line 1437: Technically this should be
```c
if ((llFrameSize_p && subFrameSize > *llFrameSize_p) || subFrameSize <= frameSize)
```
because you still want to check the latter condition if `llFrameSize_p` is null (I couldn't tell you if this matters, I don't know if this might ever be called when `parentChild` is `MODE_NEUTRAL`).

AsparagusEduardo:
Done.

PR #1932, opened 2023-10-10
Modern Optimization: Compile math_util as ARM

AreaZeroArven:
Because math_util does not branch or write to memory, instead using register, using ARM code instead of THUMB works here for modern compilers. Also no need for -fno-toplevel-reorder because this optimization is safe for this too.

Here is an example:
Screenshot 2023-10-10 at 10 27 47 AM


SBird1337:
I mean... yes, but I don't think the project's scope is micro optimizing the codebase. This one is especially random since it is only used by one part of the berry crush minigame.

PR #1934, opened 2023-10-18
ALIGNED(2) The boxTitleTiles in storage

AreaZeroArven:
This bug was found by removing unused fields, only to find the graphics were corrupted because CpuCopy16 expects an aligned-2 pointer.

### DIFF
--- a/src/pokemon_storage_system.c
+++ b/src/pokemon_storage_system.c
@@ -431,7 +431,7 @@ struct PokemonStorageSystemData
     u16 scrollUnused5; // Never read
     u16 scrollUnused6; // Never read
     u8 filler1[22];
-    u8 boxTitleTiles[1024];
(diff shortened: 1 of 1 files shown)


mrgriffin:
Nice!

mrgriffin (approved):

mrgriffin:
I think the CI didn't run for this, and I'm not sure how to force it to run. Will merge as soon as I work it out/somebody tells me how to do it/you manage to trigger it from your end.

Jaizu:
I think that luckytyphlosion said that this should be a bugfix too:
https://github.com/luckytyphlosion/pokeemerald/pull/1/files
if thats the case, can you add those 2 here? Thanks.

RSilicon:
I think you can merge it now.

PR #1943, opened 2023-11-03
Delete the redundant files generated after running build

quocmanh94:
## Description
I found those redundant files generated from sound/direct_sound_samples/phonemes after running the build. But in the Makefile, we don't remove them.

## **Discord contact info**
quocmanh94

### DIFF
--- a/src/battle_ai_script_commands.c
+++ b/src/battle_ai_script_commands.c
@@ -1484,7 +1484,13 @@ static void Cmd_get_highest_type_effectiveness(void)
 
         if (gCurrentMove != MOVE_NONE)
         {
+            // TypeCalc does not assign to gMoveResultFlags, Cmd_typecalc does
+            // This makes the check for gMoveResultFlags below always fail
+#ifdef BUGFIX
+            gMoveResultFlags = TypeCalc(gCurrentMove, sBattler_AI, gBattlerTarget);
+#else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -279,6 +281,7 @@ include songs.mk
 %.lz: % ; $(GFX) $< $@
 %.rl: % ; $(GFX) $< $@
 $(CRY_SUBDIR)/%.bin: $(CRY_SUBDIR)/%.aif ; $(AIF) $< $@ --compress
+$(PHONE_SUBDIR)/%.bin: $(PHONE_SUBDIR)/%.aif ; $(AIF) $< $@ --compress
 sound/%.bin: sound/%.aif ; $(AIF) $< $@
```
GriffinRichards, line 285: This rule is incorrect (see build failure)

Makefile
```diff
@@ -238,6 +239,7 @@ clean-tools:
 mostlyclean: tidynonmodern tidymodern
 	rm -f $(SAMPLE_SUBDIR)/*.bin
 	rm -f $(CRY_SUBDIR)/*.bin
+	rm -f $(PHONE_SUBDIR)/*.bin
```
GriffinRichards, line 242: Having 3 rules for effectively the same thing is silly, and breaks again if a new subfolder is added.
```suggestion
	find sound -iname '*.bin' -exec rm {} +
```

GriffinRichards (changes requested):

PR #1947, opened 2023-11-08
Added POKEMART_LIST_END to avoid users accidentally removing it

AsparagusEduardo:
## Description
Previously, because it had the `ITEM_`/`DECOR_` prefix, users would mistakenly remove it from their custom mart lists.
This should hopefully aleviate the issue.

## **Discord contact info**
AsparagusEduardo

### DIFF
--- a/asm/macros/event.inc
+++ b/asm/macros/event.inc
@@ -1108,6 +1108,13 @@
 	.4byte \products
 	.endm
 
+	@ Used as the endpoint for a Pokemart item list
+	.macro pokemartlistend
+	.2byte ITEM_NONE
+	release
+	end
+	.endm
+
 	@ Opens the Pokemart system and treats the list of items as decorations.
 	@ Products should be a list of .2byte decoration values preceded by an .align 2
 	.macro pokemartdecoration products:req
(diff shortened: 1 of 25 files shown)


GriffinRichards:
I think this could be much more useful not as an alias of `ITEM_NONE`, given this isn't that much more visually distinct from the end of the list, and it's very unlikely anyone would redefine it. It might be more useful to define something like this:
```asm
.macro POKEMART_LIST_END
    .2byte ITEM_NONE
    release
    end
.endm
```

So that A. it would be more visually distinct from the end of the list:
```asm
    .2byte ITEM_CARBOS
    .2byte ITEM_HP_UP
    POKEMART_LIST_END
```

and B. so the unnecessary `release` and `end` are hidden and can be removed by editing one location.

AsparagusEduardo:
Good idea! Since it's gonna be a macro, would `pokemartlistend` be a better name for it?

GriffinRichards:
> Since it's gonna be a macro, would pokemartlistend be a better name for it?

Up to you

AsparagusEduardo:
13-month PR comeback babyyy!

PR #1952, opened 2023-11-12
[Build System Rewrite] Linkerscript now tracks RAM/ROM usage

Icedude907:
No dependencies on any other PRs

## Description
- The master linkerscripts now use `.ld` as their extension
- The linkerscript no longer sets the EWRAM heap size, the macro `HEAP_SIZE` does.
- The linkerscript should warn if memory bounds are exceeded.
- Running `make` now prints memory usages (note that on non-modern ROM's free space is fragmented due to `__anim_mon_load_address`)

Example output:
image

#### **Discord contact info**
icedude907

### DIFF
--- a/gflib/malloc.c
+++ b/gflib/malloc.c
@@ -4,7 +4,7 @@
 static void *sHeapStart;
 static u32 sHeapSize;
 
-__attribute__((section("__EWRAM_HEAP"))) u8 gHeap[HEAP_SIZE] = {0};
+EWRAM_DATA u8 gHeap[HEAP_SIZE] = {0};
 
 #define MALLOC_SYSTEM_ID 0xA3A3
 
--- a/gflib/malloc.c
+++ b/gflib/malloc.c
@@ -1,7 +1,10 @@
 #include "global.h"
+#include "malloc.h"
 
 static void *sHeapStart;
 static u32 sHeapSize;
+
+__attribute__((section("__EWRAM_HEAP"))) u8 gHeap[HEAP_SIZE] = {0};
 
 #define MALLOC_SYSTEM_ID 0xA3A3
 
--- a/gflib/malloc.h
+++ b/gflib/malloc.h
@@ -1,7 +1,6 @@
 #ifndef GUARD_ALLOC_H
 #define GUARD_ALLOC_H
 
-#define HEAP_SIZE 0x1C000
 
 #define FREE_AND_SET_NULL(ptr)          \
 {                                       \
@@ -11,6 +10,8 @@
 
 #define TRY_FREE_AND_SET_NULL(ptr) if (ptr != NULL) FREE_AND_SET_NULL(ptr)
 
+// 122 KB. Max size of the heap without running into other data
+#define HEAP_SIZE 0x1C000
 extern u8 gHeap[];
 
 void *Alloc(u32 size);
--- a/gflib/malloc.h
+++ b/gflib/malloc.h
@@ -10,9 +10,8 @@
 
 #define TRY_FREE_AND_SET_NULL(ptr) if (ptr != NULL) FREE_AND_SET_NULL(ptr)
 
-// 122 KB. Max size of the heap without running into other data
 #define HEAP_SIZE 0x1C000
-extern u8 gHeap[];
+extern u8 gHeap[HEAP_SIZE];
 
 void *Alloc(u32 size);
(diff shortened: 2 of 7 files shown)


gflib/malloc.h
```diff
@@ -11,6 +10,8 @@
 
 #define TRY_FREE_AND_SET_NULL(ptr) if (ptr != NULL) FREE_AND_SET_NULL(ptr)
 
+// 122 KB. Max size of the heap without running into other data
```
GriffinRichards, line 13: I don't think having this comment makes sense? The max size of the heap should just be its current size + however many bytes remain free in EWRAM, which is true of everything in EWRAM (and if we're printing the number of free bytes it makes less sense to list a static number here). Plus users would be alerted if it's oversized when it fails to link.

Unrelated but if you want you can update that declaration to `extern u8 gHeap[HEAP_SIZE];` so people can use `sizeof` if they want.

ld_script_modern.ld
```diff
@@ -3,28 +3,28 @@ ENTRY(Start)
 gNumMusicPlayers = 4;
 gMaxLines = 0;
 
+/* Memory Spaces */
+MEMORY
```
GriffinRichards, line 7: This comment is only in the modern linker script

ld_script_modern.ld
```diff
@@ -35,14 +35,13 @@ SECTIONS {
         src/m4a.o(.bss.code);
 
         /* COMMON starts at 0x30022A8 */
-        src/*.o(COMMON);
-        gflib/*.o(COMMON);
-        *libc.a:*.o(COMMON);
+        src/*.o(COMMON);            /**/
+        gflib/*.o(COMMON);          /**/
+        *libc.a:*.o(COMMON);
```
GriffinRichards, line 40: ?

ld_script.ld
```diff
@@ -3,35 +3,36 @@ ENTRY(Start)
 gNumMusicPlayers = 4;
 gMaxLines = 0;
 
+MEMORY
+{
+    EWRAM (rwx) : ORIGIN = 0x2000000, LENGTH = 256K
+    IWRAM (rwx) : ORIGIN = 0x3000000, LENGTH = 32K
+    ROM    (rx) : ORIGIN = 0x8000000, LENGTH = 16M
```
GriffinRichards, line 10: `ROM` size should be `32`, not `16`.

GriffinRichards (changes requested):

gflib/malloc.c
```diff
@@ -1,8 +1,11 @@
 #include "global.h"
+#include "malloc.h"
 
 static void *sHeapStart;
 static u32 sHeapSize;
 
+__attribute__((section("__EWRAM_HEAP"))) u8 gHeap[HEAP_SIZE] = {0};
```
GriffinRichards, line 7: As an aside, I don't think there's a technical reason to have the heap at the start of EWRAM. For matching purposes you can link `malloc` at the start of `sym_ewram` rather than giving `gHeap` its own section, and let modern throw the heap wherever.

ld_script.ld
Icedude907: Hi sorry for the delayed response.
Isn't the base cartridge 16mb or does it auto-expand when exceeding that limit?

GriffinRichards: Correct, there's otherwise no limit here. `gbafix` will just pad the ROM size up

ld_script_modern.ld
Icedude907: Fighting syntax highlighting. Forgot to remove

gflib/malloc.h
Icedude907: In the base game the heap doesn't consume all remaining EWRAM and it was placed at the front of the rom. 
Additionally whenever the heap is initialised it's passed the max size of that instance as a parameter (in practice only HEAP_SIZE). e.g.: https://github.com/pret/pokeemerald/blob/b302291bd3ef433c6eeed9a6940036d527ab13ae/src/credits.c#L417

So my thinking was that the constant HEAP_SIZE was in effect the *maximum heap size* and not technically the true heap size at any given moment.

Probably could convey that better but for now I'll remove it.

GriffinRichards: Yeah I think it was just a miscommunication, I interpreted the comment to mean "this is the maximum value `HEAP_SIZE` can be"

PR #1956, opened 2023-11-14
[Build System Rewrite] Sorted and cleaned up `graphics_file_rules.mk`

Icedude907:
- [ ] Depends on #1953 (but this is easily removable)

## Description
Rewrote most of `graphics_file_rules.mk` and sorted rules alphabetically. Some dead code has been removed.

The motivation behind this is to make the file easier to browse and parse for newcomers.

## **Discord contact info**
icedude907


GriffinRichards:
Aside from adding comments I'm not sure this is worth it if we're not going to merge the later changes that make use of the distinction between input/output directories. With a few exceptions this file is largely populated with rules that should be ignored by downstream users (because they're just for matching) so I don't think we need to worry that much about sorting the entries.

PR #1957, opened 2023-11-15
[Build System Rewrite] Improved audio rules

Icedude907:
No dependencies

## Description
- Coalesced songs and samples into one `audio_rules.mk` file
- A special file `midi.cfg` has been created which contains the commands for mid2agb. Replaces `songs.mk`. `audio_rules` expands the file into make rules.

## **Discord contact info**
icedude907


audio_rules.mk
```diff
@@ -0,0 +1,45 @@
...
+MID_CFG_PATH := $(MID_SUBDIR)/midi.cfg
+
+# $1: Source path no extension, $2 Options
+define MID_RULE
+$(MID_ASM_DIR)/$1.s: $(MID_SUBDIR)/$1.mid $(MID_CFG_PATH)
+	$(MID) $$< $$@ $2
+endef
+#                            source path,                             remaining text (options)
+define MID_EXPANSION
+	$(eval $(call MID_RULE,$(basename $(patsubst %:,%,$(word 1,$1))),$(wordlist 2,999,$1)))
```
GriffinRichards, line 38: For posterity it might be worth pointing out that the `999` is an arbitrarily long end point. Not necessary, just raised an eyebrow at first.

GriffinRichards (commented):
Looks fine to me (+ the missing newlines). One note is that it might be better to have `midi.cfg` in `sound`? By default it's alphabetically at the top in `sound/songs/midi`, but it's easy to miss in there.

GriffinRichards (approved):

Issue #1963, opened 2023-12-02
jsonproc improperly handles certain characters


GriffinRichards:
`cleanString` has special handling for `é`: https://github.com/pret/pokeemerald/blob/116364334d74242042d3ea81c6cc029d463748b7/tools/jsonproc/jsonproc.cpp#L108
but will fail to process some other reasonable characters. See https://github.com/huderlem/porymap/issues/557 for an example.

ExcmoJack:
I would like to add that if it's possible, the right substitution should be the character without the variation, for instance: `á` for `a`.

PR #1965, opened 2023-12-08
[jsonproc] Workaround to handle basic special characters

ExcmoJack:
## Description
This is a workaround to avoid compilation fails when including special characters in map names #1963 .
`pokeemerald/tools/jsonproc` was modified to include some special characters in the list of "bad characters" and implemented some commands to replace them with the better ascii aproximation:
```
aChars = "áàâä";
AChars = "ÁÀÂÄ";
eChars = "éèêë";
EChars = "ÉÈÊË";
iChars = "íìîï";
IChars = "ÍÌÎÏ";
oChars = "óòôöœ";
OChars = "ÓÒÔÖŒ";
uChars = "úùûü";
UChars = "ÚÙÛÜ";
nChars = "ñ";
NChars = "Ñ";
```

This makes that *it won't break anymore*, although it doesn't grants the correct result yet. It changes `Villa Raíz` for `Villa Ra_iz` i.e. however, I run out of ideas. My guess is that the first try to replace a _bad character_ also inserts the character even not having to replace it, but I don't know how to fix it.

## **Discord contact info**
Discord: excmojack

### DIFF
--- a/tools/jsonproc/jsonproc.cpp
+++ b/tools/jsonproc/jsonproc.cpp
@@ -105,11 +105,76 @@ int main(int argc, char *argv[])
     });
 
     env.add_callback("cleanString", 1, [](Arguments& args) {
-        string badChars = ".'{} \n\t-\u00e9";
+        string badChars = ".'{} \n\t-&¿?";
+        string aChars = "áàâä";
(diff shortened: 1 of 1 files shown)


GriffinRichards:
Handled for now by #1975

PR #1978, opened 2024-02-02
Fix HideMapNamePopUpWindow possible overflow

DizzyEggg:
Followup to #1976

We found the cause of the glitching map graphics. It was `GetMapNamePopUpWindowId` passing 0xFF as valid windowId, which in turn modifies the data directly below `gWindows`, which happened to be `gHeap` on some compiler/linker versions. In vanilla this goes unnoticed.

### DIFF
--- a/ld_script_modern.ld
+++ b/ld_script_modern.ld
@@ -15,11 +15,6 @@ SECTIONS {
     ewram 0x2000000 (NOLOAD) :
     ALIGN(4)
     {
-        /* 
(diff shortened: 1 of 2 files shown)


GriffinRichards:
It's too bad they didn't have a single point of reference for `gWindows`, it'd be nice to do this kind of check there (because I wouldn't be surprised if there are some other places this is happening quietly). But by my count there's 23 functions that access `gWindows` with an arbitrary window id, so this is definitely the less intrusive fix.

PR #1988, opened 2024-04-07
Include gba/defines.h in gba/types.h

mrgriffin:
If `gba/types.h` is included before `gba/defines.h`, it ends up declaring a variable named `PACKED` rather than specifying that `BgCnt` should be packed.

This occurs in `src/mini_printf.c`, and also apparently in a feature branch which is how this situation was detected (a multiple definition error at link time).


mrgriffin:
Sorry, you don't actually have `PACKED` on that `struct` here. mb.

PR #1991, opened 2024-04-14
Use Only One Loop in RandomlyGivePartyPokerus

Kurausukun:
## Description
Using a compound check instead of a nested loop is a bit nicer here. Also added missing SPECIES_NONE constant usage.

## **Discord contact info**
kurausukun


GriffinRichards:
That `0` doesn't represent a species, it's a null pointer.

Kurausukun:
My bad, you're right. Forgot how GetMonData worked for a second there

PR #1999, opened 2024-05-29
Fix build on NixOS and add instructions to INSTALL.md

leo60228:
## Description
This fixes a build error on NixOS caused by bash not being at /bin/bash, and adds instructions for building on NixOS to INSTALL.md. It does *not* include instructions for installing devkitARM, as devkitARM cannot be installed on NixOS without hacks that would almost certainly not be considered supported, but `make modern` works with the distribution's `arm-none-eabi` toolchain.

Also see https://github.com/pret/agbcc/pull/70.

## **Discord contact info**
leo60228


mrgriffin (approved):
Oh my god, YES! No more `make SHELL='bash -o pipefail'` for me :tada:

PR #2005, opened 2024-06-19
Define field script opcode constants

rayrobdod:
## Description
Also replace the magic number opcodes in field script command macros with these new constants

## **Discord contact info**
yoshord

### DIFF
--- a/asm/macros/event.inc
+++ b/asm/macros/event.inc
@@ -1,52 +1,54 @@
+	.include "data/script_cmd_table.inc"
+
 	@ Does nothing.
 	.macro nop
-	.byte 0x00
+	.byte SCR_OP_NOP
 	.endm
 
 	@ Does nothing.
 	.macro nop1
-	.byte 0x01
+	.byte SCR_OP_NOP1
 	.endm
 
 	@ Terminates script execution.
 	.macro end
-	.byte 0x02
+	.byte SCR_OP_END
 	.endm
 
 	@ Jumps back to after the last-executed call statement, and continues script execution from there.
 	.macro return
-	.byte 0x03
+	.byte SCR_OP_RETURN
 	.endm
 
 	@ Jumps to destination and continues script execution from there. The location of the calling script is remembered and can be returned to later.
 	.macro call destination:req
-	.byte 0x04
+	.byte SCR_OP_CALL
 	.4byte \destination
 	.endm
 
 	@ Jumps to destination and continues script execution from there.
 	.macro goto destination:req
-	.byte 0x05
+	.byte SCR_OP_GOTO
 	.4byte \destination
 	.endm
 
 	@ If the result of the last comparison matches condition (see Comparison operators), jumps to destination and continues script execution from there.
 	.macro goto_if condition:req, destination:req
-	.byte 0x06
+	.byte SCR_OP_GOTO_IF
 	.byte \condition
 	.4byte \destination
 	.endm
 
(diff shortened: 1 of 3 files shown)


data/script_cmd_table.inc
```diff
@@ -1,232 +1,249 @@
+	.ifndef SCRIPT_CMD_TABLE_ENTRY_MACRO
+	.set SCRIPT_CMD_TABLE_ENTRY_MACRO, 1
+	.macro script_cmd_table_entry constant:req value:req
+	.ifndef ALLOCATE_SCRIPT_CMD_TABLE
+	enum \constant
+	.else
+	.4byte \value
+	.endif
```
mrgriffin, line 8: I think `.if`/`.ifne` would be clearer than `.ifdef`/`.ifndef` (and are necessary with an explicit `.set ALLOCATION_SCRIPT_CMD_TABLE, 0`), e.g.:
```suggestion
	.if ALLOCATE_SCRIPT_CMD_TABLE
	.4byte \value
	.else
	enum \constant
	.endif
```

asm/macros/event.inc
```diff
@@ -1,52 +1,54 @@
+	.include "data/script_cmd_table.inc"
```
mrgriffin, line 1: I think it makes sense to be explicit here:
```suggestion
        .set ALLOCATE_SCRIPT_CMD_TABLE, 0
	.include "data/script_cmd_table.inc"
```

data/script_cmd_table.inc
```diff
@@ -1,232 +1,249 @@
...
+	.endm
+	.endif
+
+
+	enum_start
+	.ifdef ALLOCATE_SCRIPT_CMD_TABLE
```
mrgriffin, line 14:
```suggestion
	.if ALLOCATE_SCRIPT_CMD_TABLE
```

data/script_cmd_table.inc
```diff
@@ -1,232 +1,249 @@
...
+	script_cmd_table_entry SCR_OP_MESSAGEINSTANT                ScrCmd_messageinstant                 @ 0xdb
+	script_cmd_table_entry SCR_OP_FADESCREENSWAPBUFFERS         ScrCmd_fadescreenswapbuffers          @ 0xdc
+	script_cmd_table_entry SCR_OP_BUFFERTRAINERCLASSNAME        ScrCmd_buffertrainerclassname         @ 0xdd
+	script_cmd_table_entry SCR_OP_BUFFERTRAINERNAME             ScrCmd_buffertrainername              @ 0xde
+	script_cmd_table_entry SCR_OP_POKENAVCALL                   ScrCmd_pokenavcall                    @ 0xdf
+	script_cmd_table_entry SCR_OP_WARPWHITEFADE                 ScrCmd_warpwhitefade                  @ 0xe0
+	script_cmd_table_entry SCR_OP_BUFFERCONTESTNAME             ScrCmd_buffercontestname              @ 0xe1
+	script_cmd_table_entry SCR_OP_BUFFERITEMNAMEPLURAL          ScrCmd_bufferitemnameplural           @ 0xe2
 
+	.ifdef ALLOCATE_SCRIPT_CMD_TABLE
```
mrgriffin, line 246:
```suggestion
	.if ALLOCATE_SCRIPT_CMD_TABLE
```

mrgriffin (changes requested):
I really like the idea here and I think it will reduce mistakes, I just have a few comments on style :)

PR #2014, opened 2024-07-13
Rename battle terrain to environment

Bassoonian:
This renames battle terrain to battle environment.

## Description
Back in Generation 3, terminology for the appearance of battles and its effect on moves such as Camouflage and Nature Power was inconsistent. Hoenn and Ore refer to it as "location", which FRLG uses in the description of Nature Power. "terrain" is only used in the description of Secret Power and Camouflage.

In Generation 4, they started consistently referring to it as "environment" ([See the Bulbapedia article](https://bulbapedia.bulbagarden.net/wiki/Environment)). Consider the vast amount of current pret users had Gen4+ games as their first Pokémon game, environment will be a much more familiar name than terrain.

The reason for this PR is not just to promote inconsistency, but also to avoid ambiguity with the [Terrain](https://bulbapedia.bulbagarden.net/wiki/Terrain) mechanic introduced in Gen6. While technically not necessarily an issue inside the scope of pret, this sort of ambiguity has proven confusing for expansion users, and I think it's in everyone's advantage to ensure shared functions and concepts remain the same as much as possible between pret and expansion.

## **Discord contact info**
bassoonian


GriffinRichards:
To add my two cents, Environment is not a good replacement name for Terrain, it's not clear "environment" is only referring to the ground. What we call terrain is also not the same as what that wiki page calls the environment. It says:

> The environment consists of two primary components: the background and platform

In our repo, the former is the "battle background/scene" and only the latter is the terrain.

Bassoonian:
We could use "environment" for things as a whole (i.e. the effects on the moves), but use "platform" for what are currently the terrain graphics?

GriffinRichards:
The quote isn't correct from a code perspective, the Elite 4 backgrounds all use the same terrain. And sure, any name that makes it clear it's referring just to that.

Bassoonian:
Addressed concerns, feel free to get in touch if you disagree or want more changes

Bassoonian:
Accidentally closed this when cleaning up my branches, sorry

PR #2033, opened 2024-09-18
Remove gflib

Kurausukun:
## Description
This has been a long time coming.

There was never really a reason for us to mimic GF's internal structure, and even if there were, our current attempt doesn't even match what they did anyway.

## **Discord contact info**
Kurausukun


GriffinRichards:
👍  The hardcoded symbols in `sym_common.txt` can be replaced with file includes now, but that doesn't have to happen in this PR

GriffinRichards (approved):

PR #2035, opened 2024-09-21
Replaced copyright magic numbers in intro.c with constants

pkmnsnfrn:
## Description
Changed some magic numbers to constants for readability.

## **Discord contact info**
pkmnsnfrn


GriffinRichards:
This might be better served by comments at the start of each case block? I feel like by moving the numbers away it actually makes the flow a little harder to understand at a glance (i.e. before it's clearer that `default` is going to run until `state == 140`)

If you feel strongly about defining constants for this then they should probably be an enum to enforce the order:
```c
enum {
    COPYRIGHT_INITIALIZE,
    COPYRIGHT_START_FADE = 140,
    COPYRIGHT_START_INTRO,
};
```

pkmnsnfrn:
I figured constants would be good for users trying to add functionality to skip the copyright screen, to make it clear that they need to jump to a specific state.

I'll do the enum change if you agree that constants are acceptable, I think the enum change is a good idea.

PR #2040, opened 2024-09-30
Get rid of common syms

luckytyphlosion:
> Also only build tools once per makefile usage, error out if the submakes fail (only make 4.2+), print truncated link statement in terminal

## Description
luckytyphlosion — 06/16/2019
okay so this is how a common symbol is defined:
size <= 4: ALIGN by 4
4 < size <= 8: ALIGN by 8
size > 8: ALIGN by 16

then, define the symbol
you can't simply replace it with a section like common_data because it doesn't align correctly
it'll only align by 4
however, considering that ramscrgen just reads from the object file and manually outputs the symbols in the order they're defined based on common syms
you could probably make common_data anyway, but repurpose ramscrgen to fix the alignment of common_data symbols
and this would mean no common_syms/

## **Discord contact info**
luckytyphlosion


GriffinRichards:
Please fix the changes using `.SHELLSTATUS` to work for versions of Make < 4.2 or revert them. We don't the error badly enough to break older versions of Make.

ld_script.ld
```diff
@@ -42,6 +42,7 @@ SECTIONS {
 
         /* COMMON starts at 0x30022A8 */
         INCLUDE "sym_common.ld"
+        src/*.o(COMMON);
```
GriffinRichards, line 45: This won't include any `COMMON_DATA` symbols that users add to new files if they aren't explicitly included (i.e. we should wildcard `common_data` here too).

luckytyphlosion: I can imagine wildcarding COMMON not working previously because of the way that common symbols were previously included in the build (extracting them from the object files and then re-ordering them with alignment). However, since no variables in src/ use common variables anymore, we don't run into this problem.

The problem with wildcarding COMMON however occurs with common_data. `src/*.o(common_data);` in the linkerscript will cause the link process to run out of IWRAM, probably because the wildcarding isn't excluding files already in the build.

PR #2042, opened 2024-10-03
BUGFIX: Fix L=A by comparing the rawkeys to the previous keyinput

AreaZR:
Instead of the converted keys.


mrgriffin:
Is there somewhere in-game where I can verify that this change works?

AreaZR:
> Is there somewhere in-game where I can verify that this change works?

Absolutely. Because heldKeys compares with keyInput, holding the L button alongside holding another button means neither button will never be considered to be held if L=A is set. Holding L button alongside another button means both buttons are repeating.

Honestly, we can get away with not adding A to newandrepeating because it is never checked for newandrepeating, but what was mentioned about the check above still applies.

mrgriffin:
Sorry, I think you misunderstood my question.

Where in-game should I test the L=A repeating behavior so that I can check that this PR works?

AreaZR:
In all honesty, we can completely get away with not remapping newAndRepeatedKeys, as the only values checked in this value are DPAD values, but I just have it here for completion's sake. I can remove it if you want.

AreaZR:
What do you think @mrgriffin? Should I even bother remapping newAndRepeatedKeys?

mrgriffin:
I haven't had the time to familiarize myself with the code and test your change in-game yet. But once I've done that I can give you an opinion about `newAndRepeatedKeys` :)

AreaZR:
Please let me know when you get the chance :)

AreaZR:
Any updates on this by the way?

AreaZR:
newAndRepeatedKeys is only checked for DPAD values, so it doesn't affect pret at all.

Going to just take it out.

AreaZR:
"newAndRepeatedKeys is never remapped either." Should probably be moved to where the remapping happens, honestly.

PR #2043, opened 2024-10-10
[preproc] support C23 enum underlying type syntax

mrgriffin:
Introduces support for the C23 underlying type `enum` syntax. We wouldn't be able to use that syntax with `agbcc`, but it's potentially helpful for projects which exclusively use `make modern` (e.g. Expansion from v1.10 onward).

The `do`-`while` loop is to support things like `enum : unsigned char { ... };`, albeit I don't expect we'll see anything like that in practice.

### DIFF
--- a/tools/preproc/asm_file.cpp
+++ b/tools/preproc/asm_file.cpp
@@ -524,6 +524,17 @@ bool AsmFile::ParseEnum()
     long enumCounter = 0;
     long symbolCount = 0;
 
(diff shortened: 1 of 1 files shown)


GriffinRichards:
I'm ignoring the build failure here -- it's not the PR, it's just that the runner's OS updated to Ubuntu 24.04 (which broke some assumptions about libpng that are now fixed)

PR #2051, opened 2024-10-19
BUGFIX: Have fallback if level is 1

AreaZR:
This does not happen in vanilla, but in the event level is 1, prevent GetLevelFromMonExp from returning 0.


AreaZR:
Should not matter in vanilla, but it's a good fallback just in case.

AreaZR:
This is probably more useful in emerald expansion or other romhacks but it also doesn't hurt vanilla at all, so. Maybe the maintainers can give their opinions.

AreaZR:
Nah this isn't needed unless you change gExperienceTables, but I guess this works for like I don't know.

Bleh.

PR #2053, opened 2024-10-21
Document bard music

GriffinRichards:
- Add missing usage of the constant added in #2052, and correct a mistaken use of it.
- Rename `BARD_SONG_LENGTH` to `NUM_BARD_SONG_WORDS` to avoid confusion with the length of the song in time.
- Use `EC_WORD_*` constants for the remaining bard sound arrays


GriffinRichards:
Went down a rabbit hole, this is now documenting some of the opaque parts of how the Bard NPC's music works, including using the `PH_` song constants in the bard music data.

mrgriffin (commented):
I meant to review this ages ago and then it just totally slipped my mind.

`#define USE_NEW_SONG (gSpecialVar_0x8004)` is the only change that I'm torn on. But I'm not sure I have a good argument against it, just a vague apprehension about hiding a special var behind a macro because when I see something like `StartBardSong(USE_NEW_SONG);` I assume `USE_NEW_SONG` to be a constant.

That said, I'm happy to merge this as-is if you're confident that `USE_NEW_SONG` is an improvement over writing `gSpecialVar_0x8004` :)

GriffinRichards:
Thanks! I do think there's some value in it (giving it a meaningful name, making it easier to swap out if that var will be used for something else in the script), but I'd rather it not be confusing so I'll remove it. Maybe some day someone could work through the repo and give aliases to these vars in a different, standardized way.

mrgriffin:
> Maybe some day someone could work through the repo and give aliases to these vars in a different, standardized way.

That person would be my hero :sparkles:

PR #2062, opened 2024-11-04
Lay out emerald version png horizontally

GriffinRichards:
Adds a Make rule so that the tiles in `emerald_version.png` are laid out the same way they're displayed in-game.


mrgriffin (approved):

manuman94:
Thanks for this! I think it's much better than my Python script!

PR #2069, opened 2024-11-24
Automatically update MAP_DYNAMIC if necessary

GriffinRichards:
At the moment, the 128th map in any map group will be treated as `MAP_DYNAMIC` because [it's read using only MAP_NUM](https://github.com/pret/pokeemerald/blob/master/src/field_control_avatar.c#L817). Because `MAP_DYNAMIC` is in the middle of the map value range we can alleviate this problem a little by automatically increasing its value if there are sufficient maps in any map group. Although `MAP_GROUP` isn't used in isolation to read `MAP_DYNAMIC`, we can do the same if there are sufficient map groups just in case. `MAP_UNDEFINED` *is* read with `MAP_GROUP` in isolation, but it's already at the end of the value range.

Also added error messages to `mapjson` if the number of maps or map groups exceeds the value range (including the space reserved for `MAP_DYNAMIC` and `MAP_UNDEFINED`). Frustratingly the data type for map groups / map numbers is treated inconsistently and sometimes uses `s8`, so although these values are allowed to reach 255, there can be value range errors for values > 127. This problem is left unchanged by this PR.

This is targeting the `porymap-6` branch for now because at the moment Porymap is (unnecessarily) overwriting the `map_groups.h` file.

### DIFF
--- a/include/constants/maps.h
+++ b/include/constants/maps.h
@@ -3,12 +3,6 @@
 
 #include "map_groups.h"
 
-// Warps using this map will instead use the warp data stored in gSaveBlock1Ptr->dynamicWarp.
-// Used for warps that need to change destinations, e.g. when stepping off an elevator.
-#define MAP_DYNAMIC (0x7F | (0x7F << 8))
-
-#define MAP_UNDEFINED (0xFF | (0xFF << 8))
-
 #define MAP_GROUP(map) (map >> 8)
 #define MAP_NUM(map) (map & 0xFF)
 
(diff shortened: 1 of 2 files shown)


GriffinRichards:
Updated to add a `UBFIX` that removes the cast for `MAP_NUM`/`MAP_GROUP` values from `u8` to `s8`. Without this, if a map exceeds 127 on either value it will almost certainly softlock or crash the game on load when it's later cast to `u16` to read `gMapGroups`

PR #2071, opened 2024-12-04
Fix recorded battle link player loops

AsparagusEduardo:
## Description
2 for loops used `MAX_BATTLERS_COUNT` instead of `MAX_LINK_PLAYERS`, which causes potential errors when `MAX_BATTLERS_COUNT` is increased.

## **Discord contact info**
AsparagusEduardo

### DIFF
--- a/src/recorded_battle.c
+++ b/src/recorded_battle.c
@@ -35,13 +35,13 @@ struct RecordedBattleSave
 {
     struct Pokemon playerParty[PARTY_SIZE];
     struct Pokemon opponentParty[PARTY_SIZE];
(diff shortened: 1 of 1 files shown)


GriffinRichards:
These loops are iterating over other arrays that also use `MAX_BATTLERS_COUNT`, if the loops are updated the arrays should be updated to match.

AsparagusEduardo:
> These loops are iterating over other arrays that also use `MAX_BATTLERS_COUNT`, if the loops are updated the arrays should be updated to match.

Done 👍

PR #2073, opened 2024-12-05
Improve Pyramid location documentaion

AsparagusEduardo:
## Description
The `InBattlePyramid` name was a bit of a misnomer, as it could return 0, 1 or 2 instead of just 0 and 1.
Added an in-line function with the original name to reduce verboseness and potential downstream conflicts.

## **Discord contact info**
AsparagusEduardo


SBird1337:
hmm, actually with the new semantics of `InBattlePyramid` if someone had
```c
if (InBattlePyramid() == 2)
```
they'll get surprising results. I'd personally rather have an error pointing me to the new name.

AsparagusEduardo:
Good point.
I'll change the checks for `if (InBattlePyramid())` and such to `if (CurrentBattlePyramidLocation() != PYRAMID_LOCATION_NONE)`

AsparagusEduardo:
Applied :)

src/field_poison.c
```diff
@@ -90,9 +90,9 @@ static void Task_TryFieldPoisonWhiteOut(u8 taskId)
         {
             // Battle facilities have their own white out script to handle the challenge loss
 #ifdef BUGFIX
-            if (InBattlePyramid() || InBattlePike() || InTrainerHillChallenge())
+            if (CurrentBattlePyramidLocation() != PYRAMID_LOCATION_NONE || InBattlePike() || InTrainerHillChallenge())
 #else
-            if (InBattlePyramid() | InBattlePike() || InTrainerHillChallenge())
+            if (CurrentBattlePyramidLocation() | InBattlePike() || InTrainerHillChallenge())
```
GriffinRichards, line 95: These should be the same, otherwise it looks like the difference is significant for the bug fix.

src/item_use.c
```diff
@@ -983,7 +983,7 @@ static void Task_UseStatIncreaseItem(u8 taskId)
     {
         PlaySE(SE_USE_ITEM);
         RemoveBagItem(gSpecialVar_ItemId, 1);
-        if (!InBattlePyramid())
+        if (CurrentBattlePyramidLocation() == PYRAMID_LOCATION_NONE) // CurrentBattlePyramidLocation() == PYRAMID_LOCATION_NONE doesn't match
```
GriffinRichards, line 986: Comment left behind

AsparagusEduardo:
Sorry for the wait, applied review changes

PR #2088, opened 2025-01-10
Move gTradePlatform_Tilemap to header and change to u32

DizzyEggg:


GriffinRichards:
Shouldn't this be `u16`? Those tilemap entries are 2 bytes each.

DizzyEggg:
> Shouldn't this be `u16`? Those tilemap entries are 2 bytes each.

I can change it to u16

DizzyEggg:
> Shouldn't this be `u16`? Those tilemap entries are 2 bytes each.

Fixed @GriffinRichards

PR #2096, opened 2025-01-19
Replace easy_chat alphabetialOrder magic numbers with EC_INDEX values

rayrobdod:
## Description
Changed some magic numbers to constants for readability

## **Discord contact info**
yoshord


GriffinRichards:
Yikes, talk about a maintenance nightmare. I might look into generating this data automatically, so that if someone were to add a new easy chat word they wouldn't have to shift all these alphabetical values themselves.

Issue #2093, opened 2025-01-12
Forced switchout moves causes issues in multi-battles


shachar700:
Similar issue to #5842 from the expansion
If the opponent uses forced switchout moves in a multi battle, the player side party might get the order messed up where the mons appear in the wrong slots of the party screen.
This causes a hardlock when all mons on the player side are knocked out and the player can't select their mon that's on the partner side of the party screen.
This can also cause the opposing Pokemon be on your side of the field

![pokeemerald-2](https://github.com/user-attachments/assets/c61fc563-d414-4694-a21c-ef561ee147c1)
![pokeemerald-3](https://github.com/user-attachments/assets/e69701b6-6f80-4949-b507-5d8aa29c014a)

PR #2099, opened 2025-01-28
Fixed switchout bug in multibattle where order of mons gets messed up

shachar700:
Fixed switchout bug that messed the order of pokemon in the summary screen after the opponent uses switchout moves (whirlwind, roar, etc...).

PROBLEM:
Screenshot_12

PROBLEM: If the opponent uses forced switchout moves in a multi battle, the player side party might get the order messed up where the mons appear in the wrong slots of the party screen.
This causes a hardlock when all mons on the player side are knocked out and the player can't select their mon that's on the partner side of the party screen.

FIX: Entered the switchout effects inside the if (TryDoForceSwitchOut()) statement, also adjusted firstmonId and lastmonId to stay between the boundaries of each player (0-2 PLAYER, 3-5 STEVEN)

## Description

#2093 Fixes

## **Discord contact info**

Shachar700

### DIFF
--- a/src/battle_script_commands.c
+++ b/src/battle_script_commands.c
@@ -7210,18 +7210,18 @@ static void Cmd_forcerandomswitch(void)
             {
                 firstMonId = PARTY_SIZE / 2;
                 #ifdef BUGFIX
-                    lastMonId = PARTY_SIZE - 1;
+                lastMonId = PARTY_SIZE; - 1
                 #else
-                    lastMonId = PARTY_SIZE;
+                lastMonId = PARTY_SIZE;
                 #endif
             }
             else
             {
                 firstMonId = 0;
                 #ifdef BUGFIX
-                    lastMonId = PARTY_SIZE / 2 - 1;
+                lastMonId = PARTY_SIZE / 2 - 1;
                 #else
-                    lastMonId = PARTY_SIZE;
+                lastMonId = PARTY_SIZE / 2;
                 #endif
             }
             monsCount = PARTY_SIZE / 2;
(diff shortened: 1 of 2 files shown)


mrgriffin:
This is going to need `#ifdef BUGFIX`...`#endif`s around it, because it doesn't match as-is.

mrgriffin (changes requested):

mrgriffin:
Without the fix is it possible for my Pokémon to get switched with one of Steven's or does the bug only manifest in the party screen? How would you recommend to test the fix?

shachar700:
I recommend to test the fix by giving the opponent switchout moves, and check summary screen after every turn. The fix fixes the case where mightyena is on your side of the field and also the order of the pokemon in the summary screen not getting scrambled with steven's slots.

src/battle_script_commands.c
```diff
@@ -7301,6 +7337,38 @@ static void Cmd_forcerandomswitch(void)
...
+                {
+                    do
+                    {
+                        i = Random() % monsCount;
+                        i += firstMonId;
+                    }
+                    while (i == battler2PartyId || i == battler1PartyId);
+                } while (GetMonData(&party[i], MON_DATA_SPECIES) == SPECIES_NONE
+                       || GetMonData(&party[i], MON_DATA_IS_EGG) == TRUE
+                       || GetMonData(&party[i], MON_DATA_HP) == 0); //should be one while loop, but that doesn't match.
```
AlexOn1ine, line 7353: If this is a `BUGFIX` why should it match?

src/battle_script_commands.c
```diff
@@ -7301,6 +7337,38 @@ static void Cmd_forcerandomswitch(void)
...
+                *(gBattleStruct->monToSwitchIntoId + gBattlerTarget) = i;
```
AlexOn1ine, line 7354:
```suggestion
                gBattleStruct->monToSwitchIntoId[gBattlerTarget]  = i;
```

src/battle_script_commands.c
shachar700: I didn't know how to wrap only the necessary changes in that if statement as it's most indents and a switch in position of the closing } bracket.

PR #2104, opened 2025-02-13
Fix collision comparison in PlayerNotOnBikeMoving

GriffinRichards:
The condition was optimized to a single comparison, which made it hard to read. While making it more readable I noticed that it was treating `COLLISION_ROTATING_GATE` specially despite it not having any special behavior. The result is that walking into a rotating gate that can't rotate won't update the player avatar correctly. This looks especially weird if you try to collide with a rotating gate by changing direction.

Before bug fix (walking head-on, and then trying to turn into it):
collision-bad

After bug fix (same inputs as above)
collision-good

### DIFF
--- a/src/field_player_avatar.c
+++ b/src/field_player_avatar.c
@@ -620,9 +620,27 @@ static void PlayerNotOnBikeMoving(u8 direction, u16 heldKeys)
         }
         else
         {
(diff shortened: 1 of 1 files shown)


DizzyEggg (approved):
That's a pretty cool bugfix, any reason it hasn't been merged?

PR #2110, opened 2025-03-17
Remove casts in decoration.c

rayrobdod:
## Description

Most of the removed casts are identity casts - e.g. casting a `u8 *` to a `u8 *`

Use `gTilesetPointer_SecretBase->palettes` nature as a two-dimensional array; instead of casting to a one-dimensional array and manually multiplying by the matrix width

Instead of casting `gTilesetPointer_SecretBase->tiles` to a `u8`, instead divide the index by `tiles`' element size

Also change a palette size to `PLTT_SIZE_4BPP`

## **Discord contact info**
yoshord

### DIFF
--- a/src/decoration.c
+++ b/src/decoration.c
@@ -1921,7 +1921,7 @@ static void ClearPlaceDecorationGraphicsDataBuffer(struct PlaceDecorationGraphic
 
 static void CopyPalette(u16 *dest, u16 pal)
 {
-    CpuFastCopy(&((u16 *)gTilesetPointer_SecretBase->palettes)[pal * 16], dest, sizeof(u16) * 16);
+    CpuFastCopy(gTilesetPointer_SecretBase->palettes[pal], dest, PLTT_SIZE_4BPP);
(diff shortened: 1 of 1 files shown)


GriffinRichards:
```
src/decoration.c: In function 'CopyPalette':
src/decoration.c:1924:10: error: static assertion failed: "source potentially unaligned"
 1924 |     CpuFastCopy(gTilesetPointer_SecretBase->palettes[pal], dest, PLTT_SIZE_4BPP);
      |          ^~~~~~~~~~~~~~
```
This fails to build with a modern compiler. Did you mean this?:
```c
CpuFastCopy(&gTilesetPointer_SecretBase->palettes[pal], dest, PLTT_SIZE_4BPP);
```

rayrobdod:
After seeing the modern compile fail, I started to fall into a rabbit-hole w.r.t. alignas and alignof. But that hasn't gotten anywhere, so if your suggestion is acceptable then I'll go with that.

PR #2116, opened 2025-04-02
Rename ItemId_GetX to GetItemX

AlexOn1ine:
For expansion we've been using getters for const data that begin with the prefix `Get`, e.g. `GetMoveEffect` so we've been thinking to make the change there for items but wanted to suggest it here first


mrgriffin:
I think this PR makes sense. These functions are the only ones I can think of that use the `FooId_Bar` naming scheme. In other places we have things like `GetNature`, `GetObjectEventIdByLocalId`, `GetBerryInfo`.

If somebody strongly disagrees, please let me know, otherwise I'll merge it later this week.

AlexOn1ine:
> Nobody has voiced a disagreement, so this is ready to go once the conflicts are solved :)

solved! :smile:

PR #2122, opened 2025-04-21
add compatibility with clangd for better editing with VSCode

jendrikw:
## Description
Configure clangd to allow editing with VSCode. To facilitate this, some declarations are moved around, some types are adjusted and some include guards added.

Compiles to the same rom.

## **Discord contact info**
jendr1k_


src/util.c
```diff
@@ -124,16 +124,12 @@ u8 CreateInvisibleSpriteWithCallback(void (*callback)(struct Sprite *))
...
+void StoreWordInTwoHalfwords(u16 *h, u32 w) {
+  h[0] = (u16)(w);
+  h[1] = (u16)(w >> 16);
 }
 
-void LoadWordFromTwoHalfwords(u16 *h, u32 *w)
-{
-    *w = h[0] | (s16)h[1] << 16;
-}
+void LoadWordFromTwoHalfwords(u16 *h, u32 *w) { *w = h[0] | (s16)h[1] << 16; }
```
mrgriffin, line 132: I haven't checked everything, but did you intend to reformat code like in this example?

src/pokemon_storage_system.c
```diff
@@ -9453,7 +9453,7 @@ u32 GetBoxMonLevelAt(u8 boxId, u8 boxPosition)
 #ifdef BUGFIX
     else
 #endif
-        lvl = 0;
+        lvl = 0; // NOLINT(readability-misleading-indentation)
```
mrgriffin, line 9456: Just my opinion, but I kinda hate putting these sorts of "tool-only" comments in the code.

If you're committed to `clangd` not throwing a spurious warning here, is there a different way the code could be formatted which would read naturally and not throw the warning? Or maybe `clangd` could be told to assume `BUGFIX` is defined?

src/pokedex.c
```diff
@@ -5137,11 +5136,11 @@ static void Task_SelectSearchMenuItem(u8 taskId)
 {
     u8 menuItem;
     u16 *cursorPos;
-    u16 *scrollOffset;
+    s16 *scrollOffset; // should also
```
mrgriffin, line 5139: I think this comment is incomplete?

src/m4a_tables.c
```diff
@@ -243,7 +245,7 @@ const u8 gClockTable[] =
 #define BENDR  0xc1
 #define LFOS   0xc2
 #define LFODL  0xc3
-#define MOD    0xc4
+#define MOD  0xc4
```
mrgriffin, line 248: I don't think you meant to commit this change?

src/list_menu.c
```diff
@@ -1261,17 +1261,17 @@ void ListMenuSetUpRedOutlineCursorSpriteOamTable(u16 rowWidth, u16 rowHeight, st
     s32 i, j, id = 0;
 
     subsprites[id] = sSubsprite_RedOutline1;
-    subsprites[id].x = 136;
-    subsprites[id].y = 136;
+    subsprites[id].x = (s8) 136;
+    subsprites[id].y = (s8) 136;
```
mrgriffin, line 1265: We format casts as `(s8)136` rather than `(s8) 136`. There's a few other instances in this file, and I think there's a bunch in other files too (usually to do with unsigned vs signed confusion on GF's part).

src/data/region_map/city_map_tilemaps.h
```diff
@@ -1,3 +1,6 @@
+#ifndef REGION_MAP_CITY_TILEMAPS_H
+#define REGION_MAP_CITY_TILEMAPS_H
```
mrgriffin, line 2: What's the purpose of these include guards? The `src/data/*.h` files would only be `#include`d directly into a single `.c` file (otherwise you'd get multiple definition errors from the linker), so the guards are misleading imo—they state that it's expected to include this file from multiple places.

src/data/decoration/icon.h
```diff
@@ -1,3 +1,5 @@
+#include <stddef.h>
+
```
mrgriffin, line 2: On a similar note, I get that the idea here is to tell your tools that `NULL` is in scope, but it seems really sketchy to have `#include` blocks at the top of files that are intended to be included part-way through a `.c` file.

As a user, I think it would be very confusing if there's a `#include` in a `src/data/*.h` file that isn't a no-op—in that situation you would have names and types that only start existing half-way through your `.c` file, in a way that would be very surprising.

Obviously none of your `#include`s function like that (or at least I hope they don't!), but I think it's laying a small trap for future users that aren't as savvy about the conventions in `pokeemerald`.

src/battle_anim_effects_1.c
```diff
@@ -4452,7 +4452,14 @@ static void AnimBowMon_Step1_Callback(struct Sprite *sprite)
...
         sprite->data[3] = gBattlerSpriteIds[gBattleAnimAttacker];
         PrepareBattlerSpriteForRotScale(sprite->data[3], ST_OAM_OBJ_NORMAL);
+#ifdef CLANGD
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wconstant-conversion"
+#endif
         sprite->data[4] = (sprite->data[6] = GetBattlerSide(gBattleAnimAttacker)) ? 0x300 : 0xFFFFFD00;
+#ifdef CLANGD
+#pragma clang diagnostic pop
+#endif
```
mrgriffin, line 4462: Is there some other way this could be written, or is the problem that `0xFFFFFD00` needs to be written literally like that and can't be `-0x300` (or failing that, `0xFD00`)?

It's a shame to turn one line into eight :sweat_smile:

kazoud:
Not sure if I should ask this here or on discord, but can you elaborate on how the changes to the declarations, the includes and the guards facilitate the work of clangd?

include/main.h
```diff
@@ -1,6 +1,7 @@
 #ifndef GUARD_MAIN_H
 #define GUARD_MAIN_H
 
+#include <stdint.h>
```
mrgriffin, line 4: What's `stdint.h` for?

src/util.c
jendrikw: This reformatting (and others) are on accident, VSCode just really wants to format stuff. I will revert the formatting changes.

src/pokemon_storage_system.c
jendrikw: Indenting the line with four spaces is fine for clang, if that's ok for you. I think clangd should have no warnings no matter the setting of BUGFIX.

src/pokedex.c
jendrikw: I'll remove it

src/m4a_tables.c
jendrikw: will revert

PR #2125, opened 2025-04-26
Battle link reverse-engineering

DavidJCobb:
Studied the code for battle-related multiplayer link communications. Updated the naming for relevant constants and variables, and wrote macros that describe the functionality of battle controller and link communication internals.

I tested these changes by running `make clean` followed by `make && make compare`, and got an `OK` result.

## Description

* Renamed `BUFFER_A` and `BUFFER_B` to `BATTLELINKMSGTYPE_ENGINE_TO_CONTROLLER` and `BATTLELINKMSGTYPE_CONTROLLER_TO_ENGINE`, and added `BATTLELINKMSGTYPE_CONTROLLER_BECOMING_IDLE` as the name for value 2.

* Named the task data for the "send" and "receive" tasks used to handle battle link communications (e.g. `#define data[10] tAnInformativeName`).

* Added an enum to name the state indices for the "send" task.

* Added some macros to those task function definitions, to make reads/writes into the battle communication buffers more legible. These macros are `#define`d and `#undef`ined within the functions in which they are used.

* Wrote documentation comments near the task data names, and near some of the relevant functions.

* Wrote macros for the bitwise operations used to query and update `gBattleControllerExecFlags`, and updated various functions internal to the battle engine to use these macros. A documentation comment near the macros clarifies that these are descriptive rather than being meant for use outside of those internals.

## Background

Writing this here for now, in case it helps reviewers. I'm probably going to contribute a wiki article once I have a more complete picture of how this all connects to the broader battle engine.

Battles can be thought of as a series of communications between the core **battle engine**, and the **battle controllers** that direct each battler, with messages (formally called "commands" in the PRET codebase, since "battle messages" refers to UI within the codebase) exchanged between them. The general procedure is that the battle engine sends a message to a controller and then sets flags to mark the controller as "active;" the controller then receives these messages, reacts to them (potentially including sending a message back to the engine), and then clears its flag to signal when it's done reacting. In general, only one message can be passed at a time (i.e. the engine waits for the controller to become idle, and then checks for a reply if one would be relevant).

During offline play, controller exec flags are modified directly to activate or idle a controller, and messages are written directly into `gBattleBufferA[battlerId]` (engine-to-controller) or `gBattleBufferB[battlerId]` (controller-to-engine). During multiplayer links, an additional layer of indirection is added to some of these steps: controllers are never activated directly, but rather, the engine sets a flag indicating that the controller is pending synchronization and activation amongst all players; and battle communications are written into intermediary buffers, to be sent and received over the link. Once the engine-to-controller message is synchronized amongst all players, the "pending synchronization" exec flag is cleared and the "active" exec flag is set. The procedure for a controller becoming idle is similar, except that the "pending synchronization" flags are not involved; the appropriate exec flag is cleared, but this is done via link communications rather than being done directly.

If Pokemon worked like some other games with networked play, then the "link master" might run battle controllers locally at the same time that it sends the controller messages over the wire... but that's *not* how it's done. [The way GBA link communications work](https://problemkaputt.de/gbatek.htm#siomultiplayermode) makes it so that it's "natural" to treat the sender and the recipients uniformly -- to handle the sender as one of four recipients of its own outbound data. Pokemon does that: the sender sets the "pending synchronization" flag, and then waits to receive data from itself; on both the sender and the recipients, the battle controller activates upon receipt of the data and reacts to that received data. Ditto for clearing the controller's exec flag when it finishes executing: it's done when the sender receives the "controller becoming idle" notification from itself.

## Potential future improvements

It occurs to me that it might be a good idea to rename `gBattleBufferA` and `gBattleBufferB` to something like `gCommBuffers_EngineToController` and `gCommBuffers_ControllerToEngine`, either as a condition of accepting this PR or as a future PR.

As mentioned, messages exchanged between the battle engine and battle controllers are called "commands" in the codebase. The battle engine generally is "commanding" the battle controllers to perform some action or provide some information, but when battle controllers respond, they do so via the same mechanism, and I don't think we can say that the controllers are "commanding" the engine. (Even when the exchange involves the engine asking the controller to decide what it wants to do, the engine doesn't have to honor the controller's decision, e.g. if the controller is for a battler who has been hit with Encore.) It may be a good idea to replace "command" with a different term at some point in the future, but we can't use "message" because a "battle message" is a predefined string meant for use in the UI. Maybe "signals?" Dunno.

## **Discord contact info**
`@davidjcobb`


include/battle_controllers.h
```diff
@@ -64,11 +64,56 @@ enum {
...
 // Special arguments for Battle Controller functions.
 
-enum { // Values given to the emit functions to choose gBattleBufferA or gBattleBufferB
-    BUFFER_A,
-    BUFFER_B
+enum {
+   BATTLELINKCOMMTYPE_ENGINE_TO_CONTROLLER, // gBattleBufferA
+   BATTLELINKCOMMTYPE_CONTROLLER_TO_ENGINE, // gBattleBufferB
+   BATTLELINKCOMMTYPE_CONTROLLER_BECOMING_IDLE
 };
```
GriffinRichards, line 117: These new names are a bit of a mouthful, esp. for how often they're used. At a minimum, we have other names that shorten `BATTLE` as a prefix to `B_`, and `LINKCOMM` could be shortened to `COMM` (which also avoids implying it's related to multiplayer battles. That's often what "link" is associated with)

DavidJCobb:
Good suggestions, thanks. Renamed those constants to match. Also added brief documentation comments for them.

PR #2126, opened 2025-04-27
Rename battlerId to battler for consistency

AlexOn1ine:
The var identifier `battler` is largely used in the files `battle_ai_main.c`, `battle_ai_util.c``battle_script_commands.c`, `battle_util.c` and `battle_script.inc` while most other files use `battlerId`. I would argue for consistency sake it should all be renamed to `battler`. 

Few reason why I think it should go from `battlerId` to `battler` and not the other way around:
1. By definition battler is already an id so the suffix invokes a double meaning
2. Global battler variables and function arguments `battlerAtk, battlerDef` don't use the suffix Id 
3. Derived from 2: Consistency


mrgriffin:
I agree with the thinking here, most interactions people have are with variables that just say "battler", e.g. `gActiveBattler`.

I wonder if all the battler usages are `u8`, if so we could have a `typedef u8 BattlerId;` and be even clearer. Usually when I look into that sort of idea it turns out that GF used a variety of integral types and so it's either not going to match, or I need `BattlerId8`, `BattlerId16`, etc and it becomes quite ugly.

If we did have a named type then downstream projects could probably make it a distinct type so that it's a compile error to mix battler IDs with other kinds of IDs (e.g. with something like `enum BattlerId : u8 {};`, or `struct __attribute__((packed, align(1))) BattlerId { u8 _; };`). Sadly I don't think anything like that would match in `pokeemerald` :( (_Pleeeaaase_ somebody prove me wrong :pray:)

AlexOn1ine:
There is one instance of `u32 battler`

AlexOn1ine:
> There is one instance of `u32 battler`. all other seems to use a u8

oh no there are more, sorry

PR #2128, opened 2025-04-28
Identified task and sprite fields related to battle weather animations

DavidJCobb:
Named the task and sprite fields related to the battle animations for Hail, Rain Dance, and Sandstorm. Tested via `make clean && make -j16 && make compare`; the hash matched.

## Description

* Identified `HailStruct::unk3` as a position type: fixed (i.e. always uses the position in the struct), negative (negative offset applied to a battler position, if battler is present; position in the struct otherwise), or positive. Renamed the field to `type`, and added an enum for the values.
* Used `#define` and `#undef` for the task data and sprite data for these three moves.
* Some integer constants, AFAIK, "should" be `ARRAY_COUNT`, but using that macro causes the build to fail to match, so I left the constants untouched and noted the equivalence in line comments.

## **Discord contact info**
`@davidjcobb`

### DIFF
--- a/src/battle_anim_ice.c
+++ b/src/battle_anim_ice.c
@@ -13,11 +13,16 @@
 #include "constants/battle_anim.h"
 #include "constants/rgb.h"
 
+enum {
+   HAILSTRUCTTYPE_NEGATIVE_POS_MOD = 0,
+   HAILSTRUCTTYPE_POSITIVE_POS_MOD = 1,
+   HAILSTRUCTTYPE_FIXED_POSITION   = 2,
+};
 struct HailStruct {
     s32 x:10;
     s32 y:10;
     s32 bPosition:8;
-    s32 unk3:4;
+    s32 type:4;
 };
 
 static void AnimUnusedIceCrystalThrow(struct Sprite *);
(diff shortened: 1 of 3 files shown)


src/battle_anim_ice.c
```diff
@@ -1342,71 +1354,86 @@ void AnimTask_Hail(u8 taskId)
...
-                task->data[1]++;
+            if (GenerateHailParticle(task->tHailStructId, task->tHailAffineAnimNum, taskId, 1))
+                task->tSpriteCount++;
 
-            if (++task->data[2] == 3)
+            if (++task->tHailAffineAnimNum == 3) // i.e. ARRAY_COUNT(sAffineAnims_HailParticle)
             {
-                if (++task->data[3] == 10)
-                    task->data[0]++;
+                if (++task->tHailStructId == 10) // i.e. ARRAY_COUNT(sHailCoordData)
```
GriffinRichards, line 1376: I'm assuming these are listed as comments because they didn't match? You can usually fix that by casting the result of `ARRAY_COUNT` to `int`

DavidJCobb:
Yep, casting to `int` fixed it. Thanks for the tip!

PR #2133, opened 2025-05-08
Implement abaresks bugfix to CalcBerryYieldInternal

Jaizu:
## Description
Implements bugfix from Abaresk to CalcBerryYieldInternal
Also updates the gist link to his new github username, maybe we should host the gist in pret or something else, but that's outside of my control
https://gist.github.com/abaresk/2a099dbe54aa91414decdbfaa524327d

## **Discord contact info**
671gz

### DIFF
--- a/src/berry.c
+++ b/src/berry.c
@@ -1211,10 +1211,22 @@ static u8 GetNumStagesWateredByBerryTreeId(u8 id)
 // However, this function actually skews towards higher berry yields, because
 // it rounds `extraYield` to the nearest whole number.
 //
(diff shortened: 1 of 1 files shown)


GriffinRichards:
Could you explain in more detail what the bug is and how this fixes it? I seem to recall there was some assumptions about this fix that weren't correct
https://discord.com/channels/442462691542695948/442465020291317760/1029611393693012018

GriffinRichards:
From you on Discord:
> I only copy pasted the original bug-fixes that were in the comment
if you think the bugfix is wrong feel free to close it

PR #2135, opened 2025-05-10
Add BattlerId enum and document

AlexOn1ine:
I rarely looked into the file so not 100% sure if they this is correct but I think it should be. It also matches.


include/constants/battle.h
```diff
@@ -23,12 +23,23 @@
...
-#define B_POSITION_PLAYER_LEFT        0
-#define B_POSITION_OPPONENT_LEFT      1
-#define B_POSITION_PLAYER_RIGHT       2
-#define B_POSITION_OPPONENT_RIGHT     3
+enum BattlerId
+{
+    B_BATTLER_PLAYER_LEFT,
+    B_BATTLER_OPPONENT_LEFT,
+    B_BATTLER_PLAYER_RIGHT,
+    B_BATTLER_OPPONENT_RIGHT,
```
mrgriffin, line 40: I think maybe these should have vague names because battler ID 0 isn't necessarily anything to do with player left, etc?
```suggestion
    B_BATTLER_0,
    B_BATTLER_1,
    B_BATTLER_2,
    B_BATTLER_3,
```
My reasoning is that in the PR we have lines like this:
```gBattlerControllerFuncs[B_BATTLER_OPPONENT_LEFT] = SetControllerToRecordedPlayer;```
Which imo are quite surprising, because it's setting "opponent left" to "player". With the vague names that would be:
```gBattlerControllerFuncs[B_BATLER_1] = SetControllerToRecordedPlayer;```

**Review by mrgriffin — CHANGES_REQUESTED** [review]

File: include/constants/battle.h (hunk 1 above)
Comment (AlexOn1ine) [reply include/constants/battle.h]: oops, this makes much more sense.

Suggestion applied

File: include/battle.h [hunk 2]
```diff
@@ -636,7 +636,7 @@ extern u8 gActiveBattler;
 extern u32 gBattleControllerExecFlags;
 extern u8 gBattlersCount;
 extern u16 gBattlerPartyIndexes[MAX_BATTLERS_COUNT];
-extern u8 gBattlerPositions[MAX_BATTLERS_COUNT];
+extern u8 gBattlerPositions[MAX_POSITION_COUNT];
```
mrgriffin, line 639: Rereading this, I think you've got this backwards. In `gBattlerPositions[i]`, `i` should be an `enum Battler` because we're going from battler to position, so `[MAX_BATTLERS_COUNT]` is correct.

If we had a reverse mapping, from position to battler I think that would be `[MAX_POSITION_COUNT]` (but I think `GetBattlerAtPosition` doesn't actually use an array to map).

AlexOn1ine: Ye, reverted change

mrgriffin (approved):
lgtm :+1: 

DavidJCobb's message in Discord is interesting:
> though personally if it were me i *might* perhaps do names like `B_BATTLER_TEAM_1_LEFT`, `B_BATTLER_TEAM_2_RIGHT`, and so on, so that the "symmetry" between IDs and positions is clearer (i.e. the fact that things like `BATTLER_OPPOSITE(b)` work the same on both)

But I'm not sure about that rn, so we can think about in the future :)

Looking forward to being able to write `enum BattlerPosition gBattlerPositions[MAX_BATTLERS_COUNT];` in Expansion :grin:

PR #2139, opened 2025-05-13
Streamline pointer notation

Bassoonian:
Streamlines pointer notation by having the asterisks always follow the type with a space and having them directly precede the variable name.

## **Discord contact info**
bassoonian


mrgriffin (approved):
Admittedly I have relied on skimming, `make compare`, and a confidence that this PR was made with a script of some sort, rather than verifying it by reading every single time in great detail.

PR #2142, opened 2025-05-17
Document cCURSOR_POSITION

AsparagusEduardo:
## Description
Title

## **Discord contact info**
AsparagusEduardo

### DIFF
--- a/data/battle_scripts_1.s
+++ b/data/battle_scripts_1.s
@@ -2836,7 +2836,7 @@ BattleScript_HandleFaintedMon::
 	printstring STRINGID_USENEXTPKMN
 	setbyte gBattleCommunication, 0
 	yesnobox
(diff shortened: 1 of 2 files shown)


data/battle_scripts_1.s
```diff
@@ -4432,17 +4432,17 @@ BattleScript_FlushMessageBox::
 	return
 
 BattleScript_PalacePrintFlavorText::
-	setbyte gBattleCommunication + 1, 0
+	setbyte cCURSOR_POSITION, 0
```
GriffinRichards, line 4435: Could you explain what this has to do with the cursor here? It looks to me like this is used as a loop variable

AlexOn1ine: isn't all in this pr used as a loop var?

PR #2143, opened 2025-05-18
Update header files to match source files

Bassoonian:
Many declarations in the header files were outdated, containing no argument names, containing different names for arguments, or sometimes even errors (`src` for `dest`, among others). This PR updates them to match their declarations in the source files.

## **Discord contact info**
bassoonian


mrgriffin (approved):
Nice!

I think there are a few examples where imo it's not really necessary to name the arguments, e.g. the `MetatileBehavior_IsFoo` functions or the ones that take a `struct Sprite *`. But I think the consistency is probably helpful for the users who aren't familiar with C :)

PR #2144, opened 2025-05-19
Correct C23 Checks

Kurausukun:
## Description
Previous changes ignored cases where a modern compiler that did not yet support C23 was used. Now we explicitly check the `__STDC_VERSION__` macro.

## **Discord contact info**
kurausukun


mrgriffin (approved):
Perfect, thanks for the swift fix Kura :)

PR #2146, opened 2025-05-22
Remove files that were wrongly commited

AlexOn1ine:
Wrongly committed from #2116


mrgriffin:
`src/data/map_group_count.h` needs to be deleted too

GriffinRichards:
We could add these files to pokeemerald's gitignore, I think this is round 3 of a PR adding or almost adding them.

mrgriffin:
> We could add these files to pokeemerald's gitignore, I think this is round 3 of a PR adding or almost adding them.

Good idea!

PR #2163, opened 2025-07-21
Add pack/unpack for data masks

GriffinRichards:
- Adds packing/unpacking macros for map data and metatile attribute data. There were 2 intentions here:
    - At the moment it's difficult to reassign bits on the right side of `MAPGRID_METATILE_ID_MASK` / `METATILE_ATTR_BEHAVIOR_MASK` because they have no shift value to update, and adding a shift value makes some simple code uglier. This adds a shift value for them and hides the extra complication behind the macros.
    - This should hopefully make it easier for someone downstream to utilize something like `__builtin_clz` to get rid of the explicitly-defined shifts altogether.
- Adds `MAPGRID_IMPASSABLE` as an alias for `MAPGRID_COLLISION_MASK` when it's used to set impassability for a map space.
- Adds a `BUGFIX` in `decoration.c` related to layer type usage (arguably not a bug, but a pitfall we have that would only be fixed by making the documentation worse)

### DIFF
--- a/include/global.fieldmap.h
+++ b/include/global.fieldmap.h
@@ -7,18 +7,35 @@
 #define MAPGRID_METATILE_ID_MASK 0x03FF // Bits 0-9
 #define MAPGRID_COLLISION_MASK   0x0C00 // Bits 10-11
 #define MAPGRID_ELEVATION_MASK   0xF000 // Bits 12-15
+#define MAPGRID_METATILE_ID_SHIFT 0
 #define MAPGRID_COLLISION_SHIFT  10
 #define MAPGRID_ELEVATION_SHIFT  12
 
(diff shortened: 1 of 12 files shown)


mrgriffin:
Great change! Just one comment, `GetLayerType` implicitly adds `NUM_TILES_IN_PRIMARY` because the decoration tiles are always in reference to the secondary tileset's tiles... But I think that fact might be too subtle in this code:
```c
layerType = GetLayerType(decoration->tiles[(mapY - 1 - i) * mapX + j]);
```
My weakly-held opinion is that maybe the `NUM_TILES_IN_PRIMARY +` part could come back, or perhaps renaming `decoration->tiles` to something like `decoration->secondaryTiles`?

PR #2200, opened 2025-11-06
Implicit waitstate for special/specialvars

mrgriffin:
`special` and `specialvar`s that are always followed by a `waitstate` now implicitly generate that `waitstate` when called. This should make it easier to use those specials for newcomers who have not internalized "a UI special needs a `waitstate` after it to stop the script progressing to a disastrous result".

If an implicit `waitstate` is immediately followed by an explicit `waitstate` the second one is ignored and a warning is reported, thus making the change harmless for downstream projects (if a little noisy). If there are other instructions between the implicit and explicit `waitstate`s then both are generated, which would manifest as a softlock; I think writing code like that is unidiomatic, so hopefully this will not affect (m)any downstream projects.

The candidate `special`s were mostly identified with the following: `git grep -B1 waitstate data | grep -v '^-' | sed 'N;s/\n/ /;p' | grep -o 'special [a-zA-Z0-9_]*' | sort -u`.

Notable `special`s that are not always followed by `waitstate` are `EnterSecretBase`, `RunUnionRoom`, and `ShakeCamera`. Of these, `ShakeCamera` is particularly disappointing because 22/24 of the uses are immediately followed by `waitstate`. In the future we could consider extending `special` to support this use-case.

Discussed in the [pret](https://discord.com/channels/442462691542695948/442465020291317760/1435764841594880202) and [rh-hideout](https://discord.com/channels/419213663107416084/1077168246555430962/1435760564130418830) Discord servers.

### DIFF
--- a/asm/macros/event.inc
+++ b/asm/macros/event.inc
@@ -1,6 +1,11 @@
 	.set ALLOCATE_SCRIPT_CMD_TABLE, 0
 	.include "data/script_cmd_table.inc"
 
+	.set ALLOCATE_SPECIAL_TABLE, 0
+	.include "data/specials.inc"
+
+	.set _last_implicit_waitstate, -1
+
 	@ Does nothing.
 	.macro nop
 	.byte SCR_OP_NOP
@@ -272,6 +277,9 @@
 	.macro special function:req
 	.byte SCR_OP_SPECIAL
 	.2byte SPECIAL_\function
+	.if SPECIAL_WAITSTATE_\function
+	waitstate implicit=1
+	.endif
 	.endm
 
 	@ Calls a function listed in the table in data/specials.inc.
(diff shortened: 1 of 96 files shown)


mrgriffin:
Wrote a script to automate the deleting of any explicit `waitstate`s that report a warning: [delete_implicit_waitstates.py](https://github.com/user-attachments/files/23394981/delete_implicit_waitstates.py). Usage `python3 delete_implicit_waitstates.py`. **MAKES CHANGES TO YOUR WORKING DIRECTORY, COMMIT YOUR CHANGES BEFORE RUNNING**.

I reverted the changes to `data/` (except `data/event_scripts.s` and `data/specials.inc`) from this PR, ran the script, and ended up at the same state as this PR is is.

GriffinRichards (approved):

PR #2203, opened 2025-11-10
Added bugfix for missing no weather case in Cmd_weather_get

Mitsunee:
Adds a bug fix for `Cmd_weather_get` to prevent previous values being interpreted as weather.

## Description

It has been discovered that in the case of there not being a weather set `Cmd_weather_get` will fail to write any value into `AI_THINKING_STRUCT->funcResult`. This can cause a value left by the previous AI function to be interpreted as a result of `Cmd_weather_get` if the value happens to be between 0 and 3 (inclusive).

Here is an example of this bug in real gameplay happening with a Glalie set in the Battle Frontier, alongside some more detail on the research in a youtube video:

[YouTube Video](https://youtube.com/watch?v=RNuzKcdP5YQ)

Up for discussion is what the value of `AI_WEATHER_NONE` should be. I have chosen `99` as it is well outside of the range of already existing values and `0` is already used for sunny weather. The most important part is that a value that does not match any other `AI_WEATHER_*` is set, to prevent the bug this is meant to fix.

## **Discord contact info**
Discord: `Mitsunee`
Fedi: `@mitsunee@mk.absturztau.be`

### DIFF
--- a/include/constants/battle_ai.h
+++ b/include/constants/battle_ai.h
@@ -23,7 +23,7 @@
 #define AI_EFFECTIVENESS_x0     0
 
 // ai weather
-#define AI_WEATHER_NONE 99 // added as part of bugfix, see Cmd_weather_get in battle_ai_script_commands.c
+#define AI_WEATHER_NONE UINT32_MAX // added as part of bugfix, see Cmd_weather_get in battle_ai_script_commands.c
 #define AI_WEATHER_SUN 0
 #define AI_WEATHER_RAIN 1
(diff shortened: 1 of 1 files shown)


ketsuban:
I maintain that using zero for "no weather" is the correct thing to do because "no weather" is not an error condition, but I see I'm not going to win that one so I would back the suggestion to use `-1`/`0xFFFFFFFF` as the least arbitrary value for `AI_WEATHER_NONE` that doesn't change the other `AI_WEATHER_*` constants.

PR #2210, opened 2025-11-26
Rename AttacksThisTurn to GetAttackTiming for readability

Mitsunee:
Adds enum for `AttacksThisTurn` and renames to `GetAttackTiming`.

## Description

`AttacksThisTurn` is a bit of a confusing name and the function returns a magic number. Adding an enum and renaming it to match makes its usage a lot more clear.

## **Discord contact info**
Mitsunee

### DIFF
--- a/src/battle_script_commands.c
+++ b/src/battle_script_commands.c
@@ -61,11 +61,12 @@ extern const u8 *const gBattleScriptsForMoveEffects[];
 
 #define TAG_LVLUP_BANNER_MON_ICON 55130
 
-#define ATTACKS_THIS_TURN(battler, move) (AttacksThisTurn(battler, move) == 2)
+#define TIMING_CHARGING  1
+#define TIMING_ATTACKING 2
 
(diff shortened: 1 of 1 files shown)


src/battle_script_commands.c
```diff
@@ -61,6 +61,8 @@ extern const u8 *const gBattleScriptsForMoveEffects[];
 
 #define TAG_LVLUP_BANNER_MON_ICON 55130
 
+#define ATTACKS_THIS_TURN(battler, move) (AttacksThisTurn(battler, move) == 2)
```
AlexOn1ine: This macro has the same name as the function which doesn't make it much better imo.

Instead I would suggest defines/enums for the 1 and 2.

Something like:
```
#define TWO_TURN_MOVE_CHARGES 1
#define TWO_TURN_MOVE_ATTACKS 2
```

That removes the magic numbers altogether and improves readability.

mrgriffin:
@AlexOn1ine did you have any extra feedback for this PR, or can I merge it? :)

PR #2238, opened 2026-03-01
Add gBattleOutcome == B_OUTCOME_DREW to bugfix

SiliconA-Z:
Mirroring the change that Nintendo made when rereleasing FRLG on Switch, which addresses the Roamer bug, they also added a check for B_OUTCOME_DREW.

### DIFF
--- a/src/battle_main.c
+++ b/src/battle_main.c
@@ -5234,10 +5234,13 @@ static void ReturnFromBattleToOverworld(void)
         UpdateRoamerHPStatus(&gEnemyParty[0]);
 
 #ifndef BUGFIX
+        // Bug: When Roar is used by a roamer, gBattleOutcome is B_OUTCOME_PLAYER_TELEPORTED (5),
+        // which deactivates the roamer.
         if ((gBattleOutcome & B_OUTCOME_WON) || gBattleOutcome == B_OUTCOME_CAUGHT)
 #else
-        if ((gBattleOutcome == B_OUTCOME_WON) || gBattleOutcome == B_OUTCOME_CAUGHT) // Bug: When Roar is used by roamer, gBattleOutcome is B_OUTCOME_PLAYER_TELEPORTED (5).
-#endif                                                                               // & with B_OUTCOME_WON (1) will return TRUE and deactivates the roamer.
+        if ((gBattleOutcome == B_OUTCOME_WON) || gBattleOutcome == B_OUTCOME_CAUGHT ||
+            gBattleOutcome == B_OUTCOME_DREW)
+#endif
             SetRoamerInactive();
     }
 


src/battle_main.c
```diff
@@ -5234,9 +5234,9 @@ static void ReturnFromBattleToOverworld(void)
         UpdateRoamerHPStatus(&gEnemyParty[0]);
 
 #ifndef BUGFIX
-        if ((gBattleOutcome & B_OUTCOME_WON) || gBattleOutcome == B_OUTCOME_CAUGHT)
+        if ((gBattleOutcome & 1) || gBattleOutcome == B_OUTCOME_CAUGHT) // All odd outcomes, including B_OUTCOME_CAUGHT, making this redundant.
```
ketsuban: * Removing the `B_OUTCOME_WON` symbolic constant here doesn't seem to serve any purpose.
* The comment is a bit strange and doesn't really add anything. It would I think make sense to move the comment describing the bug to this side of the `#if` block since this is the code that *has* the bug.

SiliconA-Z: B_OUTCOME_WON is 1, which means every odd condition in the defined would trigger deactivation, including draw, forfieted, etc.

SiliconA-Z: Addressed!

ketsuban (approved):

SiliconA-Z:
Done!

ketsuban (approved):

SiliconA-Z:
Drew exists because if you draw against a roamer that roamer will come back with 0 hp and cause all sorts of problems.

mrgriffin:
@ketsuban sorry for the ping, I saw you approved this PR—have you tested it locally to verify that it works and doesn't have any unintended consequences?

ketsuban:
As mentioned on Discord it matches the Switch rerelease, and @SiliconA-Z has already identified the bug it fixes which the original bugfix didn't account for, so I don't think I need to do any independent testing.

mrgriffin:
Has somebody independently verified that it matches the Switch release?

I'm not willing to take it on trust, but if you don't/can't/haven't/etc, then we can wait and see if a different maintainer is willing to :)

LumioseSil:
Yes. Check out pret/pokefirered. It's merged

LumioseSil:
So don't take my word for it.

https://github.com/pret/pokefirered/pull/722

LumioseSil:
Proof it is matching image

LumioseSil:
Anyway, an explanation of what drew does:

Without it, you end up with a situation (thanks Deo for repro on Ruby) where you encounter the roamer with 0 HP and then the game glitches out.

This is actually the bug localization tried to fix but ended up introducing the Roar bug, given its absence from the Japanese version and international Ruby and its change in the switch release of FRLG to include this condition.

PR #2251, opened 2026-03-24
Put library ro data with the other rodata in MODERN.

AZero13:
Most of these rules are dead and not followed: the linker takes everything with the wildcard. So we can optimize this.


AZero13:
Ping

AZero13:
Most of these rules are dead and not followed: the linker takes everything with the wildcard. So we can optimize this.

PR #2252, opened 2026-03-24
Edit NUM_GROUPTYPES UBFIX by mirroring Nintendo's FRLG Fix

AZero13:
By doing type < NUM_GROUPTYPES, we ensure that the type is a valid index into groupCounts.

Yes, we break out early, but this is correct (and what FRLG revision A does anyway),:

sActivityGroupInfo is effectively a lookup table.

GROUPTYPE_NONE is only used for ACTIVITY_SEARCH.

That row should contribute nothing to groupCounts.


AZero13:
Ping

AZero13:
PR used for reference:

https://github.com/pret/pokefirered/pull/744

AZero13:
Ping.

AZero13:
Ping

PR #2253, opened 2026-03-24
Simplify m4a functions

AZero13:
Improved readability of the m4a.c file. Used unions where they work and match. 

Added a dummy for the sake of making it explicit because offset calculation in assembler in m4a_1 is a thing.


AZero13:
A lot of these are cosmetic but also good. I hope these are acceptable, especially the changes to the CGB functions

AZero13:
Ping

PR #2254, opened 2026-03-25
Fix type punning that relies on undefined behavior in battle_transition.c

AZero13:


AZero13:
The merged value optimization is a memory access optimization that merges stores together. 

However if the struct is changed, this is UB. Just fix this. 

NO UBFix at all. This matches exactly without UB.

mrgriffin:
Why do many of your PRs come with more unrelated changes than fixes? It was hard to even find the lines that had any type punning, afaict there's 3 removed, 1 added to do with `mergedValue`.

Not to mention some of the unrelated changes are questionable, like `/ 256` becoming `>> 8` which I assume would be better documented as `Q_8_8_TO_INT` or some variant of that.

Recently I went through all the open PRs and merged all the ones that were obviously improvements with no downsides, some of yours got merged, but many others weren't obviously improvements, or not downside-free; e.g. too much to read, so it's not obvious; contains questionable changes; etc.

EDIT: There's no need to respond to this message, it begins with a rhetorical question.

PR #2259, opened 2026-03-28
Fix SetObjectSubpriorityByElevation fakematch

AZero13:
Same thing as done in pokefirered.


AZero13:
Literally just https://github.com/pret/pokefirered/pull/737 but emerald

PR #2263, opened 2026-03-30
Fix all fakematches in event object movement

AZero13:
Min changes needed. This also required adjusting faraway_mew's return type.


AZero13:
Had to resolve conflicts, so a force push was needed. Sorry.

ketsuban (approved):

AZero13:
Thanks for approving. Any other issues or concerns @Kurausukun

AZero13:
Ping

AZero13:
ping

AZero13:
hi mgriffin

ketsuban (commented):
* In general, consider using `direction` as the name of a function argument rather than `dir` in cases where a local variable named `direction` was removed.
* There's some cases where a multiplication by an unexplained constant is replaced by a left-shift by an equally unexplained constant; one representation is probably more meaningful than the other (e.g. `8` might be the width or height of a tile in pixels) and the constant should be given a name to reflect that.

AZero13:
Fixed!

ketsuban (commented):
I no longer have the ability to formally approve changes in review but I have no further comments. LGTM 👍

PR #2265, opened 2026-04-02
gbagfx: replace for loop with snprintf

AZero13:
This also fixes the problem where we malloc one byte too little. We need to account for the null terminator.


AZero13:
Ping

PR #2280, opened 2026-04-07
Rearrange press_start.png

mrgriffin:
Rerranged to match the segments as defined by `sStartCopyrightBannerAnimTable` and added a `-num_tiles` to handle the trailing data. Arguably the whole third line of graphics shouldn't be there since it's just zeros, but I tried deleting that line and afaict `gbagfx` doesn't support filling the difference between the actual tiles and `-num_tiles` with transparency:
```
tools/gbagfx/gbagfx graphics/title_screen/press_start.png graphics/title_screen/press_start.4bpp -mwidth 4 -mheight 1 -num_tiles 48 -Wnum_tiles
The specified number of tiles (48) is greater than the maximum possible value (40).
```


mrgriffin:
Fixed conflicts with `INCGFX` :)

PR #2284, opened 2026-04-09
Make data1 and learnMoveState an array to prevent UB

### LINKED ISSUE #2102: Bad matching(?) in src/party_menu.c
https://github.com/pret/pokeemerald/blob/95aa41f0d42e1089e4bb082a614cab02419ad701/src/party_menu.c#L4746 this is UB, what it is doing is modifying learnMoveState. that might mean that the function was not matched correctly

AZero13:
Fixes https://github.com/pret/pokeemerald/issues/2102

### DIFF
--- a/include/party_menu.h
+++ b/include/party_menu.h
@@ -16,8 +16,8 @@ struct PartyMenu
     s8 slotId2;
     u8 action;
     u16 bagItem;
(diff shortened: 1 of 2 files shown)


AZero13:
Ping? Doesn't require UBFIX by the way. Or BUGFIX.

LumioseSil:
Any update?

AZero13:
Ping

PR #2288, opened 2026-05-01
Rename mass outbreak show  to distinguish from savefile outbreakDaysLeft

FosterProgramming:
Once you get an outbreak from random chance or mix, you need to wait one day before having a chance for the tv show to appear and start the mass outbreak. (you need to roll this show if you have other shows active and if you already have an active outbreak, you will need for it to end first)

The massOutbreak has its own day timer and so after 2 days, the mass outbreak will end. 
I was working with outbreak and was getting confused by the values having the same name (number of days before the outbreaks vs number of days the outbreak lasts) so I renamed the tv show one

## **Discord contact info**
Jamie

### DIFF
--- a/include/global.tv.h
+++ b/include/global.tv.h
@@ -486,7 +486,7 @@ typedef union // size = 0x24
         /*0x13*/ u8 probability;
         /*0x14*/ u8 level;
         /*0x15*/ u8 unused5;
-        /*0x16*/ u16 daysLeft;
+        /*0x16*/ u16 daysBeforeOutbreak;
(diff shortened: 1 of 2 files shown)


GriffinRichards:
Looks like you forgot to rename the declaration for `UpdateMassOutbreakTimeLeft`. Please make sure your changes compile locally before opening a PR.

FosterProgramming:
> Looks like you forgot to rename the declaration for `UpdateMassOutbreakTimeLeft`. Please make sure your changes compile locally before opening a PR.

Fixed and sorry

PR #2291, opened 2026-05-08
INCGFX performance - define individual Make variables instead of appending to list

ravepossum:
## Description
The changes to `scaninc` introduced in #2283 increased build times noticeably. On my machine, a `make` with zero changes went from ~0.9 seconds to ~3 seconds. The effect was much more pronounced on downstream forks with additional graphics like pokeemerald-expansion, where a zero-change build took my machine ~20 seconds to run.

The time to read all of the Make rules was increased significantly by incorporating the following pattern for each INCGFX rule:
```
ifeq (,$(findstring foo,$(_INCGFX)))
_INCGFX += foo
foo: bar
endif
```

Which was intended to suppress the warning Make creates when you have multiple recipes for a single target. My fix takes an alternative approach to accomplish that in a more performant way:

```
ifndef foo
foo := defined
foo: bar
endif
```

With my fix, a zero-change build is back down to ~0.9 seconds. Building pokeemerald-expansion with zero changes is down to ~1.75 seconds.

The only caveat I see here is that it introduces a potential for duplicate recipes to be created by having two files with the same exact path except for any special characters needed to sanitize the target as a Make variable. E.g. files named `pokeball.png` and `pokeball=.png` in the same directory would create an issue. Given that most people aren't putting `=`, `#`, or `:` in their file names, and that the issue only creates a Make warning, I'm not too concerned about it. 

## **Discord contact info**
raveposum

### DIFF
--- a/tools/scaninc/scaninc.cpp
+++ b/tools/scaninc/scaninc.cpp
@@ -18,6 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include <algorithm>
 #include <cctype>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


mrgriffin:
Sorry, only just noticed that CI doesn't run automatically for you (yet!). GitHub is being useless and not showing me the logs on the build page, so I scoured the raw logs and found:
```
2026-05-11T05:04:33.0982443Z scaninc.cpp: In function ‘int main(int, char**)’:
2026-05-11T05:04:33.0983837Z scaninc.cpp:236:22: error: ‘remove_if’ is not a member of ‘std’; did you mean ‘remove_cv’?
2026-05-11T05:04:33.0989140Z   236 |                 std::remove_if(
2026-05-11T05:04:33.0989493Z       |                      ^~~~~~~~~
2026-05-11T05:04:33.0989808Z       |                      remove_cv
2026-05-11T05:04:34.2018124Z make[1]: *** [Makefile:21: scaninc] Error 1
```
A missing `#include <algorithm>`? A standard library that's too old to support `remove_if` with a predicate?

ravepossum:
Thanks! Since it compiled just fine locally, I'm inclined to assume the latter. I'll investigate.

hedara90:
I can confirm that adding
```C
#include <algorithm>
```
makes it work.

Maybe it's some difference between `gcc` and `clang`?

ravepossum:
Force pushed and added the `<algorithm>` inclusion - needs another maintainer poke to run the CI.

PR #2300, opened 2026-06-07
Remove struct offset comments

hedara90:
## Description

Remove offset comments from structs and from byte-script declarations using `enum`ified byte values.

## **Discord contact info**

hedara


hedara90:
Closing after discussion

PR #2312, opened 2026-06-14
Bugfix: Typo in IncrementPalaceStreak causes wrong streak to be checked

AZero13:
Gamefreak forgot what precedence operators were. I edited the code to make the bug more obvious and still match.


AZero13:
Ping

PR #2315, opened 2026-06-16
Make GetMonData and GetBoxMonData ABI-safe

AZero13:
Clang and LLVM hate this, so instead I am doing the manual work of using macros to do the equivalent of a default parameter in C++ to avoid work downstream.

### DIFF
--- a/include/pokemon.h
+++ b/include/pokemon.h
@@ -441,12 +441,29 @@ void SetMultiuseSpriteTemplateToTrainerFront(u16 trainerPicId, u8 battlerPositio
  * either Get(Box)MonData2 or Get(Box)MonData3 based on the number of
  * arguments. The two functions are aliases of each other, but they
  * differ for matching purposes in the caller's codegen. */
-#define GetMonData(...) CAT(GetMonData, NARG_8(__VA_ARGS__))(__VA_ARGS__)
-#define GetBoxMonData(...) CAT(GetBoxMonData, NARG_8(__VA_ARGS__))(__VA_ARGS__)
-u32 GetMonData3(struct Pokemon *mon, s32 field, u8 *data);
+#ifndef UBFIX
+#define DISPATCH_MATCH(FUNC, ...) CAT(FUNC, NARG_8(__VA_ARGS__))(__VA_ARGS__)
+
+#define GetMonData(...)    DISPATCH_MATCH(GetMonData, __VA_ARGS__)
+#define GetBoxMonData(...) DISPATCH_MATCH(GetBoxMonData, __VA_ARGS__)
+
+// agbcc still needs to know about the 2-argument aliases
 u32 GetMonData2(struct Pokemon *mon, s32 field);
-u32 GetBoxMonData3(struct BoxPokemon *boxMon, s32 field, u8 *data);
 u32 GetBoxMonData2(struct BoxPokemon *boxMon, s32 field);
+#else
+#define GetMonData_(mon, field, data, ...) (GetMonData)(mon, field, data)
+#define GetMonData(...) GetMonData_(__VA_ARGS__, NULL, NULL)
+
+#define GetBoxMonData_(boxMon, field, data, ...) (GetBoxMonData)(boxMon, field, data)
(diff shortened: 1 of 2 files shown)


include/pokemon.h
```diff
@@ -441,12 +441,33 @@ void SetMultiuseSpriteTemplateToTrainerFront(u16 trainerPicId, u8 battlerPositio
  * either Get(Box)MonData2 or Get(Box)MonData3 based on the number of
  * arguments. The two functions are aliases of each other, but they
  * differ for matching purposes in the caller's codegen. */
```
ketsuban, line 443: The comment has become orphaned by this change, referring to functions `GetMonData3` and `GetBoxMonData3` which no longer exist.

ketsuban (changes requested):

ketsuban (approved):

src/pokemon.c
```diff
@@ -48,6 +48,11 @@
 #include "constants/trainers.h"
 #include "constants/union_room.h"
 
+#ifndef UBFIX
+u32 GetMonData2(struct Pokemon *mon, s32 field) __attribute__((alias("GetMonData3")));
+u32 GetBoxMonData2(struct BoxPokemon *boxMon, s32 field) __attribute__((alias("GetBoxMonData3")));
+#endif
```
mrgriffin: How does this work? Aren't these aliasing symbols that aren't defined?

mrgriffin (changes requested):
Do we need all these changes? Couldn't it have been more like this in `include/pokemon.h`?
```diff
 u32 GetMonData3(struct Pokemon *mon, s32 field, u8 *data);
-u32 GetMonData2(struct Pokemon *mon, s32 field);
 u32 GetBoxMonData3(struct BoxPokemon *boxMon, s32 field, u8 *data);
-u32 GetBoxMonData2(struct BoxPokemon *boxMon, s32 field);
+#if UBFIX
+#define GetMonData2(mon, field) GetMonData3(mon, field, NULL)
+#define GetBoxMonData2(boxMon, field) GetBoxMonData3(boMon, field, NULL)
+#else
 u32 GetMonData2(struct Pokemon *mon, s32 field);
 u32 GetBoxMonData2(struct BoxPokemon *boxMon, s32 field);
+#endif
```
And then just the conditional declaration of aliases in `src/pokemon.c`, i.e. without renaming `GetMonData3`/`GetBoxMonData3`.

Also, have you picked `UBFIX` because on a different (non-ARM) ABI the behavior would be undefined? I don't think it could cause problems on AArch32 unless the functions took 5+ parameters?

src/pokemon.c
AZero13: > How does this work? Aren't these aliasing symbols that aren't defined?

An alias does not create a wrapper function. It doesn't create new assembly code that calls the old function. It is a zero-cost linker trick.

This is for the linker.

AZero13: Without it the compilation will fail on not UBFIX. But because this is not UBFIX and it compiles matxhing on not ubfix anyway, it shouldn't matter.

mrgriffin: Right, but the `GetMonData3` and `GetBoxMonData3` symbols did not appear to be defined on the version of the code I was commenting on because you'd unconditionally renamed them to `GetMonData` and `GetBoxMonData`.

AZero13: In pokemon.h, there is a macro: #define GetMonData(...) ... which automatically adds a 3 to the end of the name if 3 arguments are passed.

In pokemon.c, I define your function as: u32 GetMonData(struct Pokemon *mon, s32 field, u8 *data)

Because I passed 3 arguments, the C preprocessor instantly rewrites your function definition to be: u32 GetMonData3(struct Pokemon *mon, s32 field, u8 *data).

When agbcc compiles the file, the symbol that gets saved into the object file is GetMonData3.

mrgriffin: I see! That's awful.

AZero13: It's the best I could do. It sucks but

AZero13:
Redundant.

PR #2307, opened 2026-06-12
Put tilemapBuffer in PokenavListSub, and rename PokenavListSub to PokenavListView

AZero13:
PokenavListView best describes everything since it is related to graphics and what it shown.

PR #2317, opened 2026-06-17
Remove PokenavListSub

mrgriffin:
Inspired-by and an-alternative-to #2307, `struct PokenavListSub` is removed rather than renamed.

This is probably a fakematch because `LoopedTask_ReshowListFromCheckPage` has something weird going on with `list` and `listAlias` pointing to the same object, but I think it makes the code clearer? Or at least, I couldn't work out what the pattern between a member going in `struct PokenavList` and `struct PokenavListSub` is supposed to be.

Additionally, fix some fakematches in `LoopedTask_ReshowListFromCheckPage`.

### DIFF
--- a/src/pokenav_list.c
+++ b/src/pokenav_list.c
@@ -211,31 +211,31 @@ static bool32 IsPrintListItemsTaskActive(void)
 static u32 LoopedTask_PrintListItems(s32 state)
 {
     u32 row;
-    struct PokenavList *listSub = GetSubstructPtr(POKENAV_SUBSTRUCT_LIST);
+    struct PokenavList *list = GetSubstructPtr(POKENAV_SUBSTRUCT_LIST);
 
     switch (state)
     {
     case 0:
-        row = (listSub->listWindow.unkA + listSub->listWindow.numPrinted + listSub->printStart) & 0xF;
-        listSub->bufferItemFunc(listSub->listPtr, listSub->itemTextBuffer);
-        if (listSub->iconDrawFunc != NULL)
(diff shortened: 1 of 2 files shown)


src/pokenav_list.c
```diff
@@ -214,7 +211,7 @@ static bool32 IsPrintListItemsTaskActive(void)
 static u32 LoopedTask_PrintListItems(s32 state)
 {
     u32 row;
-    struct PokenavListSub *listSub = GetSubstructPtr(POKENAV_SUBSTRUCT_LIST);
+    struct PokenavList *listSub = GetSubstructPtr(POKENAV_SUBSTRUCT_LIST);
```
GriffinRichards: `listSub`->`list`

src/pokenav_list.c
```diff
@@ -55,36 +57,31 @@ struct PokenavListSub
...
-};
-
-struct PokenavList
-{
-    struct PokenavListSub sub;
     u8 tilemapBuffer[BG_SCREEN_SIZE];
     struct PokenavListWindowState windowState;
     s32 eraseIndex;
-    u32 loopedTaskId;
+    u32 loopedTaskId2;
```
GriffinRichards: I'm not attached to dealing with this now (if it can be), but it'd be nice if there was a meaningful distinction here.

mrgriffin: I'll have a look to see if there's better names I can give them :)

mrgriffin: `loopedTaskId` was only used for storing a particular task's ID, so I renamed it to `moveListWindowLoopedTaskId` (a bit of a mouthful, I know!) and then the other is used for multiple tasks but it can be `loopedTaskId`.

PR #2319, opened 2026-06-17
Clean up some decompisms from item.c

AZero13:
MoveItemSlotInList also has a UBFIX because casting from u32 to s16 has problems. Just use memmove instead.


mrgriffin:
> `MoveItemSlotInList` also has a `UBFIX` because casting from `u32` to `s16` has problems.

Would you need more than 32k slots in a single pocket cause a problem in practice?

PR #2320, opened 2026-06-17
Always call Get(Box)MonData3 in UBFIX builds

mrgriffin:
Extracted from a comment on #2315.


mrgriffin:
Unjustifiably more complex than #2321.

PR #2322, opened 2026-06-18
Implicit waitmovement 0

mrgriffin:
Convenience macros for eliminating some explicit `waitmovement 0`s:
1. `applywaitmovement`: Fused `applymovement` and `waitmovement 0`.
2. Implicit `waitmovement 0` for `moverotatingtileobjects` and `turnrotatingtileobjects`.

It is possible to create a variadic `applywaitmovements` macro, but I didn't think it would be practical because the lines would become very long. The remaining `waitmovement`s are now mostly:
1. After `special PlayerFaceTrainerAfterBattle` (could not add an implicit `waitmovement` here because not all calls to that special immediately `waitmovement`).
2. After two or more `applymovement`s.

Migration script for `applywaitmovement`: [migrate_applywaitmovement.py](https://github.com/user-attachments/files/29078395/migrate_applywaitmovement.py).
**WARNING:** if there's complex control flow the linear "is it appropriate to fuse the `waitmovement` with the previous `applymovement`" might fuse when there's technically two or more `applymovement`s and an explicit `waitmovement 0` would be more idiomatic. It should not produce an _incorrect_ result, but as always try comparing the `sha1sum` of your ROM before/after migration.


asm/macros/event.inc
```diff
@@ -637,7 +638,10 @@
 	@ If localId is LOCALID_NONE (0), then the id of the last-moved object will be used instead.
 	@ If the specified object is not currently being manipulated with applymovement, then this command does nothing.
 	@ If no map is specified, then the current map is used.
-	.macro waitmovement localId=LOCALID_NONE, map
+	.macro waitmovement localId=LOCALID_NONE, map, implicit=0
+		.if _last_implicit_waitmovement == . && \localId == LOCALID_NONE
+		.warning "explicit waitmovement follows implicit waitmovement, ignoring"
```
GriffinRichards: Correct me if I'm wrong, but aren't repeat `waitmovement` statements harmless (other than wasting a few bytes)?

asm/macros/event.inc
```diff
@@ -1675,11 +1688,13 @@
 	.macro moverotatingtileobjects puzzleNumber:req
 	.byte SCR_OP_MOVEROTATINGTILEOBJECTS
 	.2byte \puzzleNumber
+	waitmovement implicit=1
 	.endm
 
 	@ For the rotating tile puzzles in Mossdeep Gym / Trick House Room 7. Updates the facing direction of all objects on the puzzle tiles
 	.macro turnrotatingtileobjects
 	.byte SCR_OP_TURNROTATINGTILEOBJECTS
+	waitmovement implicit=1
```
GriffinRichards: What's the benefit of marking these "implicit" as opposed to embedding an unmodified `waitmovement` in these macros? If it's just the warning from earlier about repeated `waitmovement` statements, I'd argue we could massively simplify the macro changes by getting rid of the "implicit" handling.

If someone wanted to modify this macro to specify a map for instance, the "implicit" handling actually seems to work against them because it emits an error if you give it a map.

asm/macros/event.inc
mrgriffin: Oh! That sounds right, I'll check ASAP!

I copied the `waitstate` handling without thinking, but you're right—if it's just a waste of a few bytes it'd be much simpler to not build out all this warning machinery :)

mrgriffin:
Applied review suggestions: no more `implicit` argument to `waitmovemet` :)

GriffinRichards (approved):

GriffinRichards:
Was this waiting on something? I don't remember. It was approved and then there were some additional changes, it seems.

mrgriffin:
[On Discord at the time I said](<https://discord.com/channels/442462691542695948/442465020291317760/1517173255281049651>):
> Thanks Griffin! I've rebased that fixup into the history :)
>
> Potentially worth leaving it open for a little longer just in case people feel strongly that it's not an improvement.

And there was some discussion [starting roughly here](<https://discord.com/channels/442462691542695948/442465020291317760/1517301270992388219>). From my side, I think it's fine to be merged :)

PR #2323, opened 2026-06-18
Fix fakematch in match_call.c

AZero13:


AZero13:
Only PopulateSpeciesFromTrainerLocation is changed to remove the nonsense garbage check

AZero13:
Fixed stray dewford rand type change

AZero13:
Nonsense check removed

PR #2326, opened 2026-06-20
Remove packed range macros and use a union instead

AZero13:


mrgriffin (commented):
I'm open to being convinced that this is an improvement, but atm I'm not sure what the upside is, and the downside is needing to keep `struct ObjectEventLinkPlayer` in sync with `struct ObjectEvent` (which is an easy-to-miss trap and most projects extend `graphicsId` so it's not purely hypothetical).

PR #2328, opened 2026-06-20
✨Mark bugs that involve genuine undefined behavior as UBFix

AZero13:
The objEvent->localId != OBJ_EVENT_ID_CAMERA counts because this results in a null dereference, which is fine for vanilla I guess, but still wrong.

### DIFF
--- a/src/battle_ai_script_commands.c
+++ b/src/battle_ai_script_commands.c
@@ -1804,7 +1804,7 @@ static void Cmd_if_has_move(void)
 {
     s32 i;
 
-#ifdef BUGFIX
+#ifdef UBFIX
     const u16 move = T1_READ_16(gAIScriptPtr + 2);
     const u16 *movePtr = &move;    
 #else
@@ -1862,7 +1862,7 @@ static void Cmd_if_doesnt_have_move(void)
 {
     s32 i;
 
-#ifdef BUGFIX
+#ifdef UBFIX
     const u16 move = T1_READ_16(gAIScriptPtr + 2);
     const u16 *movePtr = &move;    
 #else
--- a/src/event_object_movement.c
+++ b/src/event_object_movement.c
@@ -2016,7 +2016,7 @@ static void LoadObjectEventPalette(u16 paletteTag)
     u16 i = FindObjectEventPaletteIndexByTag(paletteTag);
 
 // FindObjectEventPaletteIndexByTag returns 0xFF on failure, not OBJ_EVENT_PAL_TAG_NONE.
-#ifdef BUGFIX
+#ifdef UBFIX
     if (i != 0xFF)
 #else
     if (i != OBJ_EVENT_PAL_TAG_NONE)
--- a/src/battle_factory.c
+++ b/src/battle_factory.c
@@ -744,7 +744,7 @@ u8 GetFactoryMonFixedIV(u8 challengeNum, bool8 isLastBattle)
 // The Factory has an out-of-bounds access when generating the rental draft for round 9 (challengeNum==8),
 // or the "elevated" rentals from round 8 (challengeNum+1==8)
 // This happens to land on a number higher than 31, which is interpreted as "random IVs"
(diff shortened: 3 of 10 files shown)


src/battle_ai_script_commands.c
```diff
@@ -449,7 +449,7 @@ static u8 ChooseMoveOrAction_Doubles(void)
 {
     s32 i;
     s32 j;
-#ifndef BUGFIX
+#ifndef UBFIX
```
mrgriffin: Why's this one UB?

src/event_object_movement.c
```diff
@@ -514,7 +514,7 @@ static const struct SpritePalette sObjectEventSpritePalettes[] = {
     {gObjectEventPal_Lugia,                 OBJ_EVENT_PAL_TAG_LUGIA},
     {gObjectEventPal_RubySapphireBrendan,   OBJ_EVENT_PAL_TAG_RS_BRENDAN},
     {gObjectEventPal_RubySapphireMay,       OBJ_EVENT_PAL_TAG_RS_MAY},
-#ifdef BUGFIX
+#ifdef UBFIX
```
mrgriffin: idk if this is really a fix for UB. Yes, it's UB if you ask for a tag that isn't in this list, but that's a precondition of the code; if "can invoke UB when you violate a precondition" is the bar for what's UB or not then every function that dereferences a pointer without a `NULL` check should grow `UBFIX` that checks for `NULL` and that's obviously silly.

src/event_object_movement.c
```diff
@@ -8079,7 +8079,7 @@ static void DoGroundEffects_OnSpawn(struct ObjectEvent *objEvent, struct Sprite
 {
     u32 flags;
 
-#ifdef BUGFIX
+#ifdef UBFIX
```
mrgriffin: I haven't looked into it, but why's it UB for the camera to trigger ground effects?

mrgriffin (commented):
I'm not sure why there's a distinction between `BUGFIX` and `UBFIX`, but assuming it's for a good reason\* I've pointed out some cases where I'm not sure what the UB is.

\* ig it's for people that want to compile a ROM which works almost exactly the same as vanilla, but without invoking any UB? If that's the case, then the `src/battle_gfx_sfx_util.c`/`src/trainer_pokemon_sprites.c` ones wouldn't qualify. Maybe it's "people want to compile a ROM that's altered but can't invoke any UB without them violating a reasonable precondition", "reasonable" because "you can't have non-transparent pixels in the first tile's top row" is obviously not reasonable.

ketsuban:
> I'm not sure why there's a distinction between `BUGFIX` and `UBFIX`

I thought the goal was to make something like the PC port easier to do by providing a compile-time flag to enable null-pointer checks Gamefreak didn't do because dereferencing a null pointer just reads the last instruction read from the firmware, and that happens not to crash the game in those instances in the vanilla game.

src/event_object_movement.c
AZero13: if the check objEvent->localId != OBJ_EVENT_ID_CAMERA is missing, the game will attempt to run the ground effects logic on the camera sprite

AZero13: During the OAM update/rendering loop, the engine checks sprite->subspriteTableNum and attempts to read from sprite->subspriteTables[4].

Because subspriteTables is NULL, this compiles to an offset reference of NULL + 4 * sizeof(void *), trying to read from memory address 0x00000010.

mrgriffin: afaict the only call to `AddSubspritesToOamBuffer` is on the `else` of `if (!sprite->subspriteTables || sprite->subspriteMode == SUBSPRITES_OFF)`? (i.e. only entered if `subspriteTables != NULL && sprite->subspriteMode != SUBSPRITES_OFF`)

AZero13:
Addressed!

AZero13:
✨

PR #2329, opened 2026-06-21
Turn TextIntoHealthboxObject into a clean for loop

AZero13:
The if + do while here is redundant


AZero13:
Ping?

AZero13:
Ping?

mrgriffin:
The comment either has confusing text, or a confusing placement but either way that's not your fault.

PR #2330, opened 2026-06-22
Move StartMonScrollingBgMask to battle_anim_utility_func.c

AZero13:
So we don't need to do any extern shenanigans.


AZero13:
Ping

PR #2333, opened 2026-06-23
Split script to script_vm and script

AZero13:
Move gNullScriptPtr to the script_vm file.

By the way, modern will optimize the secondary null check out, so it won't check twice anymore :).

No preprocessing or MODERN for this at all, just agbcc being a bad compiler.


AZero13:
Merge conflicts fixed!

AZero13:
ping

PR #2334, opened 2026-06-24
comparison operators now use enum instead of raw numbers

pkmnsnfrn:
## Description
| Before | After|
| --- | --- |
| 0       | `LESS_THAN`      |
|  1      | `EQUAL`      |
|   2     |  `GREATER_THAN`     |
|   3     |   `LESS_THAN_OR_EQUAL`    |
|   4     |   `GREATER_THAN_OR_EQUAL`    |
|   5     |   `NOT_EQUAL`    |

## **Discord contact info**
pkmnsnfrn


include/constants/global.h
```diff
@@ -153,4 +153,16 @@
 #define CONNECTION_DIVE     5
 #define CONNECTION_EMERGE   6
 
+enum ComparisonOperators
```
GriffinRichards: I don't think these need to be global, given they're only used by `scrcmd` and it's associated macros (I think if anything a lot of what's currently in "global" should be moved to appropriate header files)

pkmnsnfrn: Fixed by [`d634eef` (this PR)](https://github.com/pret/pokeemerald/pull/2334/commits/d634eef5bf2a42a28d47f5e0f8aaeeec039aaef3)

PR #2349, opened 2026-08-02
Mac OS Install edits

RubyRaven6:
## Description
Edits to the installation information for Mac OS, as installing devkitpro is no longer necessary.

## **Discord contact info**
ruby.raven


RubyRaven6:
A comment since I realized it once I went to bed, but the installation instructions for these are currently incorrect. Will fix asap

